PR #23, opened 2023-03-11
Add encoding for TcbInfo extension structures

jhand2:
Add X.509 encoding for tcg-dice-TcbInfo and tcg-dice-MultiTcbInfo

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -3,7 +3,11 @@
 //! DPE requires encoding variable-length certificates. This module provides
 //! this functionality for a no_std environment.
 
-use crate::{response::DpeErrorCode, DpeProfile, DPE_PROFILE};
+use crate::{
+    dpe_instance::{TciMeasurement, TciNodeData},
+    response::DpeErrorCode,
+    DpeProfile, DPE_PROFILE,
+};
 
 pub struct EcdsaSignature {
     r: [u8; DPE_PROFILE.get_ecc_int_size()],
@@ -20,12 +24,18 @@ pub struct Name<'a> {
     serial: &'a str,
 }
 
+pub struct MeasurementData<'a> {
+    label: &'a [u8],
+    tci_nodes: &'a [&'a TciNodeData],
+}
+
 pub struct X509CertWriter<'a> {
     certificate: &'a mut [u8],
     offset: usize,
 }
 
 impl X509CertWriter<'_> {
+    const BOOL_TAG: u8 = 0x1;
     const INTEGER_TAG: u8 = 0x2;
     const BIT_STRING_TAG: u8 = 0x3;
     const OCTET_STRING_TAG: u8 = 0x4;
@@ -36,6 +46,10 @@ impl X509CertWriter<'_> {
     const SEQUENCE_OF_TAG: u8 = 0x30;
     const SET_TAG: u8 = 0x31;
     const SET_OF_TAG: u8 = 0x31;
+
+    // Constants for setting tag bits
+    const PRIVATE: u8 = 0x80; // Used for Implicit/Explicit tags
+    const CONSTRUCTED: u8 = 0x20; // SET{OF} and SEQUENCE{OF} have this bit set
 
     const X509_V3: u64 = 2;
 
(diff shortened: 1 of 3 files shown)


dpe/src/x509.rs
```diff
@@ -483,6 +563,136 @@ impl X509CertWriter<'_> {
...
+    fn encode_tcb_info(&mut self, node: &TciNodeData) -> Result<usize, DpeErrorCode> {
+        let tcb_info_size = Self::get_tcb_info_size(node, /*tagged=*/ false)?;
+        // TcbInfo sequence
+        let mut bytes_written = self.encode_byte(Self::SEQUENCE_TAG)?;
+        bytes_written += self.encode_size_field(tcb_info_size)?;
+
+        // fwids SEQUENCE OF
+        // IMPLICIT [6] Constructed
+        let fwid_size = Self::get_fwid_size(&node.tci_current.0, /*tagged=*/ true)?;
+        bytes_written += self.encode_byte(0xA6)?;
```
zhalvorsen, line 605: Maybe it would be more self-documenting if they have `const` describing the upper bits OR'd together.
```suggestion
        bytes_written += self.encode_byte(IMPLICIT | CONSTRUCTED | 0x06)?;
```

There are probably better names for them.

dpe/src/x509.rs
```diff
@@ -483,6 +563,136 @@ impl X509CertWriter<'_> {
...
+        // vendorInfo OCTET STRING
+        // IMPLICIT[8] Primitive
+        let vinfo = if node.flag_is_internal() {
+            "VNDR"
+        } else {
+            "USER"
+        };
+        bytes_written += self.encode_byte(0x88)?;
+        bytes_written += self.encode_size_field(vinfo.len())?;
+        bytes_written += self.encode_bytes(vinfo.as_bytes())?;
```
zhalvorsen, line 623: If these are byte strings, then they can be used directly below.
```suggestion
        let vinfo = if node.flag_is_internal() {
            b"VNDR"
        } else {
            b"USER"
        };
        bytes_written += self.encode_byte(0x88)?;
        bytes_written += self.encode_size_field(vinfo.len())?;
        bytes_written += self.encode_bytes(vinfo)?;
```

dpe/src/x509.rs
```diff
@@ -483,6 +563,136 @@ impl X509CertWriter<'_> {
...
+        &mut self,
+        measurements: &MeasurementData,
+    ) -> Result<usize, DpeErrorCode> {
+        let tcb_info_size =
+            Self::get_tcb_info_size(measurements.tci_nodes[0], /*tagged=*/ true)?;
+        let tcb_infos_size = tcb_info_size * measurements.tci_nodes.len();
+        let multi_tcb_info_size = Self::get_structure_size(tcb_infos_size, /*tagged=*/ true)?;
+        let extension_size = Self::get_structure_size(Self::MULTI_TCBINFO_OID.len(), /*tagged=*/true)? +
+            Self::get_structure_size(1, /*tagged=*/true)? + // bool size
+            Self::get_structure_size(multi_tcb_info_size, /*tagged=*/true)?; // Octet string
```
zhalvorsen, line 647: Can you use the helper functions for these?
```suggestion
        let multi_tcb_info_size = Self::get_multi_tcb_info_size(measurements, /*tagged=*/ true)?;
        let extension_size = Self::get_extensions_size(measurements, /*tagged=*/true)?; // Octet string
```

dpe/src/x509.rs
```diff
@@ -483,6 +563,136 @@ impl X509CertWriter<'_> {
...
+        bytes_written += self.encode_size_field(1)?;
+        bytes_written += self.encode_byte(0xFF)?; // Mark extension as critical
+
+        bytes_written += self.encode_byte(Self::OCTET_STRING_TAG)?;
+        bytes_written += self.encode_size_field(Self::get_structure_size(
+            tcb_infos_size,
+            /*tagged=*/ true,
+        )?)?;
+
+        // Encode MutliTcbInfo
```
zhalvorsen, line 664: sp: MultiTcbInfo
```suggestion
        // Encode MultiTcbInfo
```

dpe/src/x509.rs
```diff
@@ -701,15 +969,31 @@ mod tests {
...
+                &measurements,
             )
             .unwrap();
 
+        let mtcb_size =
+            X509CertWriter::get_multi_tcb_info_size(&measurements, /*tagged=*/ true).unwrap();
+        let ext_size =
+            X509CertWriter::get_extensions_size(&measurements, /*tagged=*/ true).unwrap();
+
+        let mut f = std::fs::File::create("/tmp/tbs.der").unwrap();
```
zhalvorsen, line 994: It doesn't look like this file is being used anywhere else. Do you need it?

dpe/src/x509.rs
```diff
@@ -197,13 +215,74 @@ impl X509CertWriter<'_> {
...
+    ) -> Result<usize, DpeErrorCode> {
+        if measurements.tci_nodes.is_empty() {
+            return Err(DpeErrorCode::InternalError);
+        }
+
+        // Size of concatenated tcb infos
+        let tcb_infos_size = measurements.tci_nodes.len()
+            * Self::get_tcb_info_size(measurements.tci_nodes[0], /*tagged=*/ true)?;
+
+        // Size if tcb infos including SEQUENCE OF tag/size
```
zhalvorsen, line 254: Typo

```suggestion
        // Size of tcb infos including SEQUENCE OF tag/size
```

zhalvorsen (changes requested):

dpe/src/x509.rs
jhand2: Will do. I did it this way to avoid recomputing the same intermediate values, but it's probably better to do this for clarity instead of saving a few cycles.

zhalvorsen: SG. I think it should all be inlined anyway, so it shouldn't change performance.

zhalvorsen (approved):

PR #24, opened 2023-03-13
Add a struct to describe bitmaps.

zhalvorsen:


dpe/src/bitmap.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Bitmap with only one flag set.
+    pub const fn flag(flag: usize) -> Bitmap {
+        assert!(flag < u32::BITS as usize);
+        Bitmap(1 << flag)
+    }
+
+    /// Get the value of the given flag.
+    pub const fn get(&self, flag: usize) -> bool {
+        assert!(flag < u32::BITS as usize);
```
jhand2, line 26: I know this is expected to never happen, but I think we want to avoid panic wherever possible. I wonder if it's possible to make the compiler enforce the values passed in here. Otherwise I think we should return a Result.

zhalvorsen: We could do some enum magic. Let me play around with it and see what I can come up with

dpe/src/bitmap.rs
```diff
@@ -0,0 +1,78 @@
+/*++
+Licensed under the Apache-2.0 license.
+
+Abstract:
+    Helper struct for controlling and viewing bitmaps.
+--*/
+
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq, Default)]
```
ia0, line 9: Do we want to make this Copy? If yes, we may want to take `self` instead of `&self` in methods.

dpe/src/bitmap.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct Bitmap(u32);
+
+impl Bitmap {
+    /// Empty flags.
+    pub const fn const_default() -> Bitmap {
+        Bitmap(0)
+    }
+
+    /// Bitmap with only one flag set.
+    pub const fn flag(flag: usize) -> Bitmap {
```
ia0, line 19: We might want to implement `BitOr` and `BitOr<usize>` to be able to write:
- `Bitmap::flag(FOO_FLAG) | Bitmap::flag(BAR_FLAG)`
- `Bitmap::default() | FOO_FLAG | BAR_FLAG`
- `Bitmap::flag(FOO_FLAG) | BAR_FLAG`

ia0 (approved):

jhand2:
I'm going to close this since we now use the bitflags crate

PR #28, opened 2023-03-15
Add concept of hardware localities.

zhalvorsen:
These are used to distinguish between requestors. One locality cannot access a handle from another locality. The spec describes what a locality is.

```
[A]n SoC may use hardware localities to distinguish between:

- Two co-processors each have a distinct Hardware Locality.
- An processor with multiple execution levels. Each has a distinct Hardware Locality.
```

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -14,9 +14,10 @@ use crate::{
 
 use core::marker::PhantomData;
 
-pub struct DpeInstance<C: Crypto> {
+pub struct DpeInstance<'a, C: Crypto> {
     contexts: [Context; MAX_HANDLES],
     support: Support,
+    localities: &'a [u32],
 
     // All functions/data in C are static and global. For this reason
     // DpeInstance doesn't actually need to hold an instance. The PhantomData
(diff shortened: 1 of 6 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -14,29 +14,41 @@ use crate::{
...
 
-    pub fn new(support: Support) -> DpeInstance<C> {
+    /// Create a new DPE instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `support` - optional functionality the instance supports
+    /// * `localities` - all possible valid localities for the system
+    pub fn new(support: Support, localities: &[u32]) -> DpeInstance<C> {
+        assert!(!localities.is_empty());
```
jhand2, line 39: nit: One concern I have is if we panic, it will be pretty hard to debug for SoC integrators. I think I would prefer we surface an error to runtime firmware so it can give some diagnostic information in the status registers.

The list of supported localities will probably ultimately be provided by the SoC integrator (even if it's just part of the Caliptra firmware image metadata)

zhalvorsen: That's fair. I figured this should be caught way before anything gets shipped because it wouldn't even boot. But, debugging those scenarios can be a nightmare. I changed it to return a `Result`.

PR #31, opened 2023-03-16
Implement CertifyKey command

jhand2:
Implement the CertifyKey command. This includes the following additional support features:

* Add Sign with Alias Key to Crypto trait
* Fix bugs in certificate encoding of SubjectPublicKeyInfo

### DIFF
--- a/dpe/src/commands.rs
+++ b/dpe/src/commands.rs
@@ -11,6 +11,7 @@ use core::mem::size_of;
 pub enum Command {
     GetProfile,
     InitCtx(InitCtxCmd),
+    CertifyKey(CertifyKeyCmd),
     RotateCtx(RotateCtxCmd),
     DestroyCtx(DestroyCtxCmd),
     TagTci(TagTciCmd),
@@ -42,7 +43,7 @@ impl Command {
             Command::GET_PROFILE => Ok(Command::GetProfile),
             Command::INITIALIZE_CONTEXT => Ok(Command::InitCtx(InitCtxCmd::try_from(bytes)?)),
             Command::DERIVE_CHILD => Err(DpeErrorCode::InvalidCommand),
-            Command::CERTIFY_KEY => Err(DpeErrorCode::InvalidCommand),
+            Command::CERTIFY_KEY => Ok(Command::CertifyKey(CertifyKeyCmd::try_from(bytes)?)),
             Command::SIGN => Err(DpeErrorCode::InvalidCommand),
             Command::ROTATE_CONTEXT_HANDLE => {
                 Ok(Command::RotateCtx(RotateCtxCmd::try_from(bytes)?))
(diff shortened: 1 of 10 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -144,6 +149,90 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
+        response.cert_size = match u32::try_from(bytes_written) {
+            Ok(n) => n,
+            Err(_) => return Err(DpeErrorCode::InternalError),
+        };
+
+        response.derived_pubkey_x.copy_from_slice(&pub_key.x);
+        response.derived_pubkey_y.copy_from_slice(&pub_key.y);
+
+        Ok(response)
+    }
```
zhalvorsen, line 234: It looks like the `new_context_handle` hasn't been assigned. Alternatively, the CertifyKeyResp could be created after all of the operations to help avoid missing struct members. I don't feel too strongly either way, though. Feel free to do whatever you think is best.

```suggestion
        let mut tbs_digest = [0u8; DPE_PROFILE.get_hash_size()];
        C::hash(DPE_PROFILE, &tbs_buffer[..bytes_written], &mut tbs_digest)?;
        let sig = C::ecdsa_sign_with_alias(DPE_PROFILE, &tbs_digest)?;

        let mut cert = [0u8; MAX_CERT_SIZE];
        let mut cert_writer = X509CertWriter::new(&mut cert);
        bytes_written = cert_writer.encode_ecdsa_certificate(&tbs_buffer[..bytes_written], &sig)?;
        let cert_size = match u32::try_from(bytes_written) {
            Ok(n) => n,
            Err(_) => return Err(DpeErrorCode::InternalError),
        };

        Ok(CertifyKeyResp {
            new_context_handle: [0; HANDLE_SIZE], // would need to change this value
            derived_pubkey_x: pub_key.x,
            derived_pubkey_y: pub_key.y,
            cert_size,
            cert,
        })
```

dpe/src/commands.rs
```diff
@@ -477,11 +535,22 @@ mod tests {
         }
     }
 
+    impl From<CertifyKeyCmd> for Vec<u8> {
+        fn from(value: CertifyKeyCmd) -> Self {
+            let mut raw = vec![];
+            raw.extend(value.handle);
+            raw.extend_from_slice(&value.flags.to_le_bytes());
+            raw.extend(value.label);
+            raw
```
ia0, line 544: Similarly, those types of function could be `bytemuck::bytes_of(&value).to_vec()`.

dpe/src/commands.rs
```diff
@@ -176,6 +177,54 @@ impl TryFrom<&[u8]> for RotateCtxCmd {
     }
 }
 
+#[repr(C)]
```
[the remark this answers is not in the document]
ia0: Lol, I was actually independently writing the following in another comment when I received the notification for your comment:

---

It looks to me that the "parsing" logic for commands may be somehow factorized. Here are some alternatives:
1. Use `bytemuck::try_from_bytes()` and `#[derive(bytemuck::AnyBitPattern)]`. This means adding a dependency.
2. Do the same as `bytemuck` but manually (essentially an unsafe cast). This means using unsafe and losing the static checks from the derive macro.
3. Use `bytes::Buf::copy_to_slice()` and `bytes::Buf::get_u32_le()`. This means adding a dependency and still duplicating the struct layout in the parsing code.
4. Do the same as `bytes` but manually (essentially a wrapper around the slice with appropriate reading methods). This means a bit of code, but would remove a lot of error-prone boilerplate.

I think (1) is the best solution, but if adding a dependency is an issue, then probably (4) is better than (2), just because it avoids unsafe. This is also closer to the current code, just factorizing the index arithmetic.

---

I agree with you that (1) and (2) have another advantage over (3) and (4) that I forgot. Not only do they remove duplication between struct layout and parsing logic, but they also don't allocate.

dpe/src/dpe_instance.rs
```diff
@@ -288,6 +416,22 @@ impl TciNodeData {
             tci_current: TciMeasurement([0; DPE_PROFILE.get_tci_size()]),
         }
     }
+
+    fn serialize(&self, dst: &mut [u8]) -> Result<usize, DpeErrorCode> {
```
ia0, line 420: This function could also use something like `bytes::BufMut` (or a manual similar wrapper) to factorize the offset arithmetic. It would be something like:

```rust
fn serialize(&self, mut dst: &mut [u8]) -> Result<usize, DpeErrorCode> {
    let initial_len = dst.len();
    if dst.len() < size_of::<Self>() {
         return Err(...);
    }
    dst.put_u32_le(self.tci_type);
    dst.put(&self.tci_cumulative.0);
    dst.put(&self.tci_current.0);
    Ok(initial_len - dst.remaining_mut())
}

dpe/src/commands.rs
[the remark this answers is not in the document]
jhand2: Going to leave it out of this PR, but I agree what we're doing today is a bit cumbersome and error-prone. I opened #40 to track this.

dpe/src/dpe_instance.rs
jhand2: I think the main reason I did it this way was to avoid copying the whole cert buffer. But now that I'm thinking about it, the compiler is probably smart enough to avoid that copy.

dpe/src/commands.rs
jhand2: Replaced with zerocopy

PR #33, opened 2023-03-16
Add more verification of the GetProfile command.

zhalvorsen:


verification/verification_test.go
```diff
@@ -56,3 +100,7 @@ func TestInitializeContext(t *testing.T) {
 		t.Fatal("Failed to initialize default context.")
 	}
 }
+
+func getLocalities() []uint32 {
+	return []uint32{AUTO_INIT_LOCALITY, OTHER_LOCALITY}
+}
```
jhand2, line 106: Which localities are supported will probably be a per-chip property. I wonder if it makes sense to put this as a method on Transport.

Or maybe more generally, having some set of target-specific properties.

Don't need to do it in the PR, just thinking out loud.

zhalvorsen: That's way better. I moved it as well as some of the other configurable values.

jhand2 (approved):

PR #35, opened 2023-03-20
Change command tests to use zerocopy.

zhalvorsen:


dpe/Cargo.toml
```diff
@@ -12,6 +12,8 @@ dpe_profile_p384_sha384 = []
 
 [dev-dependencies]
 asn1 = "0.13.0"
+bytemuck = {version = "1.13", features = ["derive"]}
```
korran, line 15: FYI: we're using zerocopy in the boot ROM; might want to standardize on that if you don't have a preference.

zhalvorsen: Good to know, I switched it to use `zerocopy`. I think it would be best to be consistent.

ia0 (dismissed):
The idea is to only use `AsBytes` in tests or to also use `FromBytes` in release too?

In the first case, I would also use `FromBytes` in tests to test that the current parsing functions do the same as `FromBytes::read_from_prefix()` on the hard-coded test commands.

ia0 (dismissed):

dpe/src/commands.rs
```diff
@@ -238,8 +243,8 @@ mod tests {
     use super::*;
     use crate::dpe_instance::tests::{SIMULATION_HANDLE, TEST_HANDLE};
     use crate::DpeProfile;
-    use std::vec;
-    use std::vec::Vec;
+    use zerocopy::AsBytes;
+    use zerocopy::FromBytes;
```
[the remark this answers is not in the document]
zhalvorsen: Interesting. I did it manually.

jhand2 (approved):

korran:
The ROM is using zerocopy for the actual release, so I think it's open-season to use it here too.

PR #39, opened 2023-03-23
Add running hash capability to Crypto trait.

zhalvorsen:
This will allow us to hash multiple things that may not be in consecutive memory.


dpe/src/crypto.rs
```diff
@@ -44,8 +44,31 @@ impl Default for EcdsaPub {
...
 
+pub trait Hasher {
+    /// Adds a chunk to the running hash.
+    ///
+    /// # Arguments
+    ///
+    /// * `bytes` - Value to add to hash.
+    fn update(&mut self, bytes: &[u8]) -> Result<(), DpeErrorCode>;
+
+    /// Finish a running hash operation and return the result.
```
jhand2, line 55: Is the expectation that after finish is called, the hasher can be reused? Probably doesn't matter either way, just would be good to document the expected behavior.

dpe/src/crypto.rs
```diff
@@ -44,8 +44,34 @@ impl Default for EcdsaPub {
...
+    ///
+    /// Once this function has been called, the object can no longer be used and
+    /// a new one must be created to hash more data.
+    ///
+    /// # Arguments
+    ///
+    /// * `digest` - Where the computed digest should be written.
+    fn finish_mut(self, digest: &mut [u8]) -> Result<(), DpeErrorCode>;
+
+    fn finish(self) -> Result<[u8; DPE_PROFILE.get_hash_size()], DpeErrorCode> {
```
ia0, line 65: Why do we assume `DPE_PROFILE` here and let the user choose the profile in `Crypto::hash()` and `Crypto::hash_initialize()`? It looks like we either want to always assume `DPE_PROFILE` or make `Hasher` depend on the profile, like `trait Hasher<const PROFILE: DpeProfile>` (we may need some nightly feature, not sure).

simulator/src/main.rs
ia0, line 1: I'm not sure I follow, but it looks like this `OpensslHasher` implementation is the same as in the DPE tests. Shouldn't we share them?

dpe/src/crypto.rs
zhalvorsen: That's a good point. I spaced we weren't depending on that definition anywhere else for `Crypto`. I removed the function.

simulator/src/main.rs
zhalvorsen: They are duplicates. I tried abstracting it out, but it didn't do too much. I'm having trouble finding a better way to do it. Can you think of one?

We would get a circular dependency if we put it in `ossl_crypto`. `dpe` tests depend on `ossl_crypto` and it would depend on `dpe` for the `Hasher` trait and `DpeErrorCode`.

We could put it in `dpe/src/crypto.rs` and add an `openssl` `feature` with an optional dependency. If we did this, would we then need to always include `--features=openssl` when we ran `cargo test`? Is there a way to have default features when running tests?

Can you think of a better way to detangle them?

ia0: I might be missing stuff, we could ask for help. But here are the options:

1. The circular dependency as you said. This is possible because it goes through a dev-dependency. However, it looks like there are 2 versions of the trait: one from the crate under test, and one from the helper crate implementing the trait of the crate under test. I couldn't find a way to unify them. Also, this circular dependency may bite us in the future. For example it's difficult to publish crates with dependencies, since publishing is currently crate by crate and not as a group of crates.
2. The optional dependency as you said. I couldn't find a wait to have default features for test, but a `./test.sh` script could be used (and include more than one cargo invocation).
3. Using integration tests instead of unit tests. If the tests are moved to some `dpe/tests/crypto.rs` file, then we don't hit the issue in (1). However, we introduce the circular dependency again.
4. Move the tests to the `ossl_crypto` crate instead.
5. Duplicate implementations.

I think most of those options are better than code duplication. If we're not afraid of circular dependencies, then we can use (3). Otherwise (4) is probably better than (2), given that (1) simply doesn't work. But maybe this needs more information and should be addressed later. I don't know.

Actually a 6th option would be to have yet another crate just for tests that depends on external implementations, if we don't want to single out the openssl implementation.

jhand2: I think (4) is a bit problematic because most of the command tests (in dpe_instance.rs) need to rely on some crypto implementation.

But overall I agree the code duplication isn't super fun. I think I like 3, but as you said, the circular dependency scares me a bit.

ia0 (approved):

simulator/src/main.rs
zhalvorsen: I think I figured something out that will be a lot better. It took a decent refactor though. I'll try to get it up soon.

I ended up moving crypto into its own crate so it doesn't have any dependencies. The functions now return `Option`s so they don't rely on `DpeErrorCode`. The `crypto` crate has an `openssl` feature which gets included in `dev-dependencies` but not `dependencies`. I have a partial duplicate of `DpeProfile`, but if that is the only duplicate I think its a lot cleaner.

ia0: Good idea. Indeed there could be a `dpe-types` crate with only common types/interfaces (traits, structs, enums, etc). And `dpe` would use those types, define more types, and implement something on those types. So I would still put `DpeProfile` in that `dpe-types` crate. The `DPE_PROFILE` constant though would still be in `dpe` probably.

PR #41, opened 2023-03-24
Separate Crypto into its own crate.

zhalvorsen:
Hopefully this will make it easier to detangle crypto implementations from DPE.


crypto/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        pub_x: &mut [u8],
+        pub_y: &mut [u8],
+    ) -> Option<()>;
+
+    /// Sign `digest` with the platform Alias Key
+    fn ecdsa_sign_with_alias(
+        algs: AlgLen,
+        digest: &[u8],
+        sig_r: &mut [u8],
+        sig_s: &mut [u8],
```
zhalvorsen, line 118: As a heads up, I also changed these to have mutable slice outputs because I they were relying on `DPE_PROFILE` for sizes.

crypto/src/openssl.rs
```diff
@@ -0,0 +1,158 @@
...
+        let ec_priv = EcKey::from_private_components(&group, &priv_bn, &pub_point).ok()?;
+
+        let sig = EcdsaSig::sign::<Private>(digest, &ec_priv).ok()?;
+        sig_r.copy_from_slice(sig.r().to_vec().as_slice());
+        sig_s.copy_from_slice(sig.s().to_vec().as_slice());
+        Some(())
+    }
+}
+
+struct EccPoint {
```
jhand2, line 155: Can probably ditch this struct since we aren't returning it anymore

crypto/src/openssl.rs
```diff
@@ -0,0 +1,158 @@
...
+            .mul_generator(&group, &priv_bn, &mut bn_ctx)
+            .unwrap();
+
+        let mut x = BigNum::new().unwrap();
+        let mut y = BigNum::new().unwrap();
+        pub_point
+            .affine_coordinates(&group, &mut x, &mut y, &mut bn_ctx)
+            .unwrap();
+
+        let point = EccPoint {
```
jhand2, line 119: can collapse this to just copy to the x/y slices directly

crypto/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Derive a CDI based on the current base CDI and measurements.
+    ///
+    /// # Arguments
+    ///
+    /// * `profile` - Which profile is being used. This will tell the platform
+    ///   which algorithm to use
+    /// * `measurement_digest` - A digest of the measurements which should be
+    ///   used for CDI derivation
+    /// * `info` - Caller-supplied info string to use in CDI derivation
+    fn derive_cdi(algs: AlgLen, measurement_digest: &[u8], info: &[u8]) -> Option<Self::Cdi>;
```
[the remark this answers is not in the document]
jhand2: I think this is probably ok. I see this crate as "A DPE crypto interface" rather than a general-purpose crypto interface

crypto/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "crypto"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+default = []
```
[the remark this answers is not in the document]
zhalvorsen: Oh yeah, thanks!

crypto/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+pub trait Crypto {
+    type Cdi;
+    type Hasher: Hasher;
+
+    /// Fills the buffer with random values.
+    ///
+    /// # Arguments
+    ///
+    /// * `dst` - The buffer to be filled.
+    fn rand_bytes(dst: &mut [u8]) -> Option<()>;
```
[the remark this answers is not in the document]
zhalvorsen: I'm open to either way, which do you think would be better?

crypto/src/openssl.rs
zhalvorsen: Done

crypto/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// * `label` - Caller-supplied label to use in asymmetric key derivation
+    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
+    ///
+    /// Returns a derived public key
+    fn derive_ecdsa_pub<const SIZE: usize>(
+        algs: AlgLen,
+        cdi: &Self::Cdi,
+        label: &[u8],
+        info: &[u8],
+    ) -> Option<([u8; SIZE], [u8; SIZE])>;
```
zhalvorsen, line 110: I changed these to make it easier for the compiler to catch us using the wrong sized arrays.

crypto/src/lib.rs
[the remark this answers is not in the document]
jhand2: I mildly prefer Result just because I think it makes it more clear that it's an error condition

crypto/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Cryptographically hashes the given buffer.
+    ///
+    /// # Arguments
+    ///
+    /// * `profile` - Which profile is being used. This will tell the platform which algorithm to
+    ///   use.
+    /// * `bytes` - Value to be hashed.
+    /// * `digest` - Where the computed digest should be written.
+    fn hash<const SIZE: usize>(algs: AlgLen, bytes: &[u8]) -> Option<[u8; SIZE]> {
```
jhand2, line 66: I'm not sure if I have a good suggestion, but it feels a big weird to pass both SIZE and an AlgLen since they're expected to always be the same.

zhalvorsen: That's a good point. I think I might be able to figure out a way to get rid of that enum and just use the generic. WDYT?

This'll probably have to wait till Monday though.

crypto/src/lib.rs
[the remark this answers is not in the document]
zhalvorsen: SG I'll can swap it for a Result

crypto/src/lib.rs
jhand2: I think just using the generic makes sense. No rush!

zhalvorsen: Enum const generics aren't supported by stable quite yet. I think it might actually be best to go back to using an enum. I think that is more obvious about how to use it and it is the same convention as OpenSSL.

crypto/src/lib.rs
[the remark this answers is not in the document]
zhalvorsen: Done

jhand2 (approved):

PR #43, opened 2023-03-29
Fix, add Go client support, and add test for CertifyKey

chrisfenner:
This change consists of 3 commits:

1. Introduction of type parameters to the Go `Client` type. This allows compile-time logic about the ABI of commands (which depends on the profile) without duplicating all the command code for the 256-bit and 384-bit flavors of DPE. Convenience helpers are provided so that callers do not have to interact with the generics.
2. Increase the request and response buffer size to 4KB in the simulator.
3. Add client and test for CertifyKey, which is interesting because it has an ABI that depends on the profile. Therefore, its request and response structs also have type constraints.


verification/client.go
```diff
@@ -50,7 +101,8 @@ func (c *Client) InitializeContext(cmd *InitCtxCmd) (*InitCtxResp, error) {
 	return &respStruct, nil
 }
 
-func (c *Client) GetProfile() (*GetProfileResp, error) {
+// getProfile is an internal helper for handling GetProfile as part of either the client API or initialization.
+func getProfile(t Transport) (*GetProfileResp, error) {
```
zhalvorsen, line 105: Do you think it would be worth while to change this or add another function to be a factory for `Client`. In the tests we could call the factory with the transport and it would return the generic version of the client instead of calling `NewClient256` directly.

I guess in this case we would need to add a function to the `TestDPEInstance` that returns the expected profile to make sure we got what we expected, so maybe it doesn't buy too much for us.

verification/verification_test.go
```diff
@@ -198,3 +198,59 @@ func testInitContext(d TestDPEInstance, t *testing.T) {
...
+	certifyKeyReq := CertifyKeyReq[SHA256Digest]{
+		ContextHandle: [16]byte{0},
+		Flags:         0,
+		Label:         [32]byte{0},
+	}
+	certifyKeyResp, err := client.CertifyKey(&certifyKeyReq)
+	if err != nil {
+		t.Fatalf("Could not certify key: %v", err)
+	}
+	t.Logf("Certificate: %x", certifyKeyResp.Certificate)
```
zhalvorsen, line 238: Should we do any initial checks to see if the certificate that came back can be parsed. Or would that be better in a follow on PR?

verification/client.go
```diff
@@ -76,18 +128,51 @@ func (c *Client) GetProfile() (*GetProfileResp, error) {
...
+// CertifyKey calls the DPE CertifyKey command.
+func (c *Client[CurveParameter, Digest]) CertifyKey(cmd *CertifyKeyReq[Digest]) (*CertifyKeyResp[CurveParameter, Digest], error) {
+	// Define an anonymous struct for the response, because we have to accept the variable-sized certificate.
+	respStruct := struct {
+		NewContextHandle  [16]byte
+		DerivedPublicKeyX CurveParameter
+		DerivedPublicKeyY CurveParameter
+		CertificateSize   uint32
+		Certificate       [2048]byte
+	}{}
```
zhalvorsen, line 156: I'm still pretty new to Go, is possible to do something like this to avoid having two very similar structs?

```suggestion
	respStruct := CertifyKeyResp[CurveParameter, Digest]{}
```

zhalvorsen (commented):
It looks good to me, but I don't know if I should be the approver because I don't have very much Go experience.

verification/client.go
chrisfenner: I spent a lot of time trying to think of a nice way to do this. Go's generics are quite limited by roughly "what is possible without making the compiler do a lot of work." So, we can have a factory function, but its return type would be `interface{}` (aka any) because there is not a useful interface that covers both flavors of `Client[C, D]`. (A type constraint could be written as `interface { Client[X1, X2] | Client[Y1, Y2] }`, but that is not a valid return type).

So, the approach I've taken here is to require the caller to have an opinion about what instantiation of Client they expect to get. They will get an error if they provide the wrong one. It's not ideal, maybe we can make it nicer in the future.

verification/verification_test.go
chrisfenner: That's a good idea, let me try something basic here. I expect this test will grow a bit in the future (for example, to compare the keys involved in CertifyKey and Sign)

verification/client.go
zhalvorsen: Okay, that seems good to me. Maybe `testCertifyKey` can have a parameter later on for which `Client` to create.

verification/verification_test.go
zhalvorsen: With what you found it sounds like it would be better in a follow-on PR.

chrisfenner: Sounds good to me. (For posterity, when I tried to implement this I found that the serial numbers produced by the DPE are not printable strings)

verification/verification_test.go
```diff
@@ -198,3 +198,59 @@ func testInitContext(d TestDPEInstance, t *testing.T) {
...
+
+	if d.GetSupport().Simulation {
+		// Certify keys for each new TCI
+		for i := uint32(0); i < client.MaxTciNodes-1; i++ {
+			initCtxResp, err := client.InitializeContext(NewInitCtxIsSimulation())
+			if err != nil {
+				t.Fatal("The instance should be able to create a simulation context.")
+			}
+			defer client.DestroyContext(NewDestroyCtx(initCtxResp.Handle, false))
```
jhand2, line 247: Does this get called on each for loop iteration or when the function exits?

If it's the former, it's maybe a bit of an odd test because it's basically doing the same thing over and over again.

Also maybe a more interesting test would be to call DeriveChild then CertifyKey in each iteration. Then you'll end up with certs with different numbers of TCIs populated.

chrisfenner: Ah, DeriveChild is not yet supported. 😅 I've replaced this loop with a TODO.

jhand2 (approved):

PR #45, opened 2023-03-29
Add ExtendTci Command.

zhalvorsen:


ia0 (dismissed):

dpe/src/dpe_instance.rs
```diff
@@ -375,6 +409,22 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
+    ///
+    /// * `idx` - the index of the context to rotate
+    fn rotate_context_handle(&mut self, idx: usize) -> Result<(), DpeErrorCode> {
+        if idx >= MAX_HANDLES {
+            return Err(DpeErrorCode::InternalError);
+        }
+        self.contexts[idx].handle = match self.contexts[idx].context_type {
+            ContextType::Default => Self::DEFAULT_CONTEXT_HANDLE,
+            _ => self.generate_new_handle()?,
+        };
```
jhand2, line 424: This behavior might be a little bit weird when we go to implement the RotateContextHandle command. Because in that case we actually do want to rotate away from the default context.

Maybe this function just needs a different name.

zhalvorsen: That's a good point. What do you think of `update_context_handle`?

zhalvorsen: Or `reassign_*`?

zhalvorsen: Changed to `roll_onetime_use_handle`.

jhand2 (approved):

PR #47, opened 2023-03-30
Add `log` and `env_logger` to the simulator.

zhalvorsen:
Now that there are many calls happening in the verification tests, the
output is being flooded with simulator prints. This change adds some
third party crates to add logging capabilities. This will make it easier
to parse errors when they happen.

### DIFF
--- a/simulator/src/main.rs
+++ b/simulator/src/main.rs
@@ -1,5 +1,6 @@
 use clap::Parser;
 use crypto::OpensslCrypto;
+use log::{error, info, trace, warn};
 use std::fs;
 use std::io::{Error, ErrorKind, Read, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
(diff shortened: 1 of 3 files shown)


ia0 (dismissed):
There are a few alternatives (in case they are relevant, I don't know):
1. Output the important (non-debugging) stuff on stdout and the debugging stuff on stderr (using `eprintln!` instead of `println!`), then using `2>/dev/null` to discard the useless part.
2. Use the `log` crate and output with appropriate level. Then using `env_logger` crate, it is possible to select what to see with the `RUST_LOG` environment variable instead of command line flags. It is also possible to select different levels with module granularity.

simulator/src/main.rs
```diff
@@ -24,33 +25,31 @@ fn handle_request(dpe: &mut DpeInstance<OpensslCrypto>, stream: &mut UnixStream)
...
 
 fn cleanup() {
     match fs::remove_file(SOCKET_PATH) {
-        Ok(_) => {
-            println!();
-        }
+        Ok(_) => (),
         Err(_) => {
-            println!("Warning: Unable to unlink {SOCKET_PATH}");
+            warn!("Unable to unlink {SOCKET_PATH}");
```
ia0, line 52: nit:

```rust
if let Err(e) = fs::remove_file(SOCKET_PATH) {
    warn!("Unable to unlink {SOCKET_PATH}: {e}");
}
```

ia0 (dismissed):

zhalvorsen:
I like the idea of using log levels for debugging different scenarios. Its pretty easy to enable too. Thanks for the suggestion!

simulator/src/main.rs
zhalvorsen: Done

ia0 (approved):

jhand2 (approved):

PR #49, opened 2023-03-30
Change context handle to be unique per locality.

zhalvorsen:
This change is mainly to pave the way to be able to have a default context per locality. It still checks to make sure random handles are generated uniquely. But, because all default handles have the same value, it needs to be specified which locality the context is in.


dpe/src/dpe_instance.rs
```diff
@@ -23,6 +23,7 @@ pub struct DpeInstance<'a, C: Crypto> {
     contexts: [Context; MAX_HANDLES],
     support: Support,
     localities: &'a [u32],
+    has_initialized: bool,
```
jhand2, line 26: Might be good to have a comment to explain this (something about how `InitializeContext(simulation=false)` is only allowed once)

dpe/src/dpe_instance.rs
```diff
@@ -385,10 +390,11 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
     }
 
-    fn get_default_context(&mut self) -> Option<&mut Context> {
+    fn _get_default_context(&mut self, locality: u32) -> Option<&mut Context> {
         self.contexts.iter_mut().find(|context| {
             matches!(context.state, ContextState::Active | ContextState::Locked)
                 && context.handle == Self::DEFAULT_CONTEXT_HANDLE
+                && context.locality == locality
         })
     }
```
jhand2, line 399: Is this function needed? Seems like in most cases we would just call `get_active_context_pos`

dpe/src/dpe_instance.rs
zhalvorsen: Thats a good idea. I added a couple comments.

dpe/src/dpe_instance.rs
zhalvorsen: Yeah, its pretty useless especially because I'm going to be taking out the `ContextType::Default` and `ContextState::Locked` in a PR soon.

ia0 (dismissed):

jhand2 (dismissed):

jhand2:
lgtm, will merge once conflicts are resolved

jhand2 (approved):

PR #52, opened 2023-03-31
Separate InitCtxCmd into its own file.

zhalvorsen:
This is the first of several PRs to move commands into separate files. This effort will simplify the `dpe_instance.rs` file. Currently almost 40% of the code in the `dpe` crate is in `dpe_instance.rs`.


zhalvorsen:
I'm not sure how many PRs we want to do for this. This is the biggest one though so I'll keep it separate.

jhand2 (approved):

PR #56, opened 2023-04-03
Let the platform specify certificate issuer

jhand2:
1. Pass issuer common name when creating a DPE instance
2. Compute issuer serial number from public key

I considered adding a callback for the common name as well, but it's not particularly ergonomic since in `certify_key` we don't know the size of the string.

### DIFF
--- a/crypto/src/openssl.rs
+++ b/crypto/src/openssl.rs
@@ -42,6 +42,13 @@ impl OpensslCrypto {
             AlgLen::Bit384 => Nid::SECP384R1,
         }
     }
+
+    fn get_priv_byes(algs: AlgLen) -> Result<Vec<u8>, CryptoError> {
+        let priv_bytes = vec![0u8; algs.size()];
+        let mut priv_digest = vec![0u8; algs.size()];
+        Self::hash(algs, &priv_bytes, &mut priv_digest)?;
+        Ok(priv_digest)
+    }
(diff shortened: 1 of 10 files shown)


crypto/src/openssl.rs
```diff
@@ -42,6 +42,13 @@ impl OpensslCrypto {
             AlgLen::Bit384 => Nid::SECP384R1,
         }
     }
+
+    fn get_priv_byes(algs: AlgLen) -> Result<Vec<u8>, CryptoError> {
+        let priv_bytes = vec![0u8; algs.size()];
+        let mut priv_digest = vec![0u8; algs.size()];
+        Self::hash(algs, priv_bytes.as_slice(), &mut priv_digest.as_mut_slice())?;
```
zhalvorsen, line 49: nit:

```suggestion
        Self::hash(algs, &priv_bytes, &mut priv_digest)?;
```

zhalvorsen (approved):

crypto/src/openssl.rs
jhand2: Done

jhand2 (commented):
Also updated the `Name` struct to just own the buffer, since CertifyKey was creating temporary buffers anyway

zhalvorsen (approved):

PR #58, opened 2023-04-04
Implement GetTaggedTCI and test it with TagTCI

chrisfenner:
This PR introduces an implementation of GetTaggedTci, and an integration test that tests it together with TagTci. Like CertifyKey, the test will be a lot more interesting after DeriveChild is complete.

As a small housekeeping item, this PR also introduces a type for context handle ([16]byte). Creating a type for this will keep the semantics a little clearer as the codebase grows.

### DIFF
--- a/dpe/src/commands/get_tagged_tci.rs
+++ b/dpe/src/commands/get_tagged_tci.rs
@@ -29,7 +29,7 @@ impl TryFrom<&[u8]> for GetTaggedTciCmd {
 }
 
 impl<C: Crypto> CommandExecution<C> for GetTaggedTciCmd {
-    fn execute(&self, dpe: &mut DpeInstance<C>, locality: u32) -> Result<Response, DpeErrorCode> {
+    fn execute(&self, dpe: &mut DpeInstance<C>, _: u32) -> Result<Response, DpeErrorCode> {
         // Make sure this command is supported.
         if !dpe.support.tagging {
             return Err(DpeErrorCode::InvalidCommand);
@@ -42,11 +42,6 @@ impl<C: Crypto> CommandExecution<C> for GetTaggedTciCmd {
             .iter()
             .find(|c| c.has_tag && c.tag == self.tag)
             .ok_or(DpeErrorCode::BadTag)?;
-
-        // Make sure the command is coming from the locality that owns the found context.
-        if ctx.locality != locality {
-            return Err(DpeErrorCode::InvalidHandle);
-        }
 
         Ok(Response::GetTaggedTci(GetTaggedTciResp {
             tci_cumulative: ctx.tci.tci_cumulative,
--- a/dpe/src/commands/get_tagged_tci.rs
+++ b/dpe/src/commands/get_tagged_tci.rs
@@ -1,0 +1,56 @@
+// Licensed under the Apache-2.0 license.
+use super::CommandExecution;
+use crate::{
+    dpe_instance::DpeInstance,
+    response::{DpeErrorCode, GetTaggedTciResp, Response},
+};
+use core::mem::size_of;
+use crypto::Crypto;
+
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


dpe/src/commands/get_tagged_tci.rs
```diff
@@ -0,0 +1,56 @@
...
+        let ctx = dpe
+            .contexts
+            .iter()
+            .find(|c| c.has_tag && c.tag == self.tag)
+            .ok_or(DpeErrorCode::BadTag)?;
+
+        // Make sure the command is coming from the locality that owns the found context.
+        if ctx.locality != locality {
+            return Err(DpeErrorCode::InvalidHandle);
+        }
```
jhand2, line 49: Callers should be allows to get tagged TCIs for any locality, not just their own.

The main use-case for tagged TCIs or to let a single component (like an SPDM responder) collect all the measurements for the SoC.

dpe/src/dpe_instance.rs
```diff
@@ -192,7 +193,7 @@ impl<C: Crypto> DpeInstance<'_, C> {
 }
 
 #[repr(transparent)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
jhand2, line 196: I wonder if we should derive these traits for test config only. Just want to avoid the extra code getting linked if someone inadvertently uses these traits in runtime fw.

dpe/src/response.rs
```diff
@@ -192,6 +197,29 @@ impl Default for CertifyKeyResp {
     }
 }
 
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq)]
```
jhand2, line 201: Same as above

verification/verification_test.go
```diff
@@ -250,3 +250,69 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	var ctx ContextHandle
+
+	tagResp, err := client.TagTCI(&TagTCIReq{ContextHandle: ctx, Tag: tag})
+	if err != nil {
+		t.Fatalf("Could not tag TCI: %v", err)
+	}
+
+	if tagResp.NewContextHandle != ctx {
+		t.Errorf("New context handle from TagTCI was %x, expected %x", tagResp.NewContextHandle, ctx)
+	}
```
jhand2, line 295: This is only the expected behavior because the input handle is the default context. It might be good to

* Add a comment stating this
* (optional) Add a check that if the input handle is a random handle, the output is a different random value. But this might make sense as a separate test case that just checks the handles for all commands. Not sure.

zhalvorsen: I was going to say it would be good to add a simulation context to run some extra tests against, but it doesn't really make sense without extending the simulation context's TCI before doing any tests. This sounds like it would be better in a separate PR.

zhalvorsen (commented):
I was just about to push a PR that makes a type for context handles, but in Rust. That was perfect timing.

dpe/src/dpe_instance.rs
chrisfenner: So, I had to add these derived traits to `TciMeasurement` so that the `GetTaggedTciResp` could also derive them. What would be the problem with this code getting linked if someone uses it, exactly?

verification/verification_test.go
chrisfenner: Could this be folded into the TODO at the bottom of this test? Is it really a second TODO?

zhalvorsen: For my comment, it falls in the TODO. So no action for me.

dpe/src/commands/get_tagged_tci.rs
chrisfenner: Aha, thanks. Fixed.

dpe/src/response.rs
chrisfenner: Fixed in latest commit

zhalvorsen (dismissed):

jhand2 (approved):

PR #59, opened 2023-04-04
Introduce a type for handles in Rust DPE library.

zhalvorsen:
This simplifies several aspects about handles: serializing, deserializing, and checks.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -8,7 +8,7 @@ use crate::{
     _set_flag,
     commands::{Command, CommandExecution, InitCtxCmd},
     response::{DpeErrorCode, GetProfileResp, Response},
-    DPE_PROFILE, HANDLE_SIZE, MAX_HANDLES,
+    DPE_PROFILE, MAX_HANDLES,
 };
 use core::marker::PhantomData;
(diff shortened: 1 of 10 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -422,6 +414,48 @@ impl Iterator for FlagsIter {
...
+    const DEFAULT: [u8; Self::SIZE] = [0; Self::SIZE];
+
+    /// Returns the default context handle.
+    pub const fn default() -> ContextHandle {
+        ContextHandle(Self::DEFAULT)
+    }
+
+    /// Whether the handle is the default context handle.
+    pub fn is_default(&self) -> bool {
+        self == &Self::default()
```
jhand2, line 433: nit: Is this doing a copy? Might be better to just compare `self.0` with `DEFAULT`. But not a big deal either way.

jhand2 (approved):

dpe/src/dpe_instance.rs
zhalvorsen: I don't think so... but, I'm not sure. I think your suggestion is more explicit about what it is doing anyway. I changed it.

jhand2 (approved):

PR #61, opened 2023-04-05
Implement `DeriveChild` command.

zhalvorsen:
Because this has a lot of the same functionality as `ExtendTci`, this also moves adding a TCI measurement into its own function.

### DIFF
--- a/dpe/src/commands/derive_child.rs
+++ b/dpe/src/commands/derive_child.rs
@@ -1,8 +1,10 @@
 // Licensed under the Apache-2.0 license.
 use super::CommandExecution;
 use crate::{
-    dpe_instance::{ContextHandle, DpeInstance},
-    response::{DpeErrorCode, Response},
+    dpe_instance::{
+        ActiveContextArgs, ContextHandle, ContextState, ContextType, DpeInstance, TciMeasurement,
+    },
+    response::{DeriveChildResp, DpeErrorCode, Response},
     DPE_PROFILE,
 };
 use core::mem::size_of;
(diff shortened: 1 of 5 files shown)


dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,120 @@ pub struct DeriveChildCmd {
...
+
+        dpe.add_tci_measurement(child_idx, &TciMeasurement(self.data))?;
+
+        let child_handle = if self.is_make_default() {
+            ContextHandle::default()
+        } else {
+            dpe.generate_new_handle()?
+        };
+
+        let reported_parent_handle = if !self.is_retain_parent() {
```
bluegate010, line 109: Optional:

```
let parent = &mut dpe.contexts[parent_idx];

if !self.is_retain_parent() {
  parent.state = ContextState::Retired;
  parent.handle = ContextHandle([0xff; ContextHandle::SIZE]);
} else if !parent.handle.is_default() {
  parent.handle = dpe.generate_new_handle()?;
}
```

zhalvorsen: That is a lot cleaner. I had to keep the indexing for the `else` otherwise it would have 2 mutable references to something in the DPE instance.

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,120 @@ pub struct DeriveChildCmd {
...
 
-    const fn _is_retain_parent(&self) -> bool {
-        self.flags & Self::_RETAIN_PARENT != 0
+    const fn is_retain_parent(&self) -> bool {
+        self.flags & Self::RETAIN_PARENT != 0
     }
 
-    const fn _is_target_is_default(&self) -> bool {
-        self.flags & Self::_TARGET_IS_DEFAULT != 0
+    const fn is_make_default(&self) -> bool {
```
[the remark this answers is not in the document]
zhalvorsen: Done

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
-        self.flags & Self::_TARGET_IS_DEFAULT != 0
+    const fn makes_default(&self) -> bool {
+        self.flags & Self::MAKE_DEFAULT != 0
+    }
+
+    const fn changes_locality(&self) -> bool {
+        self.flags & Self::CHANGE_LOCALITY != 0
+    }
+
+    /// Check if it will be able to create a new default.
```
jhand2, line 49: I think it would helpful to expand on this comment a little bit about what "safe" means.

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
+
+        let child_handle = if self.makes_default() {
+            ContextHandle::default()
+        } else {
+            dpe.generate_new_handle()?
+        };
+
+        let parent = &mut dpe.contexts[parent_idx];
+
+        let reported_parent_handle = if !self.retains_parent() {
```
jhand2, line 111: nit: What does "reported" mean here? Maybe `output_parent_handle`?

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
+            parent.handle
+        } else {
+            dpe.contexts[parent_idx].handle = dpe.generate_new_handle()?;
+            dpe.contexts[parent_idx].handle
+        };
+
+        let child = &mut dpe.contexts[child_idx];
+        child.activate(ContextType::Normal, target_locality, &child_handle);
+        child.tci.tci_type = self.tci_type;
+        child.parent_idx = parent_idx as u8;
```
jhand2, line 125: Should `activate` take these as parameters to make sure they get set whenever a node is created?

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
+
+        // Add child to the parent's list of children.
+        dpe.contexts[parent_idx].children |= 1 << child_idx;
```
jhand2, line 128: nit: might be good to have a function on `Context` which sets children instead of directly doing bit manipulations

jhand2 (commented):
Haven't dug deeply into the tests yet, but took a first pass

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
+            }
+        }
+        true
     }
 }
 
 impl<C: Crypto> CommandExecution<C> for DeriveChildCmd {
-    fn execute(&self, _dpe: &mut DpeInstance<C>, _locality: u32) -> Result<Response, DpeErrorCode> {
-        Err(DpeErrorCode::InvalidCommand)
+    fn execute(&self, dpe: &mut DpeInstance<C>, locality: u32) -> Result<Response, DpeErrorCode> {
```
bluegate010, line 66: Would it make sense to call this 'cmd' instead of 'self'? Idk, to me 'self' invokes connotations of an instance of a thing that's longer-lived than a single command invocation. Could just be me, up to you.

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +15,122 @@ pub struct DeriveChildCmd {
...
+        let reported_parent_handle = if !self.retains_parent() {
+            parent.state = ContextState::Retired;
+            // The value doesn't matter because the client should ignore it.
+            ContextHandle([0xff; ContextHandle::SIZE])
+        } else if parent.handle.is_default() {
+            parent.handle
+        } else {
+            dpe.contexts[parent_idx].handle = dpe.generate_new_handle()?;
+            dpe.contexts[parent_idx].handle
+        };
```
bluegate010, line 120: Hm, if we can't use `parent` in the final else clause, may as well not use it in the other clauses, for consistency.

Now that I know more about borrow checking and scopes, what about this? (Just trying to avoid a tri-case 'if' statement)

```
if !self.retains_parent() {
  dpe.contexts[parent_idx].state = ContextState::Retired;
  dpe.contexts[parent_idx].handle = ContextHandle([0xff; ContextHandle::SIZE]);
} else if !dpe.contexts[parent_idx].handle.is_default() {
  dpe.contexts[parent_idx].handle = dpe.generate_new_handle()?;
}

let reported_parent_handle = dpe.contexts[parent_idx].handle;
// Or just use dpe.contexts[parent_idx].handle directly down below
```

dpe/src/commands/derive_child.rs
zhalvorsen: That would be a more fitting name probably, but in Rust `self` is a keyword to indicate it is a member function of a variable. Kind of like python.

bluegate010: Ah ack.

jhand2: Putting self as the first parameter is the way you denote it is a member function. If you exclude it, it will be a static function.

dpe/src/commands/derive_child.rs
zhalvorsen: I added some more information. Let me know if you think that is enough.

dpe/src/commands/derive_child.rs
zhalvorsen: Done

dpe/src/commands/derive_child.rs
zhalvorsen: I got worried that there were too many arguments with similar types so I added a struct to give the arguments.

dpe/src/commands/derive_child.rs
zhalvorsen: Done

dpe/src/commands/derive_child.rs
zhalvorsen: Ahh nice! You're right the middle condition is pretty useless.

Done

dpe/src/commands/derive_child.rs
```diff
@@ -15,36 +17,125 @@ pub struct DeriveChildCmd {
...
+
+        if !self.retains_parent() {
+            dpe.contexts[parent_idx].state = ContextState::Retired;
+            dpe.contexts[parent_idx].handle = ContextHandle([0xff; ContextHandle::SIZE]);
+        } else if !dpe.contexts[parent_idx].handle.is_default() {
+            dpe.contexts[parent_idx].handle = dpe.generate_new_handle()?;
+        }
+
+        let child = &mut dpe.contexts[child_idx];
```
bluegate010, line 122: Optional: used only once, could omit.

zhalvorsen: Done

dpe/src/commands/derive_child.rs
jhand2: That looks good, thanks!

dpe/src/dpe_instance.rs
```diff
@@ -753,6 +810,51 @@ pub mod tests {
...
+        // Verify bounds checking.
+        assert_eq!(
+            Err(DpeErrorCode::InternalError),
+            dpe.add_tci_measurement(MAX_HANDLES, &TciMeasurement::default())
+        );
+
+        let md = match DPE_PROFILE {
+            DpeProfile::P256Sha256 => openssl::hash::MessageDigest::sha256(),
+            DpeProfile::P384Sha384 => openssl::hash::MessageDigest::sha384(),
+        };
```
jhand2, line 828: nit: Could use OpensslCrypto hasher instead, just because it plays nicer with the DPE profile functions

dpe/src/dpe_instance.rs
```diff
@@ -202,6 +202,33 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
+
+    pub(crate) fn add_tci_measurement(
+        &mut self,
+        idx: usize,
+        measurement: &TciMeasurement,
+    ) -> Result<(), DpeErrorCode> {
+        if idx >= MAX_HANDLES {
+            return Err(DpeErrorCode::InternalError);
+        }
+        let context = &mut self.contexts[idx];
```
jhand2, line 214: It feels a little bit error prone that we aren't checking locality, active status, etc. I know the expectation is that the caller will check this, I'm just worried this function encourages blindly adding a measurement.

dpe/src/dpe_instance.rs
zhalvorsen: I originally thought to use plain OpenSSL to prove we were using the library correctly, but I think using `OpensslCrypto` is enough proof.

Done

dpe/src/dpe_instance.rs
zhalvorsen: I could see how this could be abused if we weren't careful. The change did remind me there is a conversation I thought we should have about error handling.

I had originally placed the `add_tci_measurement` call higher because it was one of the last things that could fail before we made permanent changes to the state of the contexts. The system could have some undefined behaviors if failures happen late in any of the commands.

For example, in `DeriveChild` if the child is activated, added the parent, and then an error happens while adding the TCI measurement. It would return an error without cleaning anything up or undoing any of the partially executed work. There would now be a dangling context and nothing has the handle. This particular scenario might be really rare, but other may have the higher likelihoods.

We may want to copy out the context we are going to manipulate, work on it, and then commit it back to the DPE instance if everything passes. Do you think it is worth worrying about?

jhand2: As discussed offline, agreed this is a problem that we should fix. Might be good to open an issue to track.

zhalvorsen: Added #63 to track the issue.

PR #64, opened 2023-04-07
Add an iterator for traversing a chain of nodes from child to root.

zhalvorsen:
This also adds a common function to derive a CDI from a child context which will be useful in the `Sign` command.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -169,20 +169,19 @@ impl<C: Crypto> DpeInstance<'_, C> {
         Ok(())
     }
 
-    /// Get the TCI nodes from `context` to the root node following parent
+    /// Get the TCI nodes from the context at `start_idx` to the root node following parent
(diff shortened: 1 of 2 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -176,13 +176,12 @@ impl<C: Crypto> DpeInstance<'_, C> {
     /// Returns the number of TCIs written to `nodes`
     pub(crate) fn get_tcb_nodes(
         &self,
-        context: &Context,
+        start_idx: usize,
```
jhand2, line 179: nit: change the comment from `context` to `start_idx`

jhand2 (approved):

dpe/src/dpe_instance.rs
zhalvorsen: Done

jhand2 (approved):

PR #66, opened 2023-04-08
Add error cases to `ChildToRootIter`.

zhalvorsen:


dpe/src/context.rs
```diff
@@ -188,32 +202,143 @@ pub mod tests {
     fn test_child_to_root_iter() {
         const INITIALIZER_CONTEXT: Context = Context::new();
         let mut contexts = [INITIALIZER_CONTEXT; MAX_HANDLES];
-        let chain_indeces = [2, 4, 1, 13, MAX_HANDLES - 1, 3, 0, 9];
+        let root_index = CHAIN_INDECES[0];
```
bluegate010, line 205: Nit: indices

zhalvorsen: Done

dpe/src/context.rs
```diff
@@ -150,70 +150,173 @@ pub(crate) struct ChildToRootIter<'a> {
...
-    type Item = &'a Context;
+    type Item = Result<&'a Context, DpeErrorCode>;
 
-    fn next(&mut self) -> Option<&'a Context> {
+    fn next(&mut self) -> Option<Result<&'a Context, DpeErrorCode>> {
         if self.done {
             return None;
         }
+        if self.count >= MAX_HANDLES {
+            return Some(Err(DpeErrorCode::MaxTcis));
```
ia0, line 176: Which semantics do we want? Currently we repeatedly and indefinitely return the same error. An alternative would be to `self.done = false` before returning the first error, such that iteration stops after the first error.

It looks to me that finite iteration is "safer" in the sense that one cannot inadvertently "freeze" execution because they forgot to interrupt iteration on error.

dpe/src/context.rs
```diff
@@ -150,70 +150,173 @@ pub(crate) struct ChildToRootIter<'a> {
...
+        }
 
         let context = &self.contexts[self.idx];
+
+        // Check if context is valid.
+        const MAX_IDX: u8 = (MAX_HANDLES - 1) as u8;
+        let valid_parent_idx = matches!(context.parent_idx, 0..=MAX_IDX | Context::ROOT_INDEX);
+        if !valid_parent_idx || context.state == ContextState::Inactive {
+            return Some(Err(DpeErrorCode::InternalError));
```
ia0, line 185: Same question here regarding infinite iteration.

dpe/src/context.rs
ia0: No I mean marking the iterator as done such that if the user continues to call next after an error, then iteration would stop instead of infinitely returning the same error which would most probably mean an infinite loop.

zhalvorsen: Ahhh, yeah, got it. That is a good idea! I just added it.

dpe/src/context.rs
zhalvorsen: Done

ia0: This doesn't look done to me. I don't see any mutation before returning which means another call to next should return the same error.

zhalvorsen: Oh weird somehow that line didn't get added to my commit. Should be good now.

ia0 (dismissed):

dpe/src/context.rs
```diff
@@ -150,70 +150,175 @@ pub(crate) struct ChildToRootIter<'a> {
...
+        let mut iter = ChildToRootIter::new(0, &contexts);
+        assert!(iter.next().unwrap().is_ok());
+
+        // Root index.
+        contexts[0].parent_idx = Context::ROOT_INDEX as u8;
+        let mut iter = ChildToRootIter::new(0, &contexts);
+        assert!(iter.next().unwrap().is_ok());
+    }
+
+    const CHAIN_INDICES: [usize; MAX_HANDLES] = [
```
jhand2, line 298: Could you add a comment here about what this chain of indices is meant to test?

dpe/src/context.rs
```diff
@@ -150,70 +150,175 @@ pub(crate) struct ChildToRootIter<'a> {
...
+        if self.count >= MAX_HANDLES {
+            self.done = true;
+            return Some(Err(DpeErrorCode::MaxTcis));
+        }
 
         let context = &self.contexts[self.idx];
+
+        // Check if context is valid.
+        const MAX_IDX: u8 = (MAX_HANDLES - 1) as u8;
+        let valid_parent_idx = matches!(context.parent_idx, 0..=MAX_IDX | Context::ROOT_INDEX);
```
jhand2, line 184: Sorry, rust novice question. I can't quite figure out what this means. Is it:

`if parent_idx is in the range (0, MAX_IDX] or parent_idx is the ROOT_INDEX`?

zhalvorsen: Really close, it is inclusive for the left part of the range as well. i.e.

```
if parent_idx is in the range [0, MAX_IDX] or parent_idx is the ROOT_INDEX
```

ia0: Yes (modulo that 0 is also inclusive). Essentially, `matches!(expr, pattern)` is a sugar for:

```rust
match expr {
    pattern => true,
    _ => false,
}
```

zhalvorsen: To expand this macro as @ia0 mentioned, this would read:

```Rust
let valid_parent_idx = match context.parent_idx {
    0..=MAX_IDX | Context::ROOT_INDEX => true,
    _ => false,
};
```

For the pipe this would expand it a little further to look something like this:

```Rust
let valid_parent_idx = match context.parent_idx {
    0..=MAX_IDX => true,
    Context::ROOT_INDEX => true,
    _ => false,
};
```

dpe/src/context.rs
zhalvorsen: Added a comment with some more description of how it is used.

dpe/src/context.rs
jhand2: Thanks!

jhand2 (approved):

PR #71, opened 2023-04-11
Implement `Sign` command.

zhalvorsen:


dpe/src/commands/sign.rs
```diff
@@ -19,6 +19,20 @@ pub struct SignCmd {
...
+    const _ND_DERIVATION: u32 = 1 << 30;
+
+    const fn _uses_symmetric(&self) -> bool {
+        self.flags & Self::_IS_SYMMETRIC != 0
+    }
+
+    /// Uses non-deterministic derivation. If symmetric algorithms are used, this flag is ignored.
+    const fn _uses_nd_derivation(&self) -> bool {
+        !self._uses_symmetric() && self.flags & Self::_ND_DERIVATION != 0
+    }
```
jhand2, line 33: Maybe we should open issues to remember to implement these

dpe/src/commands/sign.rs
```diff
@@ -118,4 +169,130 @@ mod tests {
...
+            Response::Sign(resp) => resp,
+            _ => panic!("Incorrect response type"),
+        };
+
+        let sig = EcdsaSig::from_private_components(
+            BigNum::from_slice(&resp.sig_r_or_hmac).unwrap(),
+            BigNum::from_slice(&resp.sig_s).unwrap(),
+        )
+        .unwrap();
+        assert!(sig.verify(&TEST_DIGEST, &priv_key).unwrap());
```
jhand2, line 296: If it's not too much work, we should try to get the public key from CertifyKey and verify the signature with that to make sure the right public key is being output.

Could also do that test in the validation tests since it's more of a cross-command integration test.

zhalvorsen: Done

jhand2 (approved):

PR #76, opened 2023-04-28
Use `zerocopy` crate for de-serializing. 

zhalvorsen:
Changes to use crate instead of manually de-serializing structs.

### DIFF
--- a/dpe/src/commands/mod.rs
+++ b/dpe/src/commands/mod.rs
@@ -23,6 +23,7 @@ use crate::{
 };
 use core::mem::size_of;
 use crypto::Crypto;
+use zerocopy::FromBytes;
 
 mod certify_key;
 mod derive_child;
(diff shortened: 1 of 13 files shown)


dpe/src/commands/mod.rs
```diff
@@ -67,26 +68,56 @@ impl Command {
...
-        let cmd_header = CommandHdr::try_from(bytes)?;
+        let header = CommandHdr::read_from_prefix(bytes).ok_or(DpeErrorCode::InvalidArgument)?;
+        if header.magic != CommandHdr::DPE_COMMAND_MAGIC {
+            return Err(DpeErrorCode::InvalidCommand);
+        }
+        // The client doesn't know what profile is implemented when calling the `GetProfile`
+        // command. But, all other commands should be directed towards the correct profile.
+        if header.cmd_id != Command::GET_PROFILE && header.profile != DPE_PROFILE as u32 {
+            return Err(DpeErrorCode::InvalidCommand);
+        }
```
ia0, line 79: Do we believe we need to parse a `CommandHdr` somewhere else than in this `Command::deserialize()` function? If yes and if those magic and profile checks are always expected to occur, then we should probably keep this as a separate function like it was the case before.

dpe/src/commands/mod.rs
```diff
@@ -67,26 +68,56 @@ impl Command {
...
-            Command::DESTROY_CONTEXT => Ok(Command::DestroyCtx(DestroyCtxCmd::try_from(bytes)?)),
-            Command::GET_CERTIFICATE_CHAIN => Err(DpeErrorCode::InvalidCommand),
-            Command::EXTEND_TCI => Ok(Command::ExtendTci(ExtendTciCmd::try_from(bytes)?)),
-            Command::TAG_TCI => Ok(Command::TagTci(TagTciCmd::try_from(bytes)?)),
-            Command::GET_TAGGED_TCI => Ok(Command::GetTaggedTci(GetTaggedTciCmd::try_from(bytes)?)),
             _ => Err(DpeErrorCode::InvalidCommand),
         }
     }
+
+    fn deserialize_args(header: &CommandHdr, bytes: &[u8]) -> Option<Command> {
```
ia0, line 100: I suspect you had to introduce this function because `read_from_prefix()` returns an `Option` and you wanted to factorize the `.ok_or(DpeErrorCode::InvalidArgument)`. But by doing so you also duplicate the list of supported commands. Another alternative would be to introduce a polymorphic helper function:

```rust
fn parse_command<T: FromBytes>(
    build: impl FnOnce(T) -> Command,
    bytes: &[u8],
) -> Result<Command, DpeErrorCode> {
    Ok(build(T::read_from_prefix(bytes).ok_or(DpeErrorCode::InvalidArgument)?))
}

match header.cmd_id {
    Command::GET_PROFILE => Ok(Command::GetProfile),
    Command::INITIALIZE_CONTEXT => parse_command(Command::InitCtx, bytes),
    ...
    _ => Err(DpeErrorCode::InvalidCommand),
}
```

I didn't test if it compile, but I suspect something along those lines to work.

zhalvorsen: Great idea! That cleans it up a lot.

dpe/src/commands/mod.rs
zhalvorsen: I can't think of another place, but it also fits a lot better in the original location. I put the whole thing back for `CommandHdr` including taking out the derive. I didn't want a similar scenario where someone didn't realize there was special logic in `CommandHdr::try_from` that wouldn't be in `CommandHdr::read_from_prefix` and misuse it.

ia0: I'm not convinced about the "taking out the derive" part. Why not have a `TryFrom` implementation that uses the derive?

```rust
impl TryFrom<&[u8]> for CommandHdr {
    type Error = DpeErrorCode;

    fn try_from(raw: &[u8]) -> Result<Self, Self::Error> {
        let header = CommandHdr::read_from_prefix(raw).map_err(|_| DpeErrorCode::InvalidCommand)?;
        if header.magic != Self::DPE_COMMAND_MAGIC {
            return Err(DpeErrorCode::InvalidCommand);
        }
        // The client doesn't know what profile is implemented when calling the `GetProfile`
        // command. But, all other commands should be directed towards the correct profile.
        if header.cmd_id != Command::GET_PROFILE && header.profile != DPE_PROFILE as u32 {
            return Err(DpeErrorCode::InvalidCommand);
        }
        Ok(header)
    }
}
```

zhalvorsen: My concern is that someone later on could use `read_from_prefix` elsewhere and not realize they were missing key checks that exist in `try_from`. Ideally, if we could make the `derive` private that would solve the concern.

I don't think its too major of a concern though, I put in your suggestion.

zhalvorsen: I think (1) is good. It's pretty low risk so its probably not worth making too complicated of a solution. We can always change it later if we need to.

jhand2 (approved):

Issue #38, opened 2023-03-23
Change the RotateContext command to fail if the locality already has a default context.


zhalvorsen:
The spec says for [RotateContext](https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.md#rotatecontexthandle) it says:

> If TARGET_IS_DEFAULT is set, NEW_CONTEXT_HANDLE SHALL be made the default in the caller's locality. If the default context handle is already defined in the caller's locality, this command SHALL fail.

Right now there is no concept of a default context per locality. This needs to be added along with this check for the command.

PR #77, opened 2023-05-01
Fail RotateContext if caller's locality already has default context

sree-revoori1:
See issue https://github.com/chipsalliance/caliptra-dpe/issues/38 for more details.

### DIFF
--- a/dpe/src/commands/rotate_context.rs
+++ b/dpe/src/commands/rotate_context.rs
@@ -67,6 +67,7 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
             }
         }
 
+        let new_handle = dpe.generate_new_handle()?;
         dpe.contexts[idx].handle = new_handle;
         Ok(Response::RotateCtx(NewHandleResp { handle: new_handle }))
     }
--- a/dpe/src/commands/rotate_context.rs
+++ b/dpe/src/commands/rotate_context.rs
@@ -55,7 +55,7 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
 
         // Make sure the command is coming from the right locality.
         if dpe.contexts[idx].locality != locality {
-            return Err(DpeErrorCode::InvalidLocality);
+            return Err(DpeErrorCode::InvalidHandle);
         }
         
         // Make sure caller's locality does not already have a default context.
@@ -67,7 +67,6 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
             }
         }
 
-        let new_handle = dpe.generate_new_handle()?;
         dpe.contexts[idx].handle = new_handle;
         Ok(Response::RotateCtx(NewHandleResp { handle: new_handle }))
     }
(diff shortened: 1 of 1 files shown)


dpe/src/commands/rotate_context.rs
```diff
@@ -47,7 +47,14 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
...
         if dpe.contexts[idx].locality != locality {
-            return Err(DpeErrorCode::InvalidHandle);
+            return Err(DpeErrorCode::InvalidLocality);
+        }
+        
+        // Make sure caller's locality does not already have a default context.
+        let default_context_idx =
+                dpe.get_active_context_pos(&ContextHandle::default(), locality);
+        if default_context_idx.is_some() {
+            return Err(DpeErrorCode::InvalidArgument);
```
zhalvorsen, line 57: We only want to do this check if the caller is trying to make this context a default context. The caller will make that type of request with the `TARGET_IS_DEFAULT` flag in the `flags` field.

If you want an example of how we have been parsing flags so far, the `DeriveChild` command in `dpe/src/commands/derive_child.rs` has a similar flag.

zhalvorsen (commented):
Great job! It looks like you are getting a handle on the code base.

dpe/src/commands/rotate_context.rs
sree-revoori1: Done.

jhand2 (approved):

dpe/src/commands/rotate_context.rs
```diff
@@ -47,7 +55,16 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
...
+        // Make sure caller's locality does not already have a default context.
+        if self.uses_target_is_default() {
+            let default_context_idx =
+                dpe.get_active_context_pos(&ContextHandle::default(), locality);
+            if default_context_idx.is_some() {
+                return Err(DpeErrorCode::InvalidArgument);   
+            }
         }
 
         let new_handle = dpe.generate_new_handle()?;
```
zhalvorsen, line 70: Not sure we want it as part of this PR, but if the request is to make it default, it should assign the `new_handle` to be all zeroes (which indicates a default context).  We should probably add a unit test for this as well.
```suggestion
        let new_handle = if self.uses_target_is_default() {
            dpe.generate_new_handle()?
        } else {
            ContextHandle::default()
        };
```

dpe/src/commands/rotate_context.rs
```diff
@@ -17,6 +17,14 @@ pub struct RotateCtxCmd {
     target_locality: u32,
 }
 
+impl RotateCtxCmd {
+    pub const TARGET_IS_DEFAULT: u32 = 1 << 31;
+    
+    const fn uses_target_is_default(&self) -> bool {
```
zhalvorsen, line 23: optional: maybe a more concise name would be `makes_default`

zhalvorsen: Or `make_default`

sree-revoori1: Reason I did uses_target_is_default is to maintain consistency with flags in derive_child since we are using similar naming schemes there.

dpe/src/commands/rotate_context.rs
sree-revoori1: Thanks, will do in followup commit

jhand2 (approved):

PR #79, opened 2023-05-01
Make new handle the default handle if TARGET_IS_DEFAULT is set in Rot…

sree-revoori1:
…ateContext

Currently, we always generate a random new handle in rotate context, but new handle should be a buffer of 0s if the TARGET_IS_DEFAULT flag is set.

### DIFF
--- a/dpe/src/commands/rotate_context.rs
+++ b/dpe/src/commands/rotate_context.rs
@@ -67,7 +67,11 @@ impl<C: Crypto> CommandExecution<C> for RotateCtxCmd {
             }
         }
 
-        let new_handle = dpe.generate_new_handle()?;
+        let new_handle = if self.uses_target_is_default() {
+            ContextHandle::default()
+        } else {
+            dpe.generate_new_handle()?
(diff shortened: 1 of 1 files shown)


zhalvorsen (dismissed):

dpe/src/commands/rotate_context.rs
```diff
@@ -202,5 +206,18 @@ mod tests {
...
             .execute(&mut dpe, TEST_LOCALITIES[0])
         );
+        
+        // New handle is all 0s if caller requests default handle
+        assert_eq!(
+            Ok(Response::RotateCtx(NewHandleResp {
+                handle: ContextHandle::default()
+            })),
+            RotateCtxCmd {
+                handle: ContextHandle::default(),
```
zhalvorsen, line 216:
```suggestion
                handle: SIMULATION_HANDLE,
```

zhalvorsen (dismissed):

jhand2 (approved):

PR #81, opened 2023-05-02
Add base struct to simplify Crypto library.

zhalvorsen:
This adds a base struct for handling all crypto objects (digests, keys, signatures). This simplifies the `Crypto` API and removes some of the arguments that could be easy to mix up. This new struct will help streamline changes needed for symmetric signing.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -6,21 +6,25 @@ Abstract:
 
 #[cfg(feature = "openssl")]
 pub use crate::openssl::*;
+pub use signer::*;
 
 #[cfg(feature = "openssl")]
 pub mod openssl;
 
+mod signer;
 use core::fmt::{Error, Write};
 
 #[derive(Debug, Clone, Copy)]
+#[cfg_attr(test, derive(strum_macros::EnumIter))]
 pub enum AlgLen {
     Bit256,
     Bit384,
+    // NOTE: If a larger length is added, MUST update AlgLen::MAX_ALG_LEN
 }
 
-const MAX_HASH_SIZE: usize = 384 / 8;
-
 impl AlgLen {
+    const MAX_ALG_LEN: Self = Self::Bit384;
+    pub(crate) const MAX_ALG_LEN_BYTES: usize = Self::MAX_ALG_LEN.size();
     pub const fn size(self) -> usize {
         match self {
             AlgLen::Bit256 => 256 / 8,
@@ -33,6 +37,7 @@ impl AlgLen {
 pub enum CryptoError {
     AbstractionLayer,
     CryptoLibError,
+    Size,
 }
 
 pub trait Hasher: Sized {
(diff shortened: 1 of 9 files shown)


crypto/src/signer.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/// An HMAC Signature
+pub type HmacSig = CryptoBlob;
+
+/// An HMAC Key
+pub type HmacKey = CryptoBlob;
+
+/// A common base struct that can be used for all digests, signatures, and keys.
+pub struct CryptoBlob {
```
jhand2, line 42: minor nit: I think I prefer CryptoBuf. Blob implies to me that the caller shouldn't interpret the data

zhalvorsen: I like that name better too.

Done.

jhand2 (approved):

ia0 (approved):
I'll try to recap what this PR does and its implications, to make sure nothing implicit stays hidden and because I'm not able to assess whether the new state is appropriate for the project usage:
1. The `CryptoBuf` type is introduced as a length-erased byte array. Implications:
    - All instances of `CryptoBuf` pay for the longest possible array. Not an issue if the length distribution is anyway skewed favoring long arrays.
    - Length checks are dynamic, in particular in the `bytes()` method and when converting to actual array with `try_from`. Not an issue if the compiler statically figures out the length.
3. Crypto functions now return `CryptoBuf` objects instead of taking mutable byte slices as argument. Implications:
    - May incur additional copies. Not an issue if those copies are negligible to the actual crypto operation producing them.
    - May complicate zeroization requiring one zeroization per copy. Not an issue if zeroization is automated (which it probably should) or not needed.

jhand2:
> May incur additional copies. Not an issue if those copies are negligible to the actual crypto operation producing them.

I think our constraint worries are in the following order:

1. code size
2. Asymptotic performance gains (i.e. if something is very prohibitively slow)
3. Code readability
4. Constant-time performance gains (such as reducing from 2 or 3 copies to 1)

That's not to say 4 isn't important, but that I think the readability gain is justified.

ia0:
> * code size

I think it's also useful to distinguish between asymptotic and "constant"  footprint gains similarly to how performance gains are distinguished. Because I guess here what matters is that "constant" footprint gains are more important than asymptotic performance gains (or between that and code readability).

A good rule of thumb for footprint is to avoid polymorphic functions that don't get inlined (so usually big chunks of polymorphic code or traits that are implemented more than once in the same binary). The solution against that is some form of type erasure, which this PR is indeed doing by introducing this length-erased array type.

crypto/src/lib.rs
```diff
@@ -228,3 +213,20 @@ impl BufWriter<'_> {
...
+
+    #[test]
+    fn test_max_alg_len_size() {
+        let mut max_len = 0;
+
+        for alg_len in AlgLen::iter() {
+            max_len = max_len.max(alg_len.size());
+        }
+
+        assert_eq!(AlgLen::MAX_ALG_LEN_BYTES, max_len);
```
ia0, line 230: fyi: Could also be

```rust
let max_len = AlgLen::iter().map(|x| x.size()).max().unwrap();
assert_eq!(AlgLen::MAX_ALG_LEN_BYTES, max_len);
```

Or

```rust
assert!(AlgLen::iter().all(|x| x.size() <= AlgLen::MAX_ALG_LEN_BYTES));
```

zhalvorsen: Good thinking

ia0 (commented):
LGTM

jhand2 (approved):

PR #82, opened 2023-05-07
Add Non-deterministic key derivation mode

sree-revoori1:
This commit introduces non-deterministic key derivation mode. See more details about the specification here -
https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.md#non-deterministic-mode-key-derivation.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -121,35 +121,38 @@ pub trait Crypto {
     /// * `algs` - Which length of algorithm to use.
     fn hash_initialize(algs: AlgLen) -> Result<Self::Hasher, CryptoError>;
 
-    /// Derive a CDI based on the current base CDI and measurements.
+    /// Derive a CDI based on the current base CDI and measurements as well as an optional random seed value.
     ///
     /// # Arguments
     ///
     /// * `algs` - Which length of algorithms to use.
     /// * `measurement` - A digest of the measurements which should be used for CDI derivation
     /// * `info` - Caller-supplied info string to use in CDI derivation
+    /// * `rand_seed` - Caller-supplied optional random seed to use in CDI derivation
     fn derive_cdi(
         algs: AlgLen,
         measurement: &Digest,
         info: &[u8],
+        rand_seed: Option<&[u8]>,
     ) -> Result<Self::Cdi, CryptoError>;
 
-    /// Derives an ECDSA keypair from `cdi` and returns the public key
+    /// Derives a private key using the cdi
     ///
     /// # Arguments
     ///
     /// * `algs` - Which length of algorithms to use.
-    /// * `cdi` - CDI from which to derive the signing key
-    /// * `label` - Caller-supplied label to use in asymmetric key derivation
-    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
+    /// * `cdi` - Caller-supplied private key to use in public key derivation
     ///
+    fn derive_ecdsa_key(algs: AlgLen, cdi: &Self::Cdi, label: &[u8], info: &[u8]) -> EcdsaPriv;
+
+    /// Derives and returns an ECDSA public key using the caller-supplied private key
+    ///
+    /// # Arguments
+    ///
+    /// * `algs` - Which length of algorithms to use.
+    /// * `priv_key` - Caller-supplied private key to use in public key derivation
     /// Returns a derived public key
-    fn derive_ecdsa_pub(
-        algs: AlgLen,
-        cdi: &Self::Cdi,
-        label: &[u8],
-        info: &[u8],
-    ) -> Result<EcdsaPub, CryptoError>;
+    fn derive_ecdsa_pub(algs: AlgLen, priv_key: &EcdsaPriv) -> Result<EcdsaPub, CryptoError>;
 
     /// Sign `digest` with the platform Alias Key
     ///
@@ -159,21 +162,17 @@ pub trait Crypto {
     /// * `digest` - Digest of data to be signed.
     fn ecdsa_sign_with_alias(algs: AlgLen, digest: &Digest) -> Result<EcdsaSig, CryptoError>;
 
-    /// Sign `digest` with a derived key from the CDI.
+    /// Sign `digest` with a derived key-pair from the CDI and caller-supplied private key
     ///
     /// # Arguments
     ///
     /// * `algs` - Which length of algorithms to use.
-    /// * `cdi` - CDI from which to derive the signing key
-    /// * `label` - Caller-supplied label to use in asymmetric key derivation
-    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
     /// * `digest` - Digest of data to be signed.
+    /// * `priv_key` - Caller-supplied private key to use in public key derivation
     fn ecdsa_sign_with_derived(
         algs: AlgLen,
-        cdi: &Self::Cdi,
-        label: &[u8],
-        info: &[u8],
         digest: &Digest,
+        priv_key: &EcdsaPriv,
     ) -> Result<EcdsaSig, CryptoError>;
 
     /// Compute the serial number string for the alias public key
--- a/crypto/src/openssl.rs
+++ b/crypto/src/openssl.rs
@@ -1,10 +1,11 @@
 // Licensed under the Apache-2.0 license
 
-use crate::{AlgLen, Crypto, CryptoBuf, CryptoError, Digest, EcdsaPub, Hasher};
+use crate::{AlgLen, Crypto, CryptoBuf, CryptoError, Digest, EcdsaPriv, EcdsaPub, Hasher};
 use openssl::{
     bn::{BigNum, BigNumContext},
     ec::{EcGroup, EcKey, EcPoint},
     ecdsa::EcdsaSig,
+    error::ErrorStack,
     hash::MessageDigest,
     nid::Nid,
     pkey::Private,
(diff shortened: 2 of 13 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -19,21 +19,48 @@ pub struct CertifyKeyCmd {
...
+        if context.locality != locality {
             return Err(DpeErrorCode::InvalidHandle);
         }
 
-        // Derive CDI and public key
-        let cdi = dpe.derive_cdi(idx)?;
-        let pub_key = C::derive_ecdsa_pub(DPE_PROFILE.alg_len(), &cdi, &self.label, b"ECC")
-            .map_err(|_| DpeErrorCode::InternalError)?;
+        let algs = DPE_PROFILE.alg_len();
+        let pub_key = if context.cached_priv_key.is_some() {
```
zhalvorsen, line 49: Should we only take the cached key if this request is also non-deterministic? In other words

```Rust
if self.uses_nd_derivation && context.cached_priv_key.is_some()
```

dpe/src/commands/sign.rs
```diff
@@ -245,4 +278,92 @@ mod tests {
...
+                tci_type: i as u32,
+                target_locality: 0,
+            }
+            .execute(&mut dpe, TEST_LOCALITIES[0])
+            .unwrap();
+        }
+
+        let sig = {
+            // Check that private key is not cached and hence must be generated by sign
+            let idx = dpe
```
zhalvorsen, line 308: fyi: Can pull this `idx` assignment out one layer so it can be used later.

dpe/src/support.rs
```diff
@@ -149,7 +175,7 @@ pub mod test {
         .get_flags();
         assert_eq!(
             flags,
-            (1 << 31) | (1 << 29) | (1 << 27) | (1 << 25) | (1 << 23)
+            (1 << 31) | (1 << 29) | (1 << 27) | (1 << 25) | (1 << 21)
```
zhalvorsen, line 178: fyi: this check was intended to check all of the odd-numbered bit flags.

dpe/src/support.rs
```diff
@@ -159,7 +185,7 @@ pub mod test {
             ..Support::default()
         }
         .get_flags();
-        assert_eq!(flags, (1 << 30) | (1 << 28) | (1 << 26) | (1 << 24));
+        assert_eq!(flags, (1 << 30) | (1 << 28) | (1 << 26) | (1 << 22));
```
zhalvorsen, line 188: fyi: this check was intended to check all of the even-numbered bit flags.

crypto/src/openssl.rs
```diff
@@ -51,37 +52,21 @@ impl OpensslCrypto {
...
     }
 
-    pub fn derive_ecdsa_key(
+    fn ec_key_from_priv_key(
         algs: AlgLen,
-        cdi: &OpensslCdi,
-        label: &[u8],
-        info: &[u8],
-    ) -> EcKey<Private> {
+        priv_key: EcdsaPriv,
```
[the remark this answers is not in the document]
zhalvorsen: As mentioned offline we should be able to with something similar to this in certify_key and sign:

```Rust
            dpe.contexts[idx].cached_priv_key.take().unwrap_or({
                let cdi = dpe.derive_cdi(idx, true)?;
                C::derive_ecdsa_key(algs, &cdi, &self.label, b"ECC")
            })
        } else {
            C::derive_ecdsa_key(algs, &cdi, &self.label, b"ECC")
        };
        let pub_key = C::derive_ecdsa_pub(DPE_PROFILE.alg_len(), &priv_key)
            .map_err(|_| DpeErrorCode::InternalError)?;
        // cache private key
        if self.uses_nd_derivation() {
            let context = &mut dpe.contexts[idx];
            context.cached_priv_key.replace(priv_key);
        }
```

dpe/src/commands/certify_key.rs
```diff
@@ -19,21 +19,50 @@ pub struct CertifyKeyCmd {
...
+        if context.locality != locality {
             return Err(DpeErrorCode::InvalidHandle);
         }
 
-        // Derive CDI and public key
-        let cdi = dpe.derive_cdi(idx)?;
-        let pub_key = C::derive_ecdsa_pub(DPE_PROFILE.alg_len(), &cdi, &self.label, b"ECC")
+        let algs = DPE_PROFILE.alg_len();
+        let priv_key = if self.uses_nd_derivation() {
+            dpe.contexts[idx].cached_priv_key.take().unwrap_or({
```
ia0, line 50: I'm not sure of the expected behavior here, but we might want to use `unwrap_or_else()`. The difference being that if there's a cached private key, then we won't derive a new key. Right now we always generate a new key but simply discard it if there's a cached private key.

dpe/src/commands/sign.rs
```diff
@@ -20,20 +20,25 @@ pub struct SignCmd {
...
 
     const fn _uses_symmetric(&self) -> bool {
         self.flags & Self::_IS_SYMMETRIC != 0
     }
 
     /// Uses non-deterministic derivation. If symmetric algorithms are used, this flag is ignored.
-    const fn _uses_nd_derivation(&self) -> bool {
-        !self._uses_symmetric() && self.flags & Self::_ND_DERIVATION != 0
+    const fn uses_nd_derivation(&self) -> bool {
+        !self._uses_symmetric() && self.flags & Self::ND_DERIVATION != 0
```
ia0, line 31: nit: Now that this function is used we can also remove the underscore prefix from `uses_symmetric`. But could/should be a separate PR.

dpe/src/commands/sign.rs
```diff
@@ -46,12 +51,26 @@ impl<C: Crypto> CommandExecution<C> for SignCmd {
...
 
-        let cdi = dpe.derive_cdi(idx)?;
         let digest = Digest::new(&self.digest, DPE_PROFILE.alg_len())
             .map_err(|_| DpeErrorCode::InternalError)?;
-        let EcdsaSig { r, s } =
-            C::ecdsa_sign_with_derived(DPE_PROFILE.alg_len(), &cdi, &self.label, b"ECC", &digest)
-                .map_err(|_| DpeErrorCode::InternalError)?;
+        let algs = DPE_PROFILE.alg_len();
+        let priv_key = if self.uses_nd_derivation() {
+            dpe.contexts[idx].cached_priv_key.take().unwrap_or({
```
ia0, line 58: Same comment as previously regarding using `unwrap_or_else()`.

dpe/src/commands/sign.rs
```diff
@@ -46,12 +51,26 @@ impl<C: Crypto> CommandExecution<C> for SignCmd {
...
+            let cdi = dpe.derive_cdi(idx, false)?;
+            C::derive_ecdsa_key(algs, &cdi, &self.label, b"ECC")
+        };
+
+        let EcdsaSig { r, s } = C::ecdsa_sign_with_derived(algs, &digest, &priv_key)
+            .map_err(|_| DpeErrorCode::InternalError)?;
+        // cache private key
+        if self.uses_nd_derivation() {
+            let context = &mut dpe.contexts[idx];
+            context.cached_priv_key.replace(priv_key);
```
ia0, line 72: nit: Not sure introducing a variable helps here. I would expect the same usage as line 58.

```rust
dpe.contexts[idx].cached_priv_key.replace(priv_key);
```

ia0 (commented):
LGTM at Rust-level (modulo the `unwrap_or_else()` thing)

crypto/src/lib.rs
```diff
@@ -159,21 +162,17 @@ pub trait Crypto {
...
-    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
     /// * `digest` - Digest of data to be signed.
+    /// * `priv_key` - Caller-supplied private key to use in public key derivation
     fn ecdsa_sign_with_derived(
         algs: AlgLen,
-        cdi: &Self::Cdi,
-        label: &[u8],
-        info: &[u8],
         digest: &Digest,
+        priv_key: &EcdsaPriv,
```
jhand2, line 175: So the reason this takes a cdi instead of a private key is that in some integrations they actual private key will be held in hardware and firmware will never see the private key.

In the Caliptra implementation of this trait, I think the Cdi will just be the index of a KeyVault slot.

So at least for now I think it's better to rederive the key from the CDI whenever it is needed.

crypto/src/openssl.rs
```diff
@@ -116,37 +101,70 @@ impl Crypto for OpensslCrypto {
...
+                    &KdfArgument::KbMode(KdfKbMode::Counter),
+                    &KdfArgument::Mac(KdfMacType::Hmac(md)),
+                    &KdfArgument::KbInfo(measurement.bytes()),
+                    &KdfArgument::Salt(info),
+                    &KdfArgument::Key(&measurement.bytes()),
+                ];
+                perform_kdf(KdfType::KeyBased, &args, md.size())
+                    .map_err(|_| CryptoError::CryptoLibError)
+            }
+        }
```
jhand2, line 132: nit: could maybe clean this up by doing something like this. I'm not 100% sure if this works though.

```rust
let mut args = vec![
    &KdfArgument::KbMode(KdfKbMode::Counter),
    &KdfArgument::Mac(KdfMacType::Hmac(md)),
    &KdfArgument::KbInfo(measurement.bytes()),
    &KdfArgument::Salt(info),
    &KdfArgument::Key(&measurement.bytes()),
];

if let Some(seed) = rand_seed {
    args.extend(&KdfArgument::KbSeed(seed));
}

perform_kdf(KdfType::KeyBased, args.as_slice(), md.size())
       .map_err(|_| CryptoError::CryptoLibError)
```

crypto/src/openssl.rs
```diff
@@ -116,37 +101,70 @@ impl Crypto for OpensslCrypto {
...
+    }
+
+    fn derive_ecdsa_key(algs: AlgLen, cdi: &Self::Cdi, label: &[u8], info: &[u8]) -> EcdsaPriv {
+        let md = Self::get_digest(algs);
         let args = [
             &KdfArgument::KbMode(KdfKbMode::Counter),
             &KdfArgument::Mac(KdfMacType::Hmac(md)),
-            &KdfArgument::KbInfo(measurement.bytes()),
+            &KdfArgument::KbInfo(label),
```
jhand2, line 140: I'll come back to this, once the first comment is addressed.

jhand2: No, not applicable anymore

crypto/src/openssl.rs
jhand2: Ok, sounds good

jhand2 (approved):

PR #87, opened 2023-05-10
Introduce symmetric key derivation in sign

sree-revoori1:
If the is_symmetric flag is set, we derive a symmetric key and use it for signing.

### DIFF
--- a/crypto/src/openssl.rs
+++ b/crypto/src/openssl.rs
@@ -1,6 +1,6 @@
 // Licensed under the Apache-2.0 license
 
-use crate::{AlgLen, Crypto, CryptoBuf, CryptoError, Digest, EcdsaPriv, EcdsaPub, Hasher};
+use crate::{AlgLen, Crypto, CryptoBuf, CryptoError, Digest, EcdsaPub, Hasher, HmacSig, PrivKey};
 use openssl::{
     bn::{BigNum, BigNumContext},
     ec::{EcGroup, EcKey, EcPoint},
@@ -8,7 +8,8 @@ use openssl::{
     error::ErrorStack,
     hash::MessageDigest,
     nid::Nid,
-    pkey::Private,
+    pkey::{PKey, Private},
+    sign::Signer,
 };
 use openssl_kdf::{perform_kdf, KdfArgument, KdfKbMode, KdfMacType, KdfType};
 
@@ -54,7 +55,7 @@ impl OpensslCrypto {
 
     fn ec_key_from_priv_key(
         algs: AlgLen,
-        priv_key: &EcdsaPriv,
+        priv_key: &PrivKey,
     ) -> Result<EcKey<Private>, ErrorStack> {
         let nid = Self::get_curve(algs);
         let group = EcGroup::from_curve_name(nid).unwrap();
@@ -132,7 +133,7 @@ impl Crypto for OpensslCrypto {
         }
     }
 
-    fn derive_ecdsa_key(algs: AlgLen, cdi: &Self::Cdi, label: &[u8], info: &[u8]) -> EcdsaPriv {
+    fn derive_private_key(algs: AlgLen, cdi: &Self::Cdi, label: &[u8], info: &[u8]) -> PrivKey {
         let md = Self::get_digest(algs);
         let args = [
             &KdfArgument::KbMode(KdfKbMode::Counter),
@@ -153,7 +154,7 @@ impl Crypto for OpensslCrypto {
         CryptoBuf::new(&priv_bn.to_vec_padded(algs.size() as i32).unwrap(), algs).unwrap()
     }
 
-    fn derive_ecdsa_pub(algs: AlgLen, priv_key: &EcdsaPriv) -> Result<EcdsaPub, CryptoError> {
+    fn derive_ecdsa_pub(algs: AlgLen, priv_key: &PrivKey) -> Result<EcdsaPub, CryptoError> {
         let ec_priv_key = OpensslCrypto::ec_key_from_priv_key(algs, priv_key)
             .map_err(|_| CryptoError::CryptoLibError)?;
         let nid = OpensslCrypto::get_curve(algs);
@@ -204,7 +205,7 @@ impl Crypto for OpensslCrypto {
     fn ecdsa_sign_with_derived(
         algs: AlgLen,
         digest: &Digest,
-        priv_key: &EcdsaPriv,
+        priv_key: &PrivKey,
     ) -> Result<super::EcdsaSig, CryptoError> {
         let ec_priv_key = OpensslCrypto::ec_key_from_priv_key(algs, priv_key)
             .map_err(|_| CryptoError::CryptoLibError)?;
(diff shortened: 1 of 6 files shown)


crypto/src/openssl.rs
```diff
@@ -239,4 +240,28 @@ impl Crypto for OpensslCrypto {
...
+            AlgLen::Bit384 => MessageDigest::sha384(),
+        };
+        let mut signer = Signer::new(sha_size, &hmac_key).unwrap();
+        signer.update(digest.bytes()).unwrap();
+        let hmac = signer.sign_to_vec().unwrap();
+
+        let r = CryptoBuf::new(&hmac, algs).unwrap();
+        let s = CryptoBuf::default(algs);
+
+        Ok(super::Signature { r, s })
```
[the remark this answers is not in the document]
zhalvorsen: I agree, I think it makes the `Crypto` API confusing to have the same return type and key type for ECDSA and HMAC. We should have that complexity in the DPE instance.

fyi: there is an alias already for `HmacSig` and `HmacKey` in `signer.rs`

dpe/src/commands/sign.rs
```diff
@@ -54,28 +56,41 @@ impl<C: Crypto> CommandExecution<C> for SignCmd {
...
-                        algs,
-                        &dpe.derive_cdi(idx, true).unwrap(),
-                        &self.label,
-                        b"ECC",
-                    )
-                }
-            })
-        } else {
+
+        let EcdsaSig { r, s } = if self.uses_symmetric() {
```
zhalvorsen, line 60: This block is starting to get pretty complex. Do you think it would be helpful to add a couple local helper functions? They could take care of the key handling internally. It could turn this block into:

```Rust
        let EcdsaSig { r, s } = if !self.uses_symmetric() {
            self.ecdsa_sign(dpe, idx, &digest)?
        } else {
            let r = self.hmac_sign(dpe, idx, &digest)?;
            let s = CryptoBuf::default(DPE_PROFILE.alg_len());
            EcdsaSig { r, s }
        };
```

with these helper functions:

```Rust
    fn ecdsa_sign<C: Crypto>(
        &self,
        dpe: &mut DpeInstance<C>,
        idx: usize,
        digest: &Digest,
    ) -> Result<EcdsaSig, DpeErrorCode> {
        let algs = DPE_PROFILE.alg_len();
            dpe.contexts[idx].cached_priv_key.take().unwrap_or_else({
                || {
                    C::derive_private_key(
                        algs,
                        &dpe.derive_cdi(idx, true).unwrap(),
                        &self.label,
                        b"ECC",
                    )
                }
            })
        } else {
            C::derive_private_key(algs, &cdi, &self.label, b"ECC")
        };

        let sig = C::ecdsa_sign_with_derived(algs, digest, &priv_key)
            .map_err(|_| DpeErrorCode::InternalError)?;

        // cache private key
        if self.uses_nd_derivation() {
            dpe.contexts[idx].cached_priv_key.replace(priv_key);
        };

        Ok(sig)
    }

    fn hmac_sign<C: Crypto>(
        &self,
        dpe: &mut DpeInstance<C>,
        idx: usize,
        digest: &Digest,
    ) -> Result<HmacSig, DpeErrorCode> {
        let algs = DPE_PROFILE.alg_len();
        C::symmetric_sign_with_derived(algs, &cdi, &self.label, b"HMAC", digest)
            .map_err(|_| DpeErrorCode::InternalError)
    }
```

jhand2, line 60: I think this part is still a bit confusing. Might be better to just diverge these paths instead of trying to return the same type from each branch.

dpe/src/commands/sign.rs
```diff
@@ -282,6 +333,45 @@ mod tests {
...
+            DeriveChildCmd {
+                handle: ContextHandle::default(),
+                data: [i; DPE_PROFILE.get_hash_size()],
+                flags: DeriveChildCmd::MAKE_DEFAULT,
+                tci_type: i as u32,
+                target_locality: 0,
+            }
+            .execute(&mut dpe, TEST_LOCALITIES[0])
+            .unwrap();
+        }
```
zhalvorsen, line 358: I might be under thinking this, but I don't think we need to derive any children because it has the auto-initialized context

dpe/src/commands/sign.rs
```diff
@@ -282,6 +333,45 @@ mod tests {
...
+            flags: SignCmd::IS_SYMMETRIC,
+            digest: TEST_DIGEST,
+        };
+        let resp = match cmd.execute(&mut dpe, TEST_LOCALITIES[0]).unwrap() {
+            Response::Sign(resp) => resp,
+            _ => panic!("Incorrect response type"),
+        };
+
+        // Check that we are using HMAC, i.e. s must be a buffer of all 0s
+        assert!(&resp.sig_s.iter().all(|&b| b == 0x0));
```
zhalvorsen, line 372: It would be good to verify the HMAC here too. You'll have to re-derive the HMAC.

crypto/src/signer.rs
```diff
@@ -2,7 +2,8 @@
 
 use crate::{AlgLen, CryptoError};
 
-/// An ECDSA signature
+/// A cryptographic signature representing an ECDSA signature or an HMAC.
```
zhalvorsen, line 5: nit: We should probably leave out the HMAC comment here. It doesn't really apply to the `crypto` library, just the way we use it in `dpe`.

crypto/src/signer.rs
```diff
@@ -32,14 +33,11 @@ impl EcdsaPub {
     }
 }
 
-pub type EcdsaPriv = CryptoBuf;
+pub type PrivKey = CryptoBuf;
```
zhalvorsen, line 36: nit: I'm not convinced this needs to change names, `PrivKey` sounds a little too generic to me.

zhalvorsen: Ahhh cool, sorry I didn't notice that connection. LGTM.

dpe/src/commands/sign.rs
zhalvorsen: If we wanted to make this more clear, I think we have two options with tuples:

1. Use a tuple or `CryptoBuf`:

```Rust
        let (r, s) = if !self.uses_symmetric() {
            let EcdsaSig { r, s } = self.asymmetric_sign(dpe, idx, &digest)?;
            (r, s)
        } else {
            let r = self.symmetric_sign(dpe, idx, &digest)?;
            (r, CryptoBuf::default(DPE_PROFILE.alg_len()))
        };
```

2. use a tuple of `CryptoBuf` and `Option<CryptoBuf>`:

```Rust
        let (r, s) = if !self.uses_symmetric() {
            let EcdsaSig { r, s } = self.asymmetric_sign(dpe, idx, &digest)?;
            (r, Some(s))
        } else {
            let r = self.symmetric_sign(dpe, idx, &digest)?;
            (r, None)
        };

        dpe.roll_onetime_use_handle(idx)?;

        Ok(Response::Sign(SignResp {
            new_context_handle: dpe.contexts[idx].handle,
            sig_r_or_hmac: r.bytes().try_into().unwrap(),
            sig_s: s
                .unwrap_or_else(|| CryptoBuf::default(DPE_PROFILE.alg_len()))
                .bytes()
                .try_into()
                .unwrap(),
        }))
```

jhand2 (approved):

Issue #85, opened 2023-05-09
Make EcdsaPriv references an implementation defined type.


zhalvorsen:
Currently EcdsaPriv is passed directly as a `CryptoBuf` which makes sense in the OpenSSL implementation, but will not be the case for Caliptra. Caliptra will take a key handle and will never see the raw private key directly. We should make the `Crypto` implementation more generic to allow for different values to be passed in place of the private key.

The `Crypto` definition will need to add a new type trait:

```Rust
pub trait Crypto {
    type Cdi;
    type Hasher: Hasher;
    type EcdsaSigner: Signer;
...
```

The `Signer` trait will likely only need one function to implement (`sign`). This might look something like this:

```Rust
pub trait Signer {
    /// Sign `digest`.
    ///
    /// # Arguments
    ///
    /// * `digest` - Digest of data to be signed.
    fn sign(&self, digest: &Digest) -> Result<EcdsaSig, CryptoError>;
}
```

We might also want to make `Signer` more generic to allow the trait to be used with symmetric signing.

zhalvorsen:
@jhand2 and I discussed offline, this would be better if we reference the type directly without using the `Signer` trait. i.e.

```Rust
pub trait Crypto {
    type Cdi;
    type EcdsaPriv;
...
```

Caliptra would use:

```Rust
impl Crypto for CaliptraCrypto {
    type EcdsaPriv = u8;
...
```

And OpenSSL would use:

```Rust
impl Crypto for OpensslCrypto {
    type EcdsaPriv = crate::EcdsaPriv;
```

PR #89, opened 2023-05-11
Make PrivKey an implementation defined type

sree-revoori1:
This change addresses https://github.com/chipsalliance/caliptra-dpe/issues/85. We make PrivKey implementation defined so that different root of trust chips that make use of DPE can define it. For example, Caliptra does not allow the PrivKey to exist directly in the firmware. 

Note: There is a use of unsafe rust in this commit due to Rust not being able to create const values with generic type parameters. In this unsafe usage, we create and write to uninitialized memory.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -16,7 +16,7 @@ use core::{marker::PhantomData, mem::size_of};
 use crypto::{Crypto, Hasher};
 
 pub struct DpeInstance<'a, C: Crypto> {
-    pub(crate) contexts: [Context; MAX_HANDLES],
+    pub(crate) contexts: [Context<C>; MAX_HANDLES],
     pub(crate) support: Support,
     pub(crate) localities: &'a [u32],
 
@@ -37,6 +37,10 @@ impl<C: Crypto> DpeInstance<'_, C> {
     const MAX_NEW_HANDLE_ATTEMPTS: usize = 8;
     pub const AUTO_INIT_LOCALITY: u32 = 0;
 
+    pub(crate) fn new_context_handles() -> [Context<C>; MAX_HANDLES] {
+        core::array::from_fn(|_| Context::new())
+    }
+
     /// Create a new DPE instance.
     ///
     /// # Arguments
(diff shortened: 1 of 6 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -37,6 +40,17 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
 
+    pub(crate) fn new_context_handles() -> [Context<C>; MAX_HANDLES] {
+        unsafe {
+            let mut array: MaybeUninit<[Context<C>; MAX_HANDLES]> = MaybeUninit::uninit();
+            for i in 0..MAX_HANDLES {
+                let arr = &mut *array.as_mut_ptr();
+                arr[i] = Context::new();
+            }
+            array.assume_init()
+        }
```
jhand2, line 51: I don't think I understand why this part is necessary. It seems like `const CONTEXT_INITIALIZER: Context<C> = Context<C>::new();` should be sufficient.

jhand2 (commented):
Just fyi, if you add the text `Fixes #<issue number>` to your PR description, when the PR is merged, github will automatically close the issue.

jhand2:
I don't have any better suggestion, but adding @ia0 just in case he has any suggestions for idiomatic ways to do this in Rust.

dpe/src/dpe_instance.rs
ia0: Indeed, I'm not sure I'm following either. I tried the following is the tests in `src/context.rs` and it worked:

```rust
        const INITIALIZER_CONTEXT: Context<OpensslCrypto> = Context::new();
        let mut contexts = [INITIALIZER_CONTEXT; MAX_HANDLES];
```

zhalvorsen: It looks like the problem is the generic in a `const`. You can do it if the type is known, but not if it is defined yet (i.e. `C` instead of `OpensslCrypto`). I got it to work with this:

```Rust
    pub(crate) fn new_context_handles() -> [Context<C>; MAX_HANDLES] {
        [(); MAX_HANDLES].map(|_| Context::<C>::new())
    }
```

The `map` function says its not optimized very well, but that shouldn't matter because it only happens once on initialization.

ia0: There's also `core::array::from_fn(|_| Context::new())` which is probably a bit more straightforward. Here's a patch: https://github.com/sree-revoori1/caliptra-dpe/compare/main...ia0:caliptra-dpe:pr89. A helper function `new_context_handles` may still be used though.

jhand2 (approved):

PR #90, opened 2023-05-12
Implement handling of INTERNAL_INPUT_INFO flag

sree-revoori1:
If this flag is set in derive_child, we serialize the attributes of the DPE profile and add it to the hash in derive_cdi. Handling the INTERNAL_INPUT_DICE flag will come later as this is dependent on implementing the get_certificate_chain function first.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -10,7 +10,7 @@ use crate::{
     response::{DpeErrorCode, GetProfileResp, Response},
     support::Support,
     tci::{TciMeasurement, TciNodeData},
-    DPE_PROFILE, MAX_HANDLES,
+    DPE_PROFILE, INTERNAL_DPE_INFO_SIZE, MAX_HANDLES,
 };
 use core::{marker::PhantomData, mem::size_of};
 use crypto::{Crypto, Hasher};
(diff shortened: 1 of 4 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -239,6 +239,17 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
 
+    fn serialize_internal_dpe_info(&self, internal_dpe_info: &mut [u8; INTERNAL_DPE_INFO_SIZE]) {
+        // Internal DPE Info contains get profile response fields as well as the size of the hashing algorithm used
+        let profile_size = self
+            .get_profile()
+            .unwrap()
+            .serialize(internal_dpe_info)
+            .unwrap();
+        internal_dpe_info[profile_size..]
+            .copy_from_slice(&DPE_PROFILE.get_hash_size().to_le_bytes());
```
jhand2, line 250: minor: I think I would prefer to include the profile ID instead of the hash size in case we have a future profile id with the same hash size (I think that is likely).

Should be able to do `(DPE_PROFILE as u32).to_le_bytes()`

dpe/src/lib.rs
```diff
@@ -27,6 +28,8 @@ const CURRENT_PROFILE_MINOR_VERSION: u16 = 8;
 const VENDOR_ID: u32 = 0;
 const VENDOR_SKU: u32 = 0;
 
+const INTERNAL_DPE_INFO_SIZE: usize = size_of::<GetProfileResp>() + size_of::<usize>();
```
jhand2, line 31: Could just put this in the `derive_child.rs` file. It's only used there anyway

jhand2: Ok, no problem

jhand2 (approved):

PR #95, opened 2023-05-21
Change return type of derive_private_key

sree-revoori1:
We change the return type of derive_private_key from a PrivKey to a Result containing a PrivKey so that stubbing this function for the Caliptra implementation is simpler.

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -47,20 +47,17 @@ impl<C: Crypto> CommandExecution<C> for CertifyKeyCmd {
 
         let algs = DPE_PROFILE.alg_len();
         let priv_key = if self.uses_nd_derivation() {
-            dpe.contexts[idx].cached_priv_key.take().unwrap_or_else({
-                || {
-                    C::derive_private_key(
-                        algs,
-                        &dpe.derive_cdi(idx, true).unwrap(),
-                        &self.label,
-                        b"ECC",
-                    )
-                }
-            })
+            if let Some(cached) = dpe.contexts[idx].cached_priv_key.take() {
+                Ok(cached)
+            } else {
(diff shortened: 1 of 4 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -55,11 +55,12 @@ impl<C: Crypto> CommandExecution<C> for CertifyKeyCmd {
                         &self.label,
                         b"ECC",
                     )
+                    .unwrap()
```
jhand2, line 58: We should avoid calling unwrap wherever possible (here and in sign.rs). In general, it is prefferable to bubble up the result. Since we're changing error types, you can do `.map_err(|_| <some_dpe_error>)`. This will make `priv_key` a result, but you can always add`?` to the end of the assignment to say "If the result is an error, return it"

jhand2 (changes requested):

dpe/src/commands/certify_key.rs
```diff
@@ -55,11 +55,13 @@ impl<C: Crypto> CommandExecution<C> for CertifyKeyCmd {
...
```
jhand2, line 58: I think instead, you should be able to do something like this (in all the cases in this PR):

```rust
            if let Some(cached) = dpe.contexts[idx].cached_priv_key.take() {
                Ok(cached)
            } else {
                C::derive_private_key(
                    algs,
                    &dpe.derive_cdi(idx, true)?,
                    &self.label,
                    b"ECC",
                ).map_err(|_| DpeErrorCode::InternalError)
            }
        } else {
            C::derive_private_key(algs, &cdi, &self.label, b"ECC")
                .map_err(|_| DpeErrorCode::InternalError)
        }?;
```

jhand2 (approved):

PR #99, opened 2023-05-24
Handle INTERNAL_INPUT_DICE flag and add Platform trait

sree-revoori1:
If derive_child has the INTERNAL_INPUT_DICE flag set, we mix in the certificate chain into the hash when computing the CDI. 

We also add a platform trait with a get_certificate_chain function. We give a default implementation where it just returns a test ECDSA certificate generated using OpenSSL with the secp256r1 curve. The intention is for Caliptra to implement this and add parent certificates to the chain.


dpe/src/dpe_instance.rs
```diff
@@ -275,13 +281,41 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
+        if uses_internal_input_dice {
+            let ignored_locality = 0;
+            let mut offset = 0;
+            loop {
+                let cmd = GetCertificateChainCmd {
+                    size: MAX_CERT_SIZE.try_into().unwrap(),
+                    offset,
+                };
+                match cmd.execute(self, ignored_locality) {
+                    Ok(Response::GetCertificateChain(GetCertificateChainResp {
```
jhand2, line 308: I think instead of calling GetCertificateChain directly, we should maybe have a platform API that gets the certificate chain. Then this function and the GetCertificateChain command can both just call that.

If you just want to stub it for this PR that would be fine.

dpe/src/dpe_instance.rs
```diff
@@ -275,13 +284,41 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
```
jhand2, line 311: Can this call `P::get_certificate_chain` instead? That will avoid the command structure setup, calling cmd.execute, etc.

platform/src/default.rs
```diff
@@ -0,0 +1,11 @@
+// Licensed under the Apache-2.0 license
+
+use crate::{CertChunk, Platform, PlatformError};
+
+pub struct DefaultPlatform;
+
+impl Platform for DefaultPlatform {
+    fn get_certificate_chain(_offset: u32, _size: u32) -> Result<CertChunk, PlatformError> {
+        Err(PlatformError::NotImplemented)
```
jhand2, line 9: I would recommend generating a test cert (with OpenSSL or something) and putting it in a buffer so the things that depend on this can be tested. Something like [this](https://stackoverflow.com/questions/11992036/how-do-i-create-an-ecdsa-certificate-with-the-openssl-command-line).

Then you can hex dump the file (with e.g. `xxd` and paste it in the code as a `u8` array.

At some point, would be good to make that cert have the same public key as the alias key cert, but that can come later.

platform/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    NotImplemented,
+}
+
+pub struct CertChunk {
+    pub bytes: [u8; MAX_CERT_SIZE],
+    pub len: usize,
+}
+
+pub trait Platform {
+    fn get_certificate_chain(offset: u32, size: u32) -> Result<CertChunk, PlatformError>;
```
jhand2, line 19: I think returning CertChunk here requires an unnecessary copy. Instead, could do

```rust
/// Copies the out.len() bytes of the certificate chain starting at offset.
fn get_certficiate_chain(offset: u32, out: &mut [u8]) -> Result<(), PlatformError>
```

jhand2: The caller is just asking for as much of it as they want. `get_certificate_chain` would fill the whole `out` buffer.

One thing we could add is to return the number of bytes written. This way if the caller asks for more than are left, `get_certificate_chain` will just return the smaller number so the caller knows.

platform/src/default.rs
```diff
@@ -0,0 +1,153 @@
+// Licensed under the Apache-2.0 license
+
+use crate::{Platform, PlatformError, MAX_CHUNK_SIZE};
+use core::cmp::min;
+
+pub struct DefaultPlatform;
+
+pub const TEST_CERT_CHAIN: [u8; 2048] = [
```
jhand2, line 8: nit: Could you add a comment with how this is generated (in case we need to re-generate it)

platform/src/default.rs
```diff
@@ -0,0 +1,153 @@
...
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+];
+
+impl Platform for DefaultPlatform {
+    fn get_certificate_chain(
+        offset: u32,
+        size: u32,
+        out: &mut [u8; MAX_CHUNK_SIZE],
+    ) -> Result<u32, PlatformError> {
+        let len = TEST_CERT_CHAIN.len().try_into().unwrap();
```
jhand2, line 145: `TEST_CERT_CHAIN.len() as u32` is fine since we know the length of the memory fits in a u32

dpe/src/commands/get_certificate_chain.rs
```diff
@@ -15,14 +16,24 @@ pub struct GetCertificateChainCmd {
...
-impl<C: Crypto> CommandExecution<C> for GetCertificateChainCmd {
-    fn execute(&self, _dpe: &mut DpeInstance<C>, _locality: u32) -> Result<Response, DpeErrorCode> {
+impl<C: Crypto, P: Platform> CommandExecution<C, P> for GetCertificateChainCmd {
+    fn execute(
+        &self,
+        _dpe: &mut DpeInstance<C, P>,
+        _locality: u32,
+    ) -> Result<Response, DpeErrorCode> {
         // Make sure the operation is supported.
         if self.size > MAX_CERT_SIZE.try_into().unwrap() {
```
jhand2, line 26: Same as suggestion below. Can just do `MAX_CERT_SIZE as u32`

dpe/src/dpe_instance.rs
```diff
@@ -275,13 +284,33 @@ impl<C: Crypto> DpeInstance<'_, C> {
...
+                .map_err(|_| DpeErrorCode::InternalError)?;
+        }
+
+        // Add internal input dice to hash
+        if uses_internal_input_dice {
+            let mut offset = 0;
+            let mut cert_chunk = [0u8; MAX_CHUNK_SIZE];
+            while let Ok(len) = P::get_certificate_chain(
+                offset,
+                MAX_CHUNK_SIZE.try_into().unwrap(),
```
jhand2, line 307: as u32

cert.pem
```diff
@@ -0,0 +1,15 @@
+-----BEGIN CERTIFICATE-----
```
jhand2, line 1: Can you move this to a directory `platform/test_data/cert.pem`?

jhand2 (dismissed):

jhand2 (approved):

PR #109, opened 2023-06-12
Move CommandHdr::new outside test block

sree-revoori1:
We will need this in runtime integration tests so we move it outside of the test block so we can use in other crates. Also rename it to new_for_test so it is more clear it is only used in tests.

### DIFF
--- a/dpe/src/commands/mod.rs
+++ b/dpe/src/commands/mod.rs
@@ -103,6 +103,24 @@ impl Command {
     }
 }
 
+impl From<Command> for u32 {
+    fn from(cmd: Command) -> u32 {
+        match cmd {
+            Command::GetProfile => Command::GET_PROFILE,
+            Command::InitCtx(_) => Command::INITIALIZE_CONTEXT,
+            Command::DeriveChild(_) => Command::DERIVE_CHILD,
+            Command::CertifyKey(_) => Command::CERTIFY_KEY,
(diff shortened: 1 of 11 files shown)


dpe/src/commands/mod.rs
```diff
@@ -121,6 +120,27 @@ pub struct CommandHdr {
...
+            Command::RotateCtx(_) => Command::ROTATE_CONTEXT_HANDLE,
+            Command::DestroyCtx(_) => Command::DESTROY_CONTEXT,
+            Command::ExtendTci(_) => Command::EXTEND_TCI,
+            Command::TagTci(_) => Command::TAG_TCI,
+            Command::GetTaggedTci(_) => Command::GET_TAGGED_TCI,
+            Command::GetCertificateChain(_) => Command::GET_CERTIFICATE_CHAIN,
+        };
+        CommandHdr {
+            magic: Self::DPE_COMMAND_MAGIC,
+            cmd_id,
```
jhand2, line 140: nit: If you implement this as the From trait on Command, you can probably just use `command.into()`

```rust
impl From<u32> for Command {
...
}
```

For example https://github.com/chipsalliance/caliptra-sw/blob/7a3bb74f9c3c8131f4f20bbe65bb7bfffb89a490/cpu/src/trap.rs#L100

jhand2 (approved):

PR #110, opened 2023-06-13
Make DPE cmd execution return the response struct with a response hdr

sree-revoori1:
Instead of serializing the DPE response, this change makes DPE command execution return the response struct itself with a response header, so that the caller can decide what to do with it. We need to add the header to each DPE response struct so that if and when we serialize it using as_bytes, the header exists in the serialized buffer.

### DIFF
--- a/dpe/src/response.rs
+++ b/dpe/src/response.rs
@@ -19,11 +19,12 @@ pub enum Response {
     RotateCtx(NewHandleResp),
     CertifyKey(CertifyKeyResp),
     Sign(SignResp),
-    DestroyCtx,
+    DestroyCtx(ResponseHdr),
     ExtendTci(NewHandleResp),
     TagTci(NewHandleResp),
     GetTaggedTci(GetTaggedTciResp),
     GetCertificateChain(GetCertificateChainResp),
+    Error(ResponseHdr),
 }
 
 impl Response {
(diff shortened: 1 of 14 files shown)


dpe/src/response.rs
```diff
@@ -105,6 +114,7 @@ pub struct NewHandleResp {
 pub struct DeriveChildResp {
     pub handle: ContextHandle,
     pub parent_handle: ContextHandle,
+    pub resp_hdr: ResponseHdr,
```
jhand2, line 117: To make the serialization correct I think resp_hdr needs to be the first field in each of these structs

simulator/src/main.rs
```diff
@@ -32,16 +32,28 @@ fn handle_request(dpe: &mut DpeInstance<OpensslCrypto, DefaultPlatform>, stream:
...
+        Response::CertifyKey(ref res) => res.resp_hdr.status,
+        Response::Sign(ref res) => res.resp_hdr.status,
+        Response::DestroyCtx(ref resp_hdr) => resp_hdr.status,
+        Response::ExtendTci(ref res) => res.resp_hdr.status,
+        Response::TagTci(ref res) => res.resp_hdr.status,
+        Response::GetTaggedTci(ref res) => res.resp_hdr.status,
+        Response::GetCertificateChain(ref res) => res.resp_hdr.status,
+        Response::Error(ref resp_hdr) => resp_hdr.status,
+    };
+    // let response_code = u32::from_le_bytes(response[4..8].try_into().unwrap());
```
jhand2, line 51: nit: Delete this line

jhand2 (approved):

jhand2 (approved):

PR #113, opened 2023-06-14
Set X.509 TcbInfo vendor_info to target_locality of the TCI

sree-revoori1:
Also removes internal flag from TciNodeData since we don't use this in setting vendor_info anymore.

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -65,6 +65,7 @@ impl<C: Crypto> Context<C> {
         self.handle = *args.handle;
         self.tci = TciNodeData::new();
         self.tci.tci_type = args.tci_type;
+        self.tci.locality = args.locality;
         self.children = 0;
         self.parent_idx = args.parent_idx;
         self.context_type = args.context_type;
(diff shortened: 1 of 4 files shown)


dpe/src/tci.rs
```diff
@@ -8,32 +8,18 @@ use zerocopy::AsBytes;
 #[cfg_attr(test, derive(zerocopy::FromBytes))]
 pub(crate) struct TciNodeData {
     pub tci_type: u32,
-
-    // Bits
-    // 31: INTERNAL
-    // 30-0: Reserved. Must be zero
-    flags: u32,
```
jhand2, line 15: Can you also update the spect to remove the flags field from TCI_NODE_DATA? I agree it's not needed.

dpe/src/context.rs
```diff
@@ -65,6 +65,7 @@ impl<C: Crypto> Context<C> {
         self.handle = *args.handle;
         self.tci = TciNodeData::new();
         self.tci.tci_type = args.tci_type;
+        self.tci.locality = args.locality;
```
jhand2, line 68: I think we also need to update DeriveChild and InitializeContext to set the locality properly

jhand2: Oh, we're already doing the right things in both cases. I though locality was a new argument to active. Thanks!

jhand2 (approved):

PR #117, opened 2023-06-23
Move x509, tci and error codes to a common library

willyzha:
Moving the x509 cert generation code to a separate library so it can be depended on in a binary tool for generating sample DPE certificates.

### DIFF
--- a/dpe/Cargo.toml
+++ b/dpe/Cargo.toml
@@ -7,12 +7,12 @@ edition = "2021"
 
 [features]
 default = ["dpe_profile_p256_sha256"]
-dpe_profile_p256_sha256 = []
-dpe_profile_p384_sha384 = []
+dpe_profile_p256_sha256 = ["common/dpe_profile_p256_sha256"]
+dpe_profile_p384_sha384 = ["common/dpe_profile_p384_sha384"]
 
 [dependencies]
 crypto = {path = "../crypto"}
-x509 = {path = "../x509"}
+common = {path = "../common"}
 platform = {path = "../platform"}
 zerocopy = "0.6.1"
 
@@ -21,4 +21,4 @@ asn1 = "0.13.0"
 openssl = "0.10"
 x509-parser = "0.14.0"
 crypto = {path = "../crypto", features = ["deterministic_rand"]}
-x509 = {path = "../x509", features = ["dpe_profile_p256_sha256"]}
+common = {path = "../common", default-features = false}
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -7,10 +7,11 @@ Abstract:
 use crate::{
     commands::{Command, CommandExecution, InitCtxCmd},
     context::{ChildToRootIter, Context, ContextHandle, ContextState},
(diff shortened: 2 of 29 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -296,7 +296,10 @@ impl<C: Crypto, P: Platform> DpeInstance<'_, C, P> {
             let context = status?;
 
             let mut tci_bytes = [0u8; size_of::<TciNodeData>()];
-            let len = context.tci.serialize(&mut tci_bytes)?;
+            let len = match context.tci.serialize(&mut tci_bytes) {
+                Ok(usize) => Ok(usize),
+                Err(_) => Err(DpeErrorCode::InvalidArgument)
+            }?;
```
zhalvorsen, line 302: fyi: this can be simplified to
```suggestion
            let len = match context.tci.serialize(&mut tci_bytes).map_err(|_| DpeErrorCode::InvalidArgument)?;
```

willyzha: A lot of features to learn in rust :) thank you!

I changed this in 3 places.

dpe/Cargo.toml
```diff
@@ -20,3 +21,4 @@ asn1 = "0.13.0"
 openssl = "0.10"
 x509-parser = "0.14.0"
 crypto = {path = "../crypto", features = ["deterministic_rand"]}
+x509 = {path = "../x509", features = ["dpe_profile_p256_sha256"]}
```
jhand2, line 24: I would recommend reversing this dependency. Instead, above you can do something like

```
[features]
default = ["dpe_profile_p256_sha256"]
dpe_profile_p256_sha256 = ["x509/dpe_profile_p256_sha256"]
dpe_profile_p384_sha384 = ["x509/dpe_profile_p384_sha384"]
```

x509/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+    InvalidCommand = 2,
+    InvalidArgument = 3,
+}
+
+pub enum DpeProfile {
+    P256Sha256 = 1,
+    P384Sha384 = 2,
+}
+
+impl DpeProfile {
```
jhand2, line 25: I wonder if we should have a `common` crate for stuff like this and errors. Since this isn't really X.509-specific

x509/src/tci.rs
```diff
@@ -1,12 +1,12 @@
...
-use crate::{response::DpeErrorCode, DPE_PROFILE};
+use crate::{DPE_PROFILE, X509ErrorCode};
 use core::mem::size_of;
 use zerocopy::AsBytes;
 
 #[repr(C, align(4))]
 #[derive(Default, Copy, Clone, AsBytes)]
 #[cfg_attr(test, derive(zerocopy::FromBytes))]
-pub(crate) struct TciNodeData {
+pub struct TciNodeData {
```
jhand2, line 9: Would also be good to put this whole file in a common crate

jhand2 (dismissed):

jhand2 (approved):

willyzha:
Zach suggested we don't need pull this into a separate library. We can just make x509 and tci libraries public to have test binaries depend on x509 code.

PR #118, opened 2023-06-25
Implement ALLOW_X509 and ALLOW_CA CertifyKey/DeriveChild arguments

### LINKED ISSUE #106: Implement ALLOW_X509 and ALLOW_CA DeriveChild arguments
Implement DeriveChild inputs which allow X.509 format certs and IsCA=True

Spec: https://github.com/TrustedComputingGroup/Server-Internal/pull/123

sree-revoori1:
Fixes #106

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -33,7 +33,7 @@ impl CertifyKeyCmd {
         self.flags & Self::ND_DERIVATION != 0
     }
 
-    const fn _uses_is_ca(&self) -> bool {
+    const fn uses_is_ca(&self) -> bool {
         self.flags & Self::IS_CA != 0
     }
 }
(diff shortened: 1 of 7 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -127,7 +144,10 @@ impl<C: Crypto, P: Platform> CommandExecution<C, P> for CertifyKeyCmd {
                 u32::try_from(bytes_written).map_err(|_| DpeErrorCode::InternalError)?
             }
             Self::FORMAT_CSR => {
-                return Err(DpeErrorCode::ArgumentNotSupported);
+                if !dpe.support.csr {
+                    return Err(DpeErrorCode::ArgumentNotSupported);
+                }
+                return Err(DpeErrorCode::InternalError);
```
jhand2, line 150: Should probably still return NotSupported here since this isn't implemented.

dpe/src/commands/derive_child.rs
```diff
@@ -95,10 +92,20 @@ impl<C: Crypto, P: Platform> CommandExecution<C, P> for DeriveChildCmd {
         if !dpe.support.internal_info && self.uses_internal_info_input()
             || !dpe.support.internal_dice && self.uses_internal_dice_input()
         {
-            return Err(DpeErrorCode::InvalidArgument);
+            return Err(DpeErrorCode::ArgumentNotSupported);
+        }
+
+        if !dpe.support.is_ca && self.allows_ca() || !dpe.support.x509 && self.allows_x509() {
```
jhand2, line 98: I think the two clauses here need parentheses

```suggestion
        if (!dpe.support.is_ca && self.allows_ca()) || (!dpe.support.x509 && self.allows_x509()) {
```

dpe/src/commands/derive_child.rs
```diff
@@ -95,10 +92,20 @@ impl<C: Crypto, P: Platform> CommandExecution<C, P> for DeriveChildCmd {
...
+            return Err(DpeErrorCode::ArgumentNotSupported);
         }
 
         let parent_idx = dpe.get_active_context_pos(&self.handle, locality)?;
+        if !dpe.contexts[parent_idx].allow_ca && self.allows_ca()
```
jhand2, line 103: Same as above

verification/verification_test.go
```diff
@@ -238,9 +238,9 @@ func testInitContext(d TestDPEInstance, t *testing.T) {
 func TestCertifyKey(t *testing.T) {
```
jhand2, line 238: Can you add a test case here to make sure that if CertifyKey is called with is_ca set to true/false, the right values get set in the cert?

PR #119, opened 2023-06-29
Add tool for generating a sample DPE certificate

willyzha:
Binary tool for generating sample DPE certificates for testing.


tools/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "tools"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+default = ["dpe_profile_p256_sha256"]
+dpe_profile_p256_sha256 = ["dpe/dpe_profile_p256_sha256"]
+dpe_profile_p384_sha384 = ["dpe/dpe_profile_p384_sha384"]
```
zhalvorsen, line 9: It doesn't look like we are using any of these features. I think they can  be removed. We will probably want them later when we need to test the other profile though. Do you think its worth removing them?

crypto/src/openssl.rs
```diff
@@ -210,6 +211,23 @@ impl Crypto for OpensslCrypto {
         Ok(super::EcdsaSig { r, s })
     }
 
+    #[cfg(feature = "alias_key_from_file")]
```
zhalvorsen, line 214: I kind of like the idea of always having the alias key come from a file. Should we make this the only option?

crypto/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    let pem: Vec<u8>;
+    if Path::new(ALIAS_PRIV).exists() {
+        let input_pem = fs::read(ALIAS_PRIV).unwrap();
+        let ec_priv: ec::EcKey<pkey::Private> = ec::EcKey::private_key_from_pem(&input_pem).unwrap();
+        pem = ec_priv.private_key_to_pem().unwrap();
+    } else {
+        let group = ec::EcGroup::from_curve_name(nid::Nid::X9_62_PRIME256V1).unwrap();
+        let ec_key = ec::EcKey::generate(&group).unwrap();
+        pem = ec_key.private_key_to_pem().unwrap();
+    }
```
zhalvorsen, line 27: nit: This could be changed to something like this
```suggestion
    let pem = if Path::new(ALIAS_PRIV).exists() {
        let input_pem = fs::read(ALIAS_PRIV).unwrap();
        let ec_priv: ec::EcKey<pkey::Private> = ec::EcKey::private_key_from_pem(&input_pem).unwrap();
        ec_priv.private_key_to_pem().unwrap()
    } else {
        let group = ec::EcGroup::from_curve_name(nid::Nid::X9_62_PRIME256V1).unwrap();
        let ec_key = ec::EcKey::generate(&group).unwrap();
        ec_key.private_key_to_pem().unwrap()
    }
```

tools/Cargo.toml
zhalvorsen: Nevermind, I didn't see that it was passing the profile to the DPE crate.

crypto/src/openssl.rs
willyzha: Removed.

crypto/build.rs
willyzha: Done.

tools/src/sample_dpe_cert.rs
```diff
@@ -0,0 +1,56 @@
...
+        Response::DeriveChild(_res) =>  Err("Unexpected Response"),
+        Response::RotateCtx(_res) =>  Err("Unexpected Response"),
+        Response::Sign(_res) =>  Err("Unexpected Response"),
+        Response::DestroyCtx(_res) =>  Err("Unexpected Response"),
+        Response::ExtendTci(_res) =>  Err("Unexpected Response"),
+        Response::TagTci(_res) =>  Err("Unexpected Response"),
+        Response::GetTaggedTci(_res) =>  Err("Unexpected Response"),
+        Response::GetCertificateChain(_res) =>  Err("Unexpected Response"),
+        Response::Error(_res) =>  Err("Unexpected Response"),
+    }.unwrap();
```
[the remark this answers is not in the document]
willyzha: Done.

tools/src/sample_dpe_cert.rs
```diff
@@ -0,0 +1,92 @@
...
+    let certify_key_response = match resp {
+        // Expect CertifyKey response return an error in all other cases.
+        Response::CertifyKey(res) => res,
+        Response::Error(res) => panic!("Error response {}", res.status),
+        _ => panic!("Unexpected Response"),
+    };
+
+    let pem = Pem::new(
+        "CERTIFICATE",
+        &certify_key_response.cert[..usize::try_from(certify_key_response.cert_size).unwrap()],
```
zhalvorsen, line 80: fyi: you can cast between a lot of primitive types with the `as` keyword:

```suggestion
        &certify_key_response.cert[..certify_key_response.cert_size as usize],
```

crypto/src/openssl.rs
```diff
@@ -184,24 +184,14 @@ impl Crypto for OpensslCrypto {
 
         Ok(EcdsaPub { x, y })
     }
-
+
```
zhalvorsen, line 187: It looks like you might need to run `cargo fmt`. It should get rid of extra whitespace. You can run the CI tests locally with the `ci.sh` script at the top level. The script will remind you run `cargo fmt` if you need to.

zhalvorsen (dismissed):

crypto/src/openssl.rs
willyzha: Done.

tools/src/sample_dpe_cert.rs
willyzha: done.

tools/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+dpe = {path = "../dpe"}
+crypto = {path = "../crypto", features = ["deterministic_rand"]}
+pem = "2"
+platform = {path = "../platform"}
+zerocopy = "0.6.1"
+
+[[bin]]
+name = "sample_dpe_cert"
+path = "src/sample_dpe_cert.rs"
+target = 'cfg("test")'
```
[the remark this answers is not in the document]
willyzha: I believe so. I followed this https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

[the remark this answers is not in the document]
zhalvorsen: Sounds good to me

zhalvorsen (approved):

PR #124, opened 2023-07-14
Remove Crypto and Platform templating from DpeInstance

jhand2:
Remove generics for Crypto/Platform from the codebase. Instead pass a DpeEnv around where needed.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -12,12 +12,20 @@ use crate::{
     tci::{TciMeasurement, TciNodeData},
     DPE_PROFILE, INTERNAL_INPUT_INFO_SIZE, MAX_HANDLES,
 };
-use core::{marker::PhantomData, mem::size_of};
-use crypto::{Crypto, Hasher};
+use core::mem::size_of;
+use crypto::{Crypto, Digest, Hasher};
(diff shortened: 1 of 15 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -173,11 +181,11 @@ impl<C: Crypto, P: Platform> DpeInstance<'_, C, P> {
 
     pub(crate) fn generate_new_handle(
         &self,
-        crypto: &mut C,
+        env: &mut impl DpeEnv,
```
zhalvorsen, line 184: As a general comment in this file, should we just take `&mut impl Crypto` in some of these because it doesn't need platform? Its kind a tradeoff though, It will probably make the callers a bit uglier. What do you think?

zhalvorsen (approved):

zhalvorsen:
This looks great! Thanks for doing it. It is so much easier to read and understand now.

dpe/src/dpe_instance.rs
jhand2: I liked always passing DpeEnv for consistency. Espcially because it makes it easier if we later add stuff to platform since everyone has access to it already as part of the Env.

zhalvorsen: I'm good with that

PR #127, opened 2023-07-21
Fix x509 encoding bugs

willyzha:
- modify RDN encoding of cn and serialnumber attributes into seperate sets
 - change RDN cn encoding type to utf8string

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -38,6 +38,7 @@ impl X509CertWriter<'_> {
     const BIT_STRING_TAG: u8 = 0x3;
     const OCTET_STRING_TAG: u8 = 0x4;
     const OID_TAG: u8 = 0x6;
+    const UTF8_STRING_TAG: u8 = 0xc;
     const PRINTABLE_STRING_TAG: u8 = 0x13;
     const GENERALIZE_TIME_TAG: u8 = 0x18;
     const SEQUENCE_TAG: u8 = 0x30;
@@ -47,7 +48,7 @@ impl X509CertWriter<'_> {
     const BOOL_SIZE: usize = 1;
 
     // Constants for setting tag bits
-    const PRIVATE: u8 = 0x80; // Used for Implicit/Explicit tags
+    const CONTEXT_SPECIFIC: u8 = 0x80; // Used for Implicit/Explicit tags
     const CONSTRUCTED: u8 = 0x20; // SET{OF} and SEQUENCE{OF} have this bit set
 
     const X509_V3: u64 = 2;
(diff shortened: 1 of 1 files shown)


willyzha:
I think we need these changes to correctly encode the cert according to the spec. In particular the tcbinfo part I think we need the extra sequence to properly encode the fwidlist structure which was missing previously.

dpe/src/x509.rs
```diff
@@ -1028,6 +1060,11 @@ mod tests {
         pub(crate) digest: &'a [u8],
     }
 
+    #[derive(asn1::Asn1Read)]
+    pub struct FwidList<'a> {
```
jhand2, line 1064: I'm not sure this is right. I think it introduces an extra sequence around the fwidlist (because the asn1 crate treats structs as sequences)

```
DiceTcbInfo ::== SEQUENCE {
  ...
  fwids [6] IMPLICIT FWIDLIST OPTIONAL,
  ...
}


FWIDLIST ::== SEQUENCE SIZE (1..MAX) OF FWID
FWID ::== SEQUENCE {
  hashAlg OBJECT IDENTIFIER,
  digest OCTET STRING
}
```

I think this should create the hierarchy:

```
SEQUENCE // DiceTcbInfo
  SEQEUNCE OF // fwids field
    SEQUENCE // fwids[0]
    SEQUENCE // fwids[1]
    ...etc...
```

willyzha: Is it possible that asn1 crate doesn't treat `asn1::SequenceOf<'a, Fwid<'a>>` as an extra sequence if the contents is a struct (non primitive)?

When printout the raw asn1 from this test it seems correct. And asn1.parse_single is only able to parse the raw asn1 bytes with this struct definition.

Here's the ASN1 structure output from encode_tcb_info
```
SEQUENCE (3 elem)
  [6] (1 elem)
    SEQUENCE (2 elem)
      SEQUENCE (2 elem)
        OBJECT IDENTIFIER 2.16.840.1.101.3.4.2.1 sha-256 (NIST Algorithm)
        OCTET STRING (32 byte) BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB
      SEQUENCE (2 elem)
        OBJECT IDENTIFIER 2.16.840.1.101.3.4.2.1 sha-256 (NIST Algorithm)
        OCTET STRING (32 byte) AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
  [8] (4 byte) FFFFFFFF
  [9] (4 byte) 11223344
```

willyzha: Reverted the TCBINFO changes

jhand2 (approved):

PR #129, opened 2023-08-02
[fuzzing] Implement initial support for DPE services fuzzing

benjamindoron:
First implementation of fuzz testing the DPE library, forming another part of the fuzzing work. Instructions and notes are in the readme.md files, these can be removed or migrated to actual documentation if required.


dpe/fuzz/fuzz_targets/runtime_dpe_dpe_fuzz_harness.rs
```diff
@@ -0,0 +1 @@
+../../runtime_dpe_dpe_fuzz_harness.rs
```
jhand2, line 1: Is this symlinked? It might make more sense to have a helper crate that both sets of fuzz tests can depend on.

dpe/runtime_dpe_dpe_fuzz_harness.rs
```diff
@@ -0,0 +1,98 @@
...
+use log::{trace, LevelFilter};
+use simplelog::{Config, WriteLogger};
+use std::fs::OpenOptions;
+
+use crypto::OpensslCrypto;
+use dpe::dpe_instance::DpeEnv;
+use dpe::{commands::Command, response::Response, DpeInstance, Support};
+use platform::{DefaultPlatform, AUTO_INIT_LOCALITY};
+
+// TODO: Review impact.
```
jhand2, line 12: It might be good to also fuzz the support matrix. Basically if we disable some combination of the matrix, does any weird stuff happen. But that's lower priority. Not sure if it would make the search space too big.

dpe/runtime_dpe_dpe_fuzz_harness.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn crypto(&mut self) -> &mut OpensslCrypto {
+        &mut self.crypto
+    }
+
+    fn platform(&mut self) -> &mut DefaultPlatform {
+        &mut self.platform
+    }
+}
+
+// One thing to note: AFL is using persistent mode. If objects aren't dropped, this might double-initialise
```
jhand2, line 45: Can you expand on what this means?

dpe/runtime_dpe_dpe_fuzz_harness.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn harness(data: &[u8]) {
+    // NOTE: This is racey
```
jhand2, line 47: In what way?

dpe/afl/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+[package]
+name = "dpe-afl"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[dependencies]
+afl = "*"
+log = "*"
+simplelog = "*"
```
jhand2, line 12: Prefer specific versions so everyone is running tests with the same environment.

dpe/fuzz/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "*"
+log = "*"
+simplelog = "*"
```
jhand2, line 15: Same, let's version these

dpe/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,12 @@
...
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+mod runtime_dpe_dpe_fuzz_harness;
+use runtime_dpe_dpe_fuzz_harness::harness;
+
+fuzz_target!(|data: &[u8]| {
+    harness(data);
+});
```
jhand2, line 12: Do aft and fuzz need to be different crates? I would expect you could do something like:

```rust

#![no_main]

use afl;
use libfuzzer_sys;

mod runtime_dpe_dpe_fuzz_harness;
use runtime_dpe_dpe_fuzz_harness::harness;

libfuzzer_sys::fuzz_target!(|data: &[u8]| {
    harness(data);
});

afl::fuzz_target!(|data: &[u8]| {
    harness(data);
});
```

jhand2: Alternatively, if you think it's better to be able to run them separately, we can have a feature that can be enabled/disabled on the command-line.

dpe/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: We could do it using the fuzzer. I've been thinking of enabling structure-aware fuzzing, this use-case is another candidate for that. Rather than taking a byte slice as an argument, we can take any struct. So, this means that we can declare new structs here, with fields for `Support`, command, etc, and our input should match all those types.

Depending on the complexity of the struct, it would take some time. I would need to make all entries in the struct derive `Arbitrary`, from the crate that performs this mutation.

dpe/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: Executing multiple workers that all write to the same file will result in an unreadable file. I suppose I meant that there's nothing to prevent them from racing to write individual bytes/lines.

dpe/afl/Cargo.toml
benjamindoron: I specified versions of libraries used internally, such as `openssl`, I wasn't sure there was a point here. Maybe?

dpe/fuzz/fuzz_targets/fuzz_target_1.rs
benjamindoron: They have different function prototypes. fuzz is `no_main`, but afl actually expands the `fuzz!` macro inside `main()`. It might work though, I remember that the macro in libfuzzer-sys doesn't use `main` either. They're using what I think is called an intrinsic function?

I'll try it out.

benjamindoron: > Alternatively, if you think it's better to be able to run them separately, we can have a feature that can be enabled/disabled on the command-line.

Also, yes. As I remember commenting on one of the other PRs, when I was writing this, I hadn't extensively compared the two, and it was easy enough to support both, so I did. Which should be run, or whether we should run both - this is supported, by sharing a corpus directory, as I recall - is something to be discussed too.

dpe/fuzz/fuzz_targets/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: Yes. It may be possible to merge all these files into one crate. Otherwise, I suppose so, I could implement that.

dpe/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: According to cargo-afl - I haven't confirmed - rather than `fork()`-ing, which has a performance penalty, the target function is re-executed instead. Since the `fuzz!` macro expands inside a function, I had been concerned if it was effectively a `while (true)` loop, then the DPE object might still be instantiated, with its old state from previous calls still resident. This would mean that we would also be fuzzing the effect of successive DPE commands, which might be interesting, but unplanned. After thinking about this some more, I don't think that's an issue: since this helper function is used, we are actually returning once, so I think this means that the object will be dropped.

Generally, I've looked further into the internals of the `libfuzzer-sys` crate than `afl`.

dpe/fuzz/fuzz_targets/fuzz_target_1.rs
jhand2: In that case we can probably have a feature. Could use `cfg_if`, since one needs no_main. I think that is preferable to defining two separate crates. Then when we run the test we can pass `--features=afl`

benjamindoron: I thought it would be possible to compile and link both together, since they have different entrypoints, but AFL packages its own LLVM runtime, causing linker errors. So, these are now both optional dependencies, one must be selected. This is resolved when I push updates next.

dpe/fuzz/fuzz_targets/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: Marking as resolved, AFL and libFuzzer are being merged, per below.

dpe/runtime_dpe_dpe_fuzz_harness.rs
jhand2: Got it. Is it possible to detect from inside the code which worker is executing and make them write to different files?

Can probably work this out in a different PR, but if it isn't too difficult it might be good to just take care of it now.

dpe/afl/Cargo.toml
jhand2: In general we prefer to specify specific versions so that we ensure the every developer + the CI are running the same code. It helps with reproducing any issues. So I think we should specify the latest version for all of these.

dpe/fuzz/fuzz_targets/fuzz_target_1.rs
jhand2: Looks good, thanks!

dpe/runtime_dpe_dpe_fuzz_harness.rs
jhand2: Ok. Could you open an issue for this to potentially look into? But that doesn't need to block this PR.

dpe/runtime_dpe_dpe_fuzz_harness.rs
jhand2: Ok, sounds good. Thanks for the explanation!

dpe/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: One solution is appending the PID to the filename, the workers' processes persist between calls to the `fuzz_target!`. But does it matter? I used logging for development, but now I think the fuzzer outputs all relevant information. Unless there's some invalid sequence of commands, and we're looking for logic bugs too, I think dropping this is easiest.

dpe/runtime_dpe_dpe_fuzz_harness.rs
benjamindoron: Done, https://github.com/chipsalliance/caliptra-sw/issues/624. I'm changing the comment here to reflect that.

dpe/fuzz/Cargo.toml
benjamindoron: Marking as done. Crates are now merged into one, so this is now a duplicate.

dpe/afl/Cargo.toml
benjamindoron: Done, will push shortly.

jhand2 (approved):

dpe/runtime_dpe_dpe_fuzz_harness.rs
jhand2: I'm going to merge this to avoid needing to wait another day. We can drop it in a follow-up CL if you think that's best.

PR #132, opened 2023-08-03
Add extended key usage extension to certificate

willyzha:


dpe/src/x509.rs
```diff
@@ -895,6 +930,51 @@ impl X509CertWriter<'_> {
...
+
+        // Extension data is sequence -> octet string. To compute size, wrap
+        // in tagging twice.
+        bytes_written += self.encode_byte(Self::OCTET_STRING_TAG)?;
+        bytes_written += self.encode_size_field(Self::get_structure_size(
+            Self::get_structure_size(policy_oid.len(), /*tagged=*/ true)?,
+            /*tagged=*/ true,
+        )?)?;
+
+        // Sequence size to just a tagged bool
```
jhand2, line 966: Not sure I understand this comment

dpe/src/x509.rs
```diff
@@ -895,6 +930,51 @@ impl X509CertWriter<'_> {
         Ok(bytes_written)
     }
 
+    /// Encode ExtendedKeyUsage extension
```
jhand2, line 933: Could you update the comment to explain which EKU OID will be included (based on whether cert is_ca)

dpe/src/x509.rs
```diff
@@ -1323,5 +1404,14 @@ mod tests {
             Ok(None) => panic!("key usage extension not found"),
             Err(_) => panic!("multiple key usage extensions found"),
         }
+
+        match cert.extended_key_usage() {
+            Ok(Some(ext_key_usage)) => {
+                assert!(ext_key_usage.critical);
+                assert_eq!(ext_key_usage.value.other, [oid!(2.23.133 .8 .7)]);
```
jhand2, line 1411: nit: Can you add a comment of which OID this is?

jhand2 (approved):

dpe/src/x509.rs
willyzha: Done

dpe/src/x509.rs
willyzha: Done.

dpe/src/x509.rs
willyzha: I think I copied that from somewhere :) updated the comment

jhand2 (approved):

PR #135, opened 2023-08-04
Change dpe environment to accept DER encoded issuer

willyzha:
It's difficult to account for different issuer name encoding in DPE. So rather than DER encoding the issuer from string inputs, instead accept a raw DER encoded byte array to place directly in the DPE certificate.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -29,19 +29,16 @@ pub struct DpeEnv<'a, T: DpeTypes + 'a> {
     pub platform: T::Platform,
 }
 
-pub struct DpeInstance<'a> {
+pub struct DpeInstance {
     pub(crate) contexts: [Context; MAX_HANDLES],
     pub(crate) support: Support,
 
     /// Can only successfully execute the initialize context command for non-simulation (i.e.
     /// `InitializeContext(simulation=false)`) once per reset cycle.
     pub(crate) has_initialized: bool,
-
-    // Issuer Common Name to use in DPE leaf certs
-    pub(crate) issuer_cn: &'a [u8],
 }
 
-impl DpeInstance<'_> {
+impl DpeInstance {
     const MAX_NEW_HANDLE_ATTEMPTS: usize = 8;
 
     /// Create a new DPE instance.
@@ -50,17 +47,15 @@ impl DpeInstance<'_> {
     ///
     /// * `support` - optional functionality the instance supports
     /// * `issuer_cn` - issuer Common Name to use in DPE leaf certs
-    pub fn new<'a>(
+    pub fn new(
         env: &mut DpeEnv<impl DpeTypes>,
         support: Support,
-        issuer_cn: &'a [u8],
-    ) -> Result<DpeInstance<'a>, DpeErrorCode> {
+    ) -> Result<DpeInstance, DpeErrorCode> {
         const CONTEXT_INITIALIZER: Context = Context::new();
         let mut dpe = DpeInstance {
             contexts: [CONTEXT_INITIALIZER; MAX_HANDLES],
             support,
             has_initialized: false,
-            issuer_cn,
         };
 
         if dpe.support.auto_init {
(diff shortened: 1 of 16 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -73,7 +73,16 @@ impl DpeInstance<'_> {
         env: &mut impl DpeEnv,
         support: Support,
     ) -> Result<DpeInstance<'a>, DpeErrorCode> {
-        const TEST_ISSUER: &[u8] = b"Test Issuer";
+        // DER encoded issuer with CN
```
jhand2, line 76: Can you add a comment for how you generated this? Also the mix of decimal numbers and ascii bytes is a bit odd.

dpe/src/x509.rs
```diff
@@ -1281,6 +1281,9 @@ mod tests {
             cn: b"Caliptra Alias",
             serial: [0x00; DPE_PROFILE.get_hash_size() * 2],
         };
+        let mut issuer_der = [0u8; 256];
+        let mut issuer_writer = X509CertWriter::new(&mut issuer_der, true);
+        let issuer_len = issuer_writer.encode_rdn(&test_issuer_name).unwrap();
```
jhand2, line 1286: nit: Could have a test helper for this. Something like

```rust
const TEST_ISSUER = Name {
    cn: b"Caliptra Alias",
    serial: [0x00; DPE_PROFILE.get_hash_size() * 2],
};

fn encode_test_issuer() -> Vec<u8> {
        let mut issuer_der = vec![0u8; 256];
        let mut issuer_writer = X509CertWriter::new(&mut issuer_der, true);
        let issuer_len = issuer_writer.encode_rdn(&TEST_ISSUER).unwrap();
        issuer_der.resize(issuer_len);
        issuer_der
}
```

dpe/src/dpe_instance.rs
```diff
@@ -33,8 +33,8 @@ pub struct DpeInstance<'a> {
     /// `InitializeContext(simulation=false)`) once per reset cycle.
     pub(crate) has_initialized: bool,
 
-    // Issuer Common Name to use in DPE leaf certs
-    pub(crate) issuer_cn: &'a [u8],
+    // DER encoded Issuer Common Name to use in DPE leaf certs
+    pub(crate) issuer_der: &'a [u8],
```
jhand2, line 37: I wonder if we should make this a platform API instead of passing to the DpeInstance constructor. The reason is that the platform may be dynamically encoding it (that's what Caliptra will probably do, because the issuer name can change). So if we pass it in, the lifetime stuff could get tricky.

For a platform API, I'm thinking something like

```rust

struct DerName {
  der: [u8; Self::MAX_ISSUER_SIZE];
  len: usize;
}

impl DerName {
  const MAX_ISSUER_SIZE = 256; // Maybe?
}

fn get_encoded_issuer() -> Result<DerName, PlatformError>;
```

dpe/src/dpe_instance.rs
```diff
@@ -73,12 +68,11 @@ impl DpeInstance<'_> {
         Ok(dpe)
     }
 
-    pub fn new_for_test<'a>(
+    pub fn new_for_test(
```
jhand2, line 71: nit: could just remove this. But that would require a large-scale find/replace, so I understand if you don't want to in this PR 😄

platform/src/default.rs
```diff
@@ -159,6 +168,15 @@ impl Platform for DefaultPlatform {
         Ok(cert_chunk_range_end - offset)
     }
 
+    fn get_issuer_name(&mut self, out: &mut [u8; MAX_CHUNK_SIZE]) -> Result<u32, PlatformError> {
+        let len = TEST_ISSUER.len() as u32;
```
jhand2, line 172: nit: Could return usize from this function to avoid casting. It's probably more useful to the caller as a usize anyway.

jhand2 (dismissed):

platform/src/default.rs
willyzha: Done.

dpe/src/dpe_instance.rs
willyzha: Done.

jhand2 (approved):

PR #141, opened 2023-08-11
Split the test file based on DPE commands

steffy3494:
This pull request contains only the test cases which are split based on DPE commands. The option to switch between emulator & simulator will be raised as a separate PR.


steffy3494:
This pull request contains only the test cases which are split based on DPE commands. The option to switch between emulator & simulator will be raised as a separate PR.

jhand2:
It looks like CI is failing because `go fmt` needs to be run. Can you address that?

In general, you can repro the CI by running `ci.sh`

steffy3494:
I execute the 'go fmt' command in verification folder, and it adjusts the space in verification_test.go. Pushed the same changes in the repo.

jhand2 (approved):

PR #142, opened 2023-08-14
Rename dpe crates to have dpe prefix

jhand2:
Rename crypto, platform, and simulator crates to use dpe-* naming to avoid confusing name collisions.


sree-revoori1 (dismissed):

clundin25 (dismissed):
@zhalvorsen if we are going to make this change, should we also include "caliptra"?

E.g. "caliptra-dpe-crypto". I think that would be necessary for publishing to crates.io

PR #150, opened 2023-08-18
Switch between simulator and emulator

steffy3494:
-Added the flag option to switch between emulator and simulator. 
-Added the dummy support to emulator. Once the emulator is implemented, will add the actual enabled feature.


verification/getProfile_test.go
```diff
@@ -10,40 +10,48 @@ import (
...
-		// Supports a couple combos.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{Simulation: true, AutoInit: true, RotateContext: true, Csr: true, InternalDice: true, IsCA: true}},
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{ExtendTci: true, Tagging: true, X509: true, InternalInfo: true}},
-		// Supports everything.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{Simulation: true, ExtendTci: true, AutoInit: true, Tagging: true, RotateContext: true, X509: true, Csr: true, IsSymmetric: true, InternalInfo: true, InternalDice: true, IsCA: true}},
+	var instances []TestDPEInstance
+	if *isEmulator {
+		//Added dummy support for emulator. Once the emulator is implemented, will add the actual enabled feature
+		instances = []TestDPEInstance{
+			&DpeInstance{exe_path: *socket_exe, supports: Support{AutoInit: true}},
```
jhand2, line 17: The Caliptra emulator won't support turning specific features off and on, it will just support whatever it supports.

verification/getProfile_test.go
```diff
@@ -10,40 +10,48 @@ import (
...
+		}
+	} else {
+		instances = []TestDPEInstance{
```
jhand2, line 20: Since testing the different support vectors only works with the simulator, I would recommend splitting this into its own test (something like `TestProfileMatrix`) and skipping it if the test target isn't the simulator.

verification/simulator.go
```diff
@@ -27,7 +27,7 @@ const (
 	DPE_SIMULATOR_VENDOR_SKU            uint32  = 0
 )
 
-type DpeSimulator struct {
```
jhand2, line 30: I think we should leave this as DpeSimulator. The DpeEmulator should just also implement the TestDPEInstance interface (but the implementation will be different)

verification/certifyKey_test.go
```diff
@@ -17,16 +17,27 @@ import (
...
-	simulators := []TestDPEInstance{
-		// No extra options besides AutoInit.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{AutoInit: true, X509: true}},
-		// Supports AutoInit and simulation contexts.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{AutoInit: true, Simulation: true, X509: true}},
+	var instances []TestDPEInstance
+	if *isEmulator {
+		//Added dummy support for emulator. Once the emulator is implemented, will add the actual enabled feature
+		instances = []TestDPEInstance{
+			&DpeInstance{exe_path: *socket_exe, supports: Support{AutoInit: true}},
```
jhand2, line 24: Should instantiate a DpeEmulator (and below instantiate a DpeSimulator)

verification/initializeContext_test.go
```diff
@@ -11,17 +11,25 @@ import (
...
+		//Added dummy support for emulator. Once the emulator is implemented, will add the actual enabled feature
+		instances = []TestDPEInstance{
+			&DpeInstance{exe_path: *socket_exe, supports: Support{AutoInit: true}},
+		}
+	} else {
+		instances = []TestDPEInstance{
+			// No extra options.
+			&DpeInstance{exe_path: *socket_exe},
+			// Supports simulation.
+			&DpeInstance{exe_path: *socket_exe, supports: Support{Simulation: true}},
```
jhand2, line 25: I think we may want to change how we do this. Instead, I'm thinking:

* Each test may require specific `support` features to be enabled
* If the test target matches the requires support, the test will run
* If the target doesn't match, we skip the test

This will mean we have more test cases, but it will be easier to run the tests against targets that have different support flags set.

We can have a GetTestTarget function that does something like this:

```go
fun GetTestTarget(support_needed Support) (error, TestDPEInstance) {
  if *isEmulator {
    // Check if all the support==True are supported in the emulator. If not, return error. If yes, return a DpeEmulator instance
  } else {
    // Make a DpeSimulator with the requested support
  }
}
```

verification/verification_test.go
```diff
@@ -4,9 +4,26 @@ package verification
...
+	"testing"
 )
 
-var sim_exe = flag.String("sim", "../simulator/target/debug/simulator", "path to simulator executable")
+var socket_exe *string
+var isEmulator *bool
+
+// This will be called before running tests, and it assigns the socket path based on command line flag.
+func TestMain(m *testing.M) {
+	isEmulator = flag.Bool("emulator", false, "socket type - emulator")
```
jhand2, line 16: Instead of a boolean flag, it would be nice to have a test target enum, where `emulator` and `simulator` are the current two options (and the default is simulator)

jhand2 (changes requested):

verification/initializeContext_test.go
steffy3494: Here, assumed like a fixed set of support features(support_needed) has been enabled in DpeEmulator struct and checked the Boolean value for the support needed when returning DPE instance for the emulator. Please check the GetEmulatorTarget code and provide any comments you may have.

verification/certifyKey_test.go
```diff
@@ -14,19 +15,43 @@ import (
...
+
+	if testTargetType == EMULATOR {
+		return GetEmulatorTarget(support_needed, instances)
+
+	} else if testTargetType == SIMULATOR {
+		instances = []TestDPEInstance{
+			// No extra options besides AutoInit.
+			&DpeSimulator{exe_path: *socket_exe, supports: Support{AutoInit: true, X509: true}},
+			// Supports AutoInit and simulation contexts.
+			&DpeSimulator{exe_path: *socket_exe, supports: Support{AutoInit: true, Simulation: true, X509: true}},
```
jhand2, line 35: Instead of making multiple instances  with different support matrices, I think we should have a separate test for each different support configuration.

verification/certifyKey_test.go
```diff
@@ -14,19 +15,43 @@ import (
...
+// This file is used to test the certify key command by using a simulator/emulator
+
+func GetTestTarget_CertifyKey() ([]TestDPEInstance, error) {
+
+	// Added dummy support for emulator
+	support_needed := []string{"AutoInit", "X509"}
+
+	var instances []TestDPEInstance
+
+	if testTargetType == EMULATOR {
```
jhand2, line 27: If testTargetType == EMULATOR and the requested support configuration isn't supported by the emualtor, return an error.

verification/certifyKey_test.go
```diff
@@ -14,19 +15,43 @@ import (
...
 func TestCertifyKey(t *testing.T) {
-	simulators := []TestDPEInstance{
-		// No extra options besides AutoInit.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{AutoInit: true, X509: true}},
-		// Supports AutoInit and simulation contexts.
-		&DpeSimulator{exe_path: *sim_exe, supports: Support{AutoInit: true, Simulation: true, X509: true}},
+
+	instances, err := GetTestTarget_CertifyKey()
+	if err != nil {
+		log.Fatal(err)
```
jhand2, line 50: If we error out because the requested `Support` isn't supported by the target, we should print a warning and then skip the test (instead of a fatal error).

verification/certifyKey_test.go
```diff
@@ -14,19 +15,43 @@ import (
 	"github.com/zmap/zlint/v3/lint"
 )
 
-// This file is used to test the certify key command by using a simulator
+// This file is used to test the certify key command by using a simulator/emulator
+
+func GetTestTarget_CertifyKey() ([]TestDPEInstance, error) {
```
jhand2, line 20: Take `Support` as an input parameter here to denote the support requested for the test.

jhand2, line 20: If you make all the below changes, I think this function should be reusable by all tests (instead of being specific to each command)

jhand2 (changes requested):

verification/certifyKey_test.go
steffy3494: GetTestTarget methods from each test file will invoke the GetEmulatorTarget which is a common function in verification_test.go and it is reused by all the command tests.

verification/certifyKey_test.go
steffy3494: Regarding the comment, my understanding is we need to create a seperate  test for each support instead of muliptle dpe instances with mulitple supports. I have added a snippet with my understanding. Shall I proceed with this.
	
	 For example:
		func TestCertifyKey_<name1>(t *testing.T){
			support_needed := []string{"AutoInit", "X509"}
			
			instance, err := GetTestTarget_CertifyKey(support_needed)
			if err != nil {
			...
			}

			testCertifyKey(instance, t)		
		}
		
		func TestCertifyKey_<name2>(t *testing.T){
			support_needed := []string{"AutoInit", "X509","Simulation"}
			
			instance, err := GetTestTarget_CertifyKey(support_needed)
			if err != nil {
			...
			}

			testCertifyKey(instance, t)		
		}
		
		-----------------------------------------------------------------
		
		func GetTestTarget_CertifyKey(support_needed []string) (TestDPEInstance, error) {
			if testTargetType == EMULATOR {
				//Create the dpe instance with the support_needed if it is a emulator
			} else if testTargetType == SIMULATOR {
				//Create the dpe instance with the support_needed if it is a simulator
			}
			return instance, nil
		}

verification/certifyKey_test.go
steffy3494: Addressed

verification/certifyKey_test.go
steffy3494: Addressed as per your comment

verification/certifyKey_test.go
steffy3494: Addressed

verification/initializeContext_test.go
```diff
@@ -6,62 +6,45 @@ import (
...
-			// No extra options.
-			&DpeSimulator{exe_path: *socket_exe},
-			// Supports simulation.
-			&DpeSimulator{exe_path: *socket_exe, supports: Support{Simulation: true}},
+	support_needed := []string{""}
+	instance, err := GetTestTarget(support_needed)
+	if err != nil {
+		if err.Error() == "Requested support is not supported in emulator" {
+			log.Print("Warning: Failed executing TestInitializeContext_EmptySupportSet command due to unsupported request. Hence, skipping it")
+			t.Skipf("Skipping the command execution")
```
jhand2, line 20: The string passed to Skipf will be printed in the test log anyway. We should merge these two lines so that the log message in just passed to Skipf

verification/initializeContext_test.go
```diff
@@ -6,62 +6,45 @@ import (
...
+		instance.SetLocality(locality)
+		testInitContext(instance, t)
+	}
 }
 
-func TestInitializeContext(t *testing.T) {
-
-	// Added dummy support for emulator
-	support_needed := []string{"AutoInit", "X509"}
+func TestInitializeContext_SupportSet_With_Simulation(t *testing.T) {
```
jhand2, line 32: Prefer naming tests based on what they're testing rather than the support test. For example. `TestInitializeContext_SimulationMode` would make sense here. We are testing whether simulation mode works. But if the target doesn't support simulation mode, then we just skip the test.

verification/initializeContext_test.go
```diff
@@ -6,62 +6,45 @@ import (
...
-	"golang.org/x/exp/slices"
 )
 
 // This file is used to test the initialize context command by using a simulator/emulator
 
-func GetTestTarget_InitializeContext(support_needed []string) ([]TestDPEInstance, error) {
-
-	var instances []TestDPEInstance
-	var err error
+func TestInitializeContext_EmptySupportSet(t *testing.T) {
```
jhand2, line 13: Can just name this `TestInitializeContext`

steffy3494: Modified the test function name based on the testing mode. But there are certain support sets which have a combination of 4 or 5 supports. For that, I have given a numeric value followed by the command name like below.

**Test Function name:**
									
**TestGetProfile_SupportMode_01** 			==>support_needed := []string{"Simulation", "AutoInit", "RotateContext", "Csr", "InternalDice","IsCA"}
**TestGetProfile_SupportMode_02**           ==>support_needed := []string{"ExtendTci", "Tagging", "X509", "InternalInfo"}
**TestGetProfile_AllSupportMode**           ==>support_needed := []string{"Simulation", "ExtendTci", "AutoInit", "Tagging","RotateContext",	"X509", "Csr", "IsSymmetric", "InternalInfo", "InternalDice", "IsCA"}

Are the changes okay for these function names?

verification/verification_test.go
```diff
@@ -3,10 +3,37 @@
...
+// This will be called before running tests, and it assigns the socket path based on command line flag.
+func TestMain(m *testing.M) {
+	testTarget := flag.String("target", "simulator", "socket type - emulator")
+	flag.Parse()
+	testTargetType = *testTarget
+	if testTargetType == SIMULATOR {
+		socket_exe = flag.String("sim", "../simulator/target/debug/simulator", "path to simulator executable")
+	} else if testTargetType == EMULATOR {
+		socket_exe = flag.String("emu", "../simulator/target/debug/emulator", "path to emulator executable")
+	}
```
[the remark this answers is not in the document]
jhand2: Oh never-mind, this is for the exe path to start the process. This is fine. I think socket_exe is just a confusing name. Maybe `target_exe`?

jhand2 (approved):

PR #154, opened 2023-08-23
Refactor derive_private_key and derive_ecdsa_pub

sree-revoori1:
This commit removes these two functions in favor of a derive_key_pair function which derives both private and public keys. This is necessary for the Caliptra implementation of the Crypto trait.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -142,7 +142,7 @@ pub trait Crypto {
         info: &[u8],
     ) -> Result<Self::Cdi, CryptoError>;
 
-    /// Derives a private key using a cryptographically secure KDF
+    /// Derives a key pair using a cryptographically secure KDF
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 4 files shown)


jhand2:
> Thoughts on passing some sort of an enum to derive_key_pair in order to only compute priv_key or pub_key? currently, in multiple places we call derive_key_pair to use only one of the priv key or pub key, so we could skip some unnecessary work with this.

Ya this makes sense to me. I think a bool would be fine. In Caliptra the public key always gets derived anyway. But in general deriving a public key isn't super fast, so it would be good to avoid deriving it if we can avoid it. And then you can return an `Option<EcdsaPubKey>`. If the pubkey wasn't requested, can just return None.

crypto/src/lib.rs
```diff
@@ -196,6 +183,7 @@ pub trait Crypto {
         algs: AlgLen,
         digest: &Digest,
         priv_key: &Self::PrivKey,
+        pub_key: EcdsaPub,
```
jhand2, line 186: Can you add the the comment what the public key is for? Something like

```rust
/// * `pub_key` - The public key corresponding to `priv_key`. An implementation may
///    optionally use pub_key to validate any generated signatures.
```

jhand2 (approved):

PR #157, opened 2023-08-24
Remove panic from DPE

sree-revoori1:
There are a few more places for potential panic in DPE which this PR fixes.

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -154,7 +154,11 @@ impl TryFrom<&[u8]> for ContextHandle {
             return Err(DpeErrorCode::InternalError);
         }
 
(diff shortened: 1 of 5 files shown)


dpe/src/context.rs
```diff
@@ -141,7 +141,7 @@ impl ContextHandle {
             return Err(DpeErrorCode::InternalError);
         }
 
-        dst[..ContextHandle::SIZE].copy_from_slice(&self.0);
+        dst[..ContextHandle::SIZE].copy_from_slice(&self.0[..ContextHandle::SIZE]);
```
jhand2, line 144: Is this necessary? `self.0` should already be this size.

jhand2 (approved):

PR #163, opened 2023-08-28
Refactor Support to use bitflags

sree-revoori1:
Also refactor flags passed to each DPE command to use bitflags instead of u32.

### DIFF
--- a/dpe/src/commands/derive_child.rs
+++ b/dpe/src/commands/derive_child.rs
@@ -7,6 +7,24 @@ use crate::{
     tci::TciMeasurement,
     DPE_PROFILE,
 };
+use bitflags::bitflags;
+
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq, zerocopy::FromBytes)]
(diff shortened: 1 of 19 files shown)


dpe/src/commands/derive_child.rs
```diff
@@ -7,53 +7,63 @@ use crate::{
...
+use bitflags::bitflags;
+
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq, zerocopy::FromBytes)]
+#[cfg_attr(test, derive(zerocopy::AsBytes))]
+pub struct DeriveChildFlags(u32);
+
+bitflags! {
+    impl DeriveChildFlags: u32 {
+        const INTERNAL_INPUT_INFO = 0b10000000_00000000_00000000_00000000;
```
jhand2, line 19: (here and elsewhere) I think it would be more clear to define these in the style `1u32 << 31`

jhand2 (approved):

PR #167, opened 2023-08-30
Draft PR requesting review for GetCertificateChain_test

hpya93:
Hi @jhand2,

- Added a new test file for GetCertificateChain command validation  (getCertificateChain_test.go)
- Added a new function to execute GetCertificateChain command  (client.go) 
- Could resolve the EOF issue that we discussed in last day's call
- Could get the certificate in command response and parse it.

**Queries:**
1. The GetCertificateChain command does not return _CertificateSize_ as returned by CertifyKey command.   
It is always 2048, I couldn't exclude the junk data at the end. I have tried to TrimRight("\x00") on cert chain, 
but I feel it is not correct do so. **Can you please let me know the current working of command is correct i.e always returning CertificateSize=2048?**

2. I observed that certificate chain returned by GetCertificateChain is in PEM format, but it's header, footer for PEM block type are missing i.e starts like "MIiJYqwdf...." without "----BEGIN CERTIFICATE----\nMIijYqw....At==\n-----END CERTIFICATE-----". Will it not cause trouble to demarcate the certificates in certificate chain?  **Can the command response send the cert chain with BEGIN CERTIFICATE and END CERTIFICATE?**

3. The certificate chain returned by command response has just one certificate and seems be a dummy certificate. There is no link between the DPE leaf cert returned by CertifyKey and  Issuer certificate returned by GetCertificateChain. **Is it possible to have a realistic cert chain and DPE leaf cert to properly validate the positive usecases like Issue validation, chain validation etc?**

4. I have added test for AutoInit support and Simulation support for GetCertificateCommand test. **Adding tests for aforementioned "support{}" is sufficient. Let me know if there are any other "support" to be added.**

6. Kindly share your feedback on the implementation so far..

### DIFF
--- a/verification/abi.go
+++ b/verification/abi.go
@@ -13,12 +13,13 @@ const (
 type CommandCode uint32
 
 const (
-	CommandGetProfile        CommandCode = 0x1
-	CommandInitializeContext CommandCode = 0x7
-	CommandCertifyKey        CommandCode = 0x9
-	CommandDestroyContext    CommandCode = 0xf
-	CommandTagTCI            CommandCode = 0x82
-	CommandGetTaggedTCI      CommandCode = 0x83
+	CommandGetProfile          CommandCode = 0x1
+	CommandInitializeContext   CommandCode = 0x7
+	CommandCertifyKey          CommandCode = 0x9
+	CommandDestroyContext      CommandCode = 0xf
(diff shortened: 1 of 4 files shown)


hpya93:
Missed some details

PR #169, opened 2023-08-30
Remove padding from TEST_CERT_CHAIN and make it DER encoded

sree-revoori1:


platform/src/default.rs
```diff
@@ -163,9 +77,16 @@ impl Platform for DefaultPlatform {
         if offset >= len {
             return Err(PlatformError::CertificateChainError);
         }
+
         let cert_chunk_range_end = min(offset + size, len);
-        out.copy_from_slice(&TEST_CERT_CHAIN[offset as usize..cert_chunk_range_end as usize]);
-        Ok(cert_chunk_range_end - offset)
+        let bytes_written = cert_chunk_range_end - offset;
+        if bytes_written as usize >= MAX_CHUNK_SIZE {
```
jhand2, line 83: If `bytes_written == MAX_CHUNK_SIZE` that should be ok. Just can't be larger.

jhand2 (approved):

Issue #74, opened 2023-04-24
Check certificates against zlint.


zhalvorsen:
We should check our certificates against [zlint](https://github.com/zmap/zlint) to make sure we are constructing the certificates correctly.

chrisfenner:
We could use this bug to track the following fixes in #75:

```
    verification_test.go:329: [ERROR] RFC5280: Certificates valid through the year 2049 MUST be encoded in UTC time (RFC 5280: 4.1.2.5)
    verification_test.go:329: [WARN] RFC5280: Sub certificates SHOULD include Subject Key Identifier in end entity certs (RFC 5280: 4.2 & 4.2.1.2)
    verification_test.go:329: [ERROR] RFC5280: CAs must support key identifiers and include them in all certificates (RFC 5280: 4.2 & 4.2.1.1)
    verification_test.go:329: [ERROR] RFC5280: CAs must include keyIdentifer field of AKI in all non-self-issued certificates (RFC 5280: 4.2.1.1)
```

jhand2:
3 & 4 should be solved by https://github.com/chipsalliance/caliptra-dpe/issues/80

PR #170, opened 2023-08-31
Added tests for certificate chain validation using GetCertificateChain command

hpya93:
Hi @jhand2 
Could you please review this? I took latest code and built the simulator afresh. The OIDs came as per spec.
- Have added test support for GetCertificateChain command
- Have added cert chain validation for intermediate CA certs collected from GetCertificateChain using x509.Verify() 
- Have added DPE leaf cert chain validation by calling GetCertificateChain

**Query 01**
- The cert chain validation fails if there any UnhandledCriticalExtension that crypto/x509  package doesnt know.
- It throws UnhandledCriticalExtensionError
- In code documentation they have suggested to overwrite that field with empty array if we know those extensions prior based on this comment [](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/crypto/x509/x509.go;l=714) 
- **Is this correct?**

**Query 02**
- Now I get "UnknownAuthorityError" on cert chain validation for leaf. It is because the SubjectName of cert I collect from GetCertificateChain Command doesnot match the IssuerName field of CertifyKey DPE leaf cert.
**- Can you please provide valid cert chain in response instead of dummy certs to validate positive cases?**

**Query 03**
- I see that the DPE leaf cert still returns errors in linting. Can we have a sample certificate from CertifyKey response without lint errors?
```
"e_ext_authority_key_identifier_missing":{
      "Status":Error (6),
      "Details":""
   } "e_ext_authority_key_identifier_no_key_identifier":{
      "Status":Error (6),
      "Details":""
   }"e_key_usage_incorrect_length":{
      "Status":Error (6),
      "Details":"the key usage ([128 1]) contains a value that is out of bounds of the range of possible KU values. (raw ASN: [3 3 0 128 1])"
   }"e_wrong_time_format_pre2050":{
      "Status":Error (6),
      "Details":""
   }

```

### DIFF
--- a/verification/client.go
+++ b/verification/client.go
@@ -198,8 +198,14 @@ func (c *Client[CurveParameter, Digest]) CertifyKey(cmd *CertifyKeyReq[Digest]) 
 }
 
 // GetCertificateChain calls the DPE GetCertificateChain command.
-func (c *Client[_, _]) GetCertificateChain(cmd *GetCertificateChainReq) (*GetCertificateChainResp, error) {
+func (c *Client[_, _]) GetCertificateChain() (*GetCertificateChainResp, error) {
 	var certs GetCertificateChainResp
+
+	// Initialize request input parameters
+	cmd := GetCertificateChainReq{
+		Offset: 0,
+		Size:   MAX_CHUNK_SIZE,
+	}
 
 	for {
 		respStruct := struct {
--- a/verification/client.go
+++ b/verification/client.go
@@ -3,7 +3,12 @@
 package verification
 
 import (
+	"errors"
 	"fmt"
+)
+
+const (
+	MAX_CHUNK_SIZE = 2048
 )
 
 type Support struct {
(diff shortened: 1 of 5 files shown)


jhand2:
> **Query 01**
> 
> * The cert chain validation fails if there any UnhandledCriticalExtension that crypto/x509  package doesnt know.
> * It throws UnhandledCriticalExtensionError
> * In code documentation they have suggested to overwrite that field with empty array if we know those extensions prior based on this comment
> * **Is this correct?**

We should first check that we know all of the OIDs for the unhandled extensions and make sure they are expected to be there. And then yes, overwriting with an empty array once they are known is the right thing to do.

> 
> **Query 02**
> 
> * Now I get "UnknownAuthorityError" on cert chain validation for leaf. It is because the SubjectName of cert I collect from GetCertificateChain Command doesnot match the IssuerName field of CertifyKey DPE leaf cert.
>   **- Can you please provide valid cert chain in response instead of dummy certs to validate positive cases?**

Thanks for pointing this out. We will fix this. I've opened this issue to track: https://github.com/chipsalliance/caliptra-dpe/issues/172

The caliptra emulator target has a valid cert chain once https://github.com/chipsalliance/caliptra-sw/pull/717 is merged. So that may be a good option for testing (once the task to hook up the emulator to these tests is done)

> 
> **Query 03**
> 
> * I see that the DPE leaf cert still returns errors in linting. Can we have a sample certificate from CertifyKey response without lint errors?
> 

This is being tracked in https://github.com/chipsalliance/caliptra-dpe/issues/74. We will work to address this, but it will not be right away. You can ignore the lint errors for now

verification/getCertificateChain_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	getCertificateChainResp, err := client.GetCertificateChain(&getCertificateChainReq)
+	if err != nil {
+		t.Fatalf("Could not get Certificate Chain: %v", err)
+	}
+
+	//fmt.Println(getCertificateChainResp)
+	checkCertificateChain(t, getCertificateChainResp.CertificateChain)
+
+	// TODO: When DeriveChild is implemented, call it here to add more TCIs and call GetCertificateChain again.
```
jhand2, line 73: This isn't relevant to GetCertificateChain in the same way it is for CertifyKey

verification/client.go
```diff
@@ -192,6 +197,41 @@ func (c *Client[CurveParameter, Digest]) CertifyKey(cmd *CertifyKeyReq[Digest])
 	}, nil
 }
 
+// GetCertificateChain calls the DPE GetCertificateChain command.
+func (c *Client[_, _]) GetCertificateChain(cmd *GetCertificateChainReq) (*GetCertificateChainResp, error) {
```
jhand2, line 201: The goal of this function is just to get the full cert chain, so I don't think you need to pass in GetCertificateChainReq. Instead this function can internally start from offset 0 and read until it gets to the end.

jhand2 (commented):
Overall this looks good. Left a couple suggestions.

verification/getCertificateChain_test.go
hpya93: Did you mean the comment? Yes I removed it.

verification/client.go
hpya93: Yes, moved it into the client function

hpya93:
@jhand2 
Hi, I have raised a PR based on your comments, can you please review and merge if ok? 
Thanks!

PR #173, opened 2023-09-01
Make alias cert match crypto algorithm

sree-revoori1:
This change adds a test cert created with a p384 key and makes platform/crypto run based on the algorithm being used. Also, we make TEST_ISSUER equal to the subject name of TEST_CERT_CHAIN so cert validation can work.


platform/src/default.rs
```diff
@@ -10,11 +10,14 @@ pub const VENDOR_ID: u32 = 0;
 pub const VENDOR_SKU: u32 = 0;
 
 // Generated with openssl using the command openssl req -new -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 -x509 -nodes -days 365 -out cert.pem -keyout cert.pem
-// Converted to der format using the command openssl x509 -in cert.pem -outform DER -out cert.der
+// Pasted key into a new file key_256.pem
+// Converted cert to der format using the command openssl x509 -in cert.pem -outform DER -out cert.der
+// Converted key to der format using the command openssl ec -in key_256.pem -outform DER -out key_256.der
```
jhand2, line 15: nit: Maybe a shell script in `test_data` for re-generating all of the test data would be useful. Then we can update these comments to just point to the shell script.

crypto/Cargo.toml
```diff
@@ -8,6 +8,10 @@ edition = "2021"
 [features]
 openssl = ["dep:openssl", "dep:hkdf", "dep:sha2"]
 deterministic_rand = ["openssl"]
+default = ["dpe_profile_p256_sha256"]
+dpe_profile_p256_sha256 = []
+dpe_profile_p384_sha384 = []
```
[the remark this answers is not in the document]
sree-revoori1: Done.

platform/src/default.rs
sree-revoori1: A good portion of generating the test data was manual work (cut-pasting the private key into a new file, running the haskell script to format, and then removing quotations manually). I will look into a way to automate all of the generation via commands and then we can make the shell script.

platform/src/default.rs
```diff
@@ -57,7 +60,57 @@ pub const TEST_CERT_CHAIN: [u8; 613] = [
     0x5a, 0x64, 0xd7, 0xe1, 0x9a,
 ];
 
-// Generated manually using the x509 issuer name encoder in x509.rs.
+// Same steps as above except ran openssl req -new -newkey ec -pkeyopt ec_paramgen_curve:secp384r1 -x509 -nodes -days 365 -out cert_384.pem -keyout cert_384.pem
+#[cfg(feature = "dpe_profile_p384_sha384")]
+pub const TEST_CERT_CHAIN: [u8; 674] = [
```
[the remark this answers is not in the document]
sree-revoori1: Thanks, thats way cleaner.

platform/src/default.rs
jhand2: Ok. If it ends up being a lot of work, we can also do it in a follow-up PR

sree-revoori1: Actually, I figured out a way to do it, I just had to split up generating the pkey and the cert.

platform/src/default.rs
```diff
@@ -9,61 +9,22 @@ pub const AUTO_INIT_LOCALITY: u32 = 0;
...
-// Generated manually using the x509 issuer name encoder in x509.rs.
-const TEST_ISSUER: [u8; 99] = [
-    48, 97, 49, 20, 48, 18, 6, 3, 85, 4, 3, 19, 11, 84, 101, 115, 116, 32, 73, 115, 115, 117, 101,
-    114, 49, 73, 48, 71, 6, 3, 85, 4, 5, 19, 64, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97,
-    97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97,
-    97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97, 97,
-    97, 97, 97, 97,
+// Equal to the subject name of TEST_CERT_CHAIN
+// Extracted subject name of TEST_CERT_CHAIN by running X509::from_der(include_bytes!("../platform/src/test_data/cert_384.der")).unwrap().subject_name().to_der().unwrap()
+const TEST_ISSUER: [u8; 136] = [
```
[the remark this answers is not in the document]
jhand2: Sorry, I tough I sent this review last week, but I had it sitting as a draft.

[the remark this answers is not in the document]
sree-revoori1: Took a different approach, and did this in platform itself by depending on openssl. Looks like there is an issue with actually importing openssl though.

platform/src/default.rs
```diff
@@ -90,11 +41,16 @@ impl Platform for DefaultPlatform {
     }
 
     fn get_issuer_name(&mut self, out: &mut [u8; MAX_CHUNK_SIZE]) -> Result<usize, PlatformError> {
-        if TEST_ISSUER.len() >= out.len() {
+        let issuer_name = X509::from_pem(include_bytes!("test_data/cert_256.pem"))
+            .unwrap()
+            .subject_name()
+            .to_der()
+            .unwrap();
+        if issuer_name.len() >= out.len() {
```
jhand2, line 49:
```suggestion
        if issuer_name.len() > out.len() {
```

jhand2 (approved):

PR #174, opened 2023-09-05
X509 fields and extensions validation

hpya93:
Hi @jhand2 
**Recent status**
I have updated the PR [#174](https://github.com/chipsalliance/caliptra-dpe/pull/174) with following items as per your comments.

 
Updated dpe/src/x509.rs for OID fix
Updated dpe/src/x509.rs for ADDISFLAG fix (setting 30th bit instead of 31st bit)
Updated the MultiTcb information parsing using cryptobye library instead of struct tags to be more accurate and deterministic (certifyKey_test.go).
With this, I could get "PASS" for all the tests. Kindly review the same.
Thanks!

  
**Initial description** 
I have added the x509 validation support in addition to GetCertificateChain test.
I have followed the dpe-irot profile spec and annotated the validation checks in function comments.
I have suppressed fatal errors due to certs in response, I have marked them as TODO, and shall replace once they are expected to pass the test.

Please let me know your feedback.
I couldnt validate MultiTcbInfo extension with certificate returned, I took reference from spec and tried to build a struct in golang to unmarshal ASN1

Kindly let me know your feedback

Thanks!

 Basic constraint extension parsing fails in cert returned by CertifyKey when ADDISCA flag is set to true.
 I have mailed the issue details
 Please let me know if Im missing anything soo that we can finalize the changes made for this PR.
 
 Thank you


verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		t.Logf("Intermediate certificates chain validation is done")
+	}
+}
+
+func buildVerifyOptions(certChain []*x509.Certificate) x509.VerifyOptions {
+	roots := x509.NewCertPool()
+	intermediates := x509.NewCertPool()
+
+	for _, cert := range certChain {
+		if cert.Subject.String() == cert.Issuer.String() {
```
jhand2, line 571: The root should be the highest cert in the chain. In DPE, the root is the certificate which occurs first in the GetCertificateChain response. All the other certs returned are intermediates.

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	if foundError {
+		// TODO: Fail the test when the certificate is expected to pass
+		t.Logf("Found certificates with mismatching signature")
+	}
+}
+
+// Build certificate chain and calls to validateSignature on each chain.
+func validateCertChain(t *testing.T, certChain []*x509.Certificate, leafCert *x509.Certificate) {
+	t.Helper()
+	if leafCert != nil {
```
jhand2, line 504: A lot of this code looks to be duplicated. Can we move the if statement to only do the `leafCert` calls if `leafCert != nil`, but otherwise keep the other code common.

Specifically, I think lines 510-518 are the only ones that differ.

verification/certifyKey_test.go
```diff
@@ -44,18 +100,260 @@ func TestCertifyKey_SimulationMode(t *testing.T) {
...
+	}
+}
+
+// The KeyUsage extension MUST contain only DigitalSignature
+// If IsCA, KeyUsage extension MUST contain DigitalSignature and KeyCertSign
+func checkCertifyKeyExtensions(t *testing.T, c *x509.Certificate) {
+	t.Helper()
+
+	//Check for keyusage extension
+	if c.IsCA {
```
jhand2, line 278: This code can mostly be shared. The only difference is allowedKeyUsages

verification/certifyKey_test.go
```diff
@@ -44,18 +100,260 @@ func TestCertifyKey_SimulationMode(t *testing.T) {
...
+		}
+	}
+
+	if len(extKeyUsage) == 0 {
+		t.Fatalf("The Extended Key Usage extension is empty")
+	}
+
+	// Iterate over the OIDs in the ExtKeyUsage extension
+	isExtendedKeyUsageValid := false
+	if c.IsCA {
```
jhand2, line 234: Only difference between these branches is which OID to look for

verification/client.go
```diff
@@ -3,9 +3,14 @@
 package verification
 
 import (
+	"errors"
 	"fmt"
 )
 
+const (
+	MAX_CHUNK_SIZE = 2048
```
jhand2, line 11: nit: Same style comment as above

verification/client.go
```diff
@@ -192,6 +197,47 @@ func (c *Client[CurveParameter, Digest]) CertifyKey(cmd *CertifyKeyReq[Digest])
...
+		_, err := execCommand(c.transport, CommandGetCertificateChain, c.Profile, cmd, &respStruct)
+		if err == StatusInvalidArgument {
+			// This indicates that there are no more bytes to be read in certificate chain
+			break
+		} else if err != nil {
+			// This indicates error in processing GetCertificateChain command
+			return nil, err
+		}
+		// fmt.Println(respStruct.CertificateChain[:])
+		// fmt.Println(string(respStruct.CertificateChain[:]))
```
jhand2, line 225: nit: remove

verification/certifyKey_test.go
```diff
@@ -3,23 +3,79 @@
...
 	"github.com/zmap/zlint/v3/lint"
+
+	"golang.org/x/exp/slices"
+)
+
+type CertifyKeyFlag int
+
+const (
+	// RESERVED_0 CertifyKeyFlag = 31
+	ADD_IS_CA CertifyKeyFlag = 30
```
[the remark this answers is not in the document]
hpya93: Yes this is changed.

verification/certifyKey_test.go
```diff
@@ -3,23 +3,79 @@
...
+	ExtensionTcgDiceKpIdentityInit  = "2.23.133.5.4.100.6"  // OID for tcg-dice-kp-identityInit
+	ExtensionTcgDiceKpIdentityLoc   = "2.23.133.5.4.100.7"  // OID for tcg-dice-kp-identityLoc
+	ExtensionTcgDiceKpAttestInit    = "2.23.133.5.4.100.8"  // OID for tcg-dice-kp-attestInit
+	ExtensionTcgDiceKpAttestLoc     = "2.23.133.5.4.100.9"  // OID for tcg-dice-kp-attestLoc
+	ExtensionTcgDiceKpAssertInit    = "2.23.133.5.4.100.10" // OID for tcg-dice-kp-assertInit
+	ExtensionTcgDiceKpAssertLoc     = "2.23.133.5.4.100.11" // OID for tcg-dice-kp-assertLoc
+	ExtensionTcgDiceKpEca           = "2.23.133.5.4.100.12" // OID for tcg-dice-kp-eca
+)
+
+var TCG_DICE_CRITICAL_EXTENSIONS = [...]string{
```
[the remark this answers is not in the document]
hpya93: Modified to TcgDiceCriticalExtensions.

verification/certifyKey_test.go
```diff
@@ -3,23 +3,79 @@
...
+	ExtensionTcgDiceMultiTcbInfo,
+	ExtensionTcgDiceUeid,
+	ExtensionTcgDiceKpIdentityLoc,
+	ExtensionTcgDiceKpAttestLoc,
+}
+
+var certifyKeyReq CertifyKeyReq[SHA256Digest]
```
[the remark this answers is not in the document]
hpya93: Passed the required fields as args, removed the global variable.

verification/certifyKey_test.go
```diff
@@ -44,18 +100,260 @@ func TestCertifyKey_SimulationMode(t *testing.T) {
...
+	now := time.Now().UTC()
+	if now.Before(c.NotBefore) {
+		t.Fatalf("Certificate is not valid yet, current time : %s, not before: %s", now, c.NotBefore)
+	} else if now.After(c.NotAfter) {
+		t.Fatalf("Certificate is expired, current time : %s, not after: %s", now, c.NotAfter)
+	}
+}
+
+// The KeyUsage extension MUST contain only DigitalSignature
+// If IsCA, KeyUsage extension MUST contain DigitalSignature and KeyCertSign
```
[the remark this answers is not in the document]
hpya93: Modified comment

verification/certifyKey_test.go
```diff
@@ -44,18 +100,260 @@ func TestCertifyKey_SimulationMode(t *testing.T) {
...
+	flagsBuf := &bytes.Buffer{}
+	binary.Write(flagsBuf, binary.LittleEndian, certifyKeyReq.Flags)
+
+	flagIsCA := extractFlagBit(int(ADD_IS_CA), uint32(certifyKeyReq.Flags))
+
+	if flagIsCA == c.IsCA {
+		t.Logf("ADD_IS_CA is set to %v and the basic constraint IsCA is set to %v", flagIsCA, c.IsCA)
+	} else {
+		// Fail when ADD_IS_CA flag is set properly in input parameters
+		t.Logf("ADD_IS_CA is set to %v but the basic constraint IsCA is set to %v", flagIsCA, c.IsCA)
```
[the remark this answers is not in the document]
hpya93: Yes, this was a mistake

verification/certifyKey_test.go
hpya93: Removed duplication

verification/certifyKey_test.go
hpya93: Have cut down the duplicate code

verification/certifyKey_test.go
hpya93: Done

verification/client.go
hpya93: Changed to MaxChunkSize

verification/client.go
hpya93: Removed stray comments

verification/certifyKey_test.go
```diff
@@ -44,18 +100,260 @@ func TestCertifyKey_SimulationMode(t *testing.T) {
...
+		}
+	} else {
+		allowedKeyUsages := x509.KeyUsageDigitalSignature
+		certKeyUsageList := getKeyUsageNames(c.KeyUsage)
+		allowedKeyUsageList := getKeyUsageNames(allowedKeyUsages)
+
+		if c.KeyUsage == allowedKeyUsages {
+			t.Logf("The leaf Certificate has ONLY the expected key usage %v ", certKeyUsageList)
+		} else {
+			t.Fatalf("The leaf Certificate has got %v but want %v ", certKeyUsageList, allowedKeyUsageList)
```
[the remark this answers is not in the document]
hpya93: Removed Fatalf wherever unnecessary

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
 	certifyKeyReq := CertifyKeyReq[SHA256Digest]{
 		ContextHandle: [16]byte{0},
 		Flags:         0,
```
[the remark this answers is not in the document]
hpya93: The if condition below covers for both test cases.

> 
> 	if flagIsCA == c.IsCA {
> 		t.Logf("ADD_IS_CA is set to %v and the basic constraint IsCA is set to %v", flagIsCA, c.IsCA)
> 	} else {
> 		// Fail when ADD_IS_CA flag is set properly in input parameters
> 		t.Errorf("ADD_IS_CA is set to %v but the basic constraint IsCA is set to %v", flagIsCA, c.IsCA)
> 	}

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	validateCertifyKeyFields(t, leafCert)
+
+	// Validate that proper TcbInfo extensions are populated with expected measurements and values as defined in the DPE iRoT profile
+
+	// Validate correct KeyUsage and BasicConstraints extensions for all DICE certificates
+
+	// Ensure full certificate chain has valid signatures
+	validateCertChain(t, certChain, leafCert)
 
 	// TODO: When DeriveChild is implemented, call it here to add more TCIs and call CertifyKey again.
```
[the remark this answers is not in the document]
hpya93: The TODO is not removed because the DeriveChildTest is not implemented, it is yet to be implemented, so I still kept that comment

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		t.Fatalf("Could not obtain certificate chain from GetCertificateChain command.")
+	}
+	t.Logf("Certificate chain is DER encoded")
+
+	// Validate that all X.509 fields conform with the format defined in the DPE iRoT profile
+	validateCertifyKeyFields(t, leafCert)
+
+	// Validate that proper TcbInfo extensions are populated with expected measurements and values as defined in the DPE iRoT profile
+
+	// Validate correct KeyUsage and BasicConstraints extensions for all DICE certificates
```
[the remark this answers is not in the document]
hpya93: This was a misleading comment, the checks were added in a function call, removed the comments

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	//err := root.CheckSignatureFrom(root)
+	err := root.CheckSignature(root.SignatureAlgorithm, root.RawTBSCertificate, root.Signature)
+	if err != nil {
+		t.Logf("Root certificate  %s Signature validation failed with error %s", root.Subject.String(), err.Error())
+		foundError = true
+	}
+	t.Logf("Root certificate Signature validation passed %s\n", root.Subject.String())
+
+	if foundError {
+		// TODO: Fail the test when the certificate is expected to pass
```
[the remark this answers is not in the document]
hpya93: Signature validation of leaf certificate fails, so added TODO.
I have added comments for that.

verification/certifyKey_test.go
```diff
@@ -124,15 +423,210 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		t.Log("Checking for unhandled critical certificate extensions unknown to DPE certificates profile spec...")
+		removeTcgDiceCriticalExtensions(t, []*x509.Certificate{leafCert})
+
+		// Build verify options
+		opts := buildVerifyOptions(certChain)
+
+		// Certificate chain validation for leaf
+		chains, err := leafCert.Verify(opts)
+		if err != nil {
+			// TODO: Fail the test with Errorf here once we expect it to pass.
```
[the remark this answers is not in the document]
hpya93: Yes updated comments.
Here are the consolidated list of errors ignored for now, will be enabled when certs are received proper format.

1. Signature validation of leaf certificate fails
2. Certificate chain cannot be built from leaf to root
3. When DeriveChild is implemented, call it here to add more TCIs and call CertifyKey again.
4. multiTcb info is not provided in leaf
5. Linter error
6. Ueid extn value doen not match the label	(as per irot spec saying "UEID shall be populated by "label"  ")

verification/certifyKey_test.go
hpya93: Ok, 
Added an additional check if cert in 0th index is a root cert.

verification/certifyKey_test.go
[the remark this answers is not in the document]
jhand2: Right, but it seems to me that we only actually run a test where ADD_IS_CA is false. Am I missing something?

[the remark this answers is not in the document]
hpya93: Ok, are you trying to convey that, the certifyKey request  should be built with different "flag" field (Flag="30th bit unset", Flag="30th bit set") and tested as two different functions "testCertifyKeyWithAddCA()", without "testCertifyKeyWithoutAddCA() instead of single testCertifyKey()"?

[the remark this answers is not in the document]
hpya93: Which way would be the best? - a test function for each input combination / single function for all input parameter combinations?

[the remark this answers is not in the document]
hpya93: Reg the set/unset ADD_IS_CA flag, I thought it could be implemented as part of third activity where the combination of input params are dealt.
To address your comment, I have created an array of inputs (CertifyKeyRequests), as we usually do in Go tests - iterated  validated.
Since, setting ADD_IS_CA=True in CertifyKeyReq.Flags does not make isCA=True in certificate, so I have logged instead of throwing error.
I have added a TODO comment mentioning this issue.

verification/certifyKey_test.go
```diff
@@ -121,18 +404,221 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+			t.Logf("Signature validation failed for certificate %s in chain with error %s", certchain[i].Subject.String(), err.Error())
+			foundError = true
+		} else {
+			t.Logf("Signature validation passed %s", certchain[i].Subject.String())
+		}
+
+	}
+
+	root := certchain[len(certchain)-1]
+	//err := root.CheckSignatureFrom(root)
```
jhand2, line 471: This would just check that the root is self-signed, which is not a requirement. This should be removed.

verification/certifyKey_test.go
```diff
@@ -121,18 +404,221 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	err := root.CheckSignature(root.SignatureAlgorithm, root.RawTBSCertificate, root.Signature)
```
jhand2, line 472: Same. This is checking that the root is self-signed.

verification/certifyKey_test.go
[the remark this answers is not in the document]
jhand2: I don't see anywhere in this function where the leaf is being validated

verification/certifyKey_test.go
```diff
@@ -121,18 +404,221 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+			}
+
+			// Log certificate chains linked to each cetificate in chain
+			for _, chain := range chains {
+				for i, cert := range chain {
+					t.Logf("%d %s", i, (*cert).Subject)
+				}
+				// Validate signature of all certificate chains
+				validateSignature(t, chain)
+			}
```
jhand2, line 542: Same as above, cert.Verify will do signature checks

verification/certifyKey_test.go
```diff
@@ -121,18 +404,221 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	}
+
+	// Remove unhandled critical extensions reported by x509 but defined in spec
+	t.Log("Checking for unhandled critical certificate extensions unknown to DPE certificates profile spec...")
+	removeTcgDiceCriticalExtensions(t, certsToProcess)
+
+	// Build verify options
+	opts := buildVerifyOptions(t, certChain)
+
+	if leafCert != nil {
```
jhand2, line 505: I don't think leafCert should ever be nil here, right? This is the CertifyKey test, so there should always be a leaf.

verification/certifyKey_test.go
hpya93: Yes, removed the redundant signature checks.

verification/certifyKey_test.go
hpya93: Was removed along with the redundant function,

verification/certifyKey_test.go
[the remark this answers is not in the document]
hpya93: THis was part of redundant code and it is removed now .

verification/certifyKey_test.go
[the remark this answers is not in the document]
hpya93: Reg 4; the spec says it would return array of DiceTcbInfo, but we receive single DiceTcbInfo, is this expected?

Reg 6: This SUCCEEDS if processed as per your comment, I just compared Label.Equals(UeidExtnValue), 

Reg 7: Yes, when ADDISCA is set 1, isCA is still false

> 
> 	certifyKeyReq := []CertifyKeyReq[SHA256Digest]{
> 		{
> 			ContextHandle: [16]byte{0},
> 			Flags:         0,
> 			Label:         [32]byte{},
> 			Format:        CertifyKeyX509,
> 		},
> 		{
> 			ContextHandle: [16]byte{0},
> 			Flags:         1073741824, // Set the 31st bit: 01000000 00000000 00000000 00000000
> 			Label:         [32]byte{},
> 			Format:        CertifyKeyX509,
> 		},
> 	}

	
For nonzero flags also, I receive only isCAFalse, so the test fails. I have enabled error, you may notice this is go test.

verification/certifyKey_test.go
hpya93: Yes removed

verification/certifyKey_test.go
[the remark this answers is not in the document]
hpya93: Item 7 related reference : https://github.com/hpya93/caliptra-dpe/blob/main/verification/certifyKey_test.go#L371

verification/certifyKey_test.go
hpya93: This function validates leaf+certchain (from CertifyKey) OR certchain (from GetCertChain). So if leafcert is passed, it is inferred we need validate leaf+certchain.

verification/certifyKey_test.go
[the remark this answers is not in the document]
hpya93: **Reg your question  -- whether the leaf cert validation succeeds now?**
No, because key usage in cert chain is empty - not listed and this causes leaf to fail with "Invalid key usage"

Please find the decoded certs I collected from "go test --verbose".

```
haripriya@haripriyab:~$ openssl x509 -inform PEM -in k.pem -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            62:30:37:30:65:64:36:31:39:39:39:31:39:38:39:39:38:33:62:63
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: C = US, ST = Illinois, L = Chicago, O = Google, OU = PSS, CN = Sree, emailAddress = srrevoori@google.com
        Validity
            Not Before: Feb 27 00:00:00 2023 GMT
            Not After : Dec 31 23:59:59 9999 GMT
        Subject: CN = DPE Leaf, serialNumber = b070ed619991989983bc305ff9beb69c67e9e8895ed5417f846af4cf483dc120
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:29:78:6b:1a:85:23:9b:53:e4:e7:22:4b:2b:35:
                    87:68:e9:06:a6:ee:3c:12:28:64:32:83:c0:cf:6a:
                    70:bf:ed:1a:a8:b5:50:ce:e9:22:c7:08:99:79:d4:
                    0f:37:e4:15:92:d9:c3:53:1b:8e:67:20:c9:9f:e4:
                    92:b9:d5:d8:3d
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            2.23.133.5.4.5: critical
                0n0l.^0-..`.H.e..... ................................0-..`.H.e..... ............................................
            2.23.133.5.4.4: critical
                0". ................................
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Key Usage: critical
                Digital Signature
            X509v3 Extended Key Usage: critical
                2.23.133.5.4.100.9
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:45:02:20:39:a0:b8:d3:e3:e2:88:9d:e5:74:e7:89:ef:36:
        a9:81:c6:d3:f2:5f:5b:d5:ea:21:dd:3b:c7:39:b4:73:e2:29:
        02:21:00:bd:f1:1c:c3:41:3b:13:1f:38:ea:7e:d7:f5:c9:0b:
        80:7d:8c:c3:ca:1d:88:48:7e:1c:ea:49:00:c6:ff:30:9a
-----BEGIN CERTIFICATE-----
MIIC2TCCAn+gAwIBAgIUYjA3MGVkNjE5OTkxOTg5OTgzYmMwCgYIKoZIzj0EAwIw
gYUxCzAJBgNVBAYTAlVTMREwDwYDVQQIDAhJbGxpbm9pczEQMA4GA1UEBwwHQ2hp
Y2FnbzEPMA0GA1UECgwGR29vZ2xlMQwwCgYDVQQLDANQU1MxDTALBgNVBAMMBFNy
ZWUxIzAhBgkqhkiG9w0BCQEWFHNycmV2b29yaUBnb29nbGUuY29tMCIYDzIwMjMw
MjI3MDAwMDAwWhgPOTk5OTEyMzEyMzU5NTlaMF4xETAPBgNVBAMTCERQRSBMZWFm
MUkwRwYDVQQFE0BiMDcwZWQ2MTk5OTE5ODk5ODNiYzMwNWZmOWJlYjY5YzY3ZTll
ODg5NWVkNTQxN2Y4NDZhZjRjZjQ4M2RjMTIwMFkwEwYHKoZIzj0CAQYIKoZIzj0D
AQcDQgAEKXhrGoUjm1Pk5yJLKzWHaOkGpu48EihkMoPAz2pwv+0aqLVQzukixwiZ
edQPN+QVktnDUxuOZyDJn+SSudXYPaOB7jCB6zB9BgZngQUFBAUBAf8EcDBuMGym
XjAtBglghkgBZQMEAgEEIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
MC0GCWCGSAFlAwQCAQQgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACI
BAAAAACJBAAAAAAwMQYGZ4EFBQQEAQH/BCQwIgQgAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAwDwYDVR0TAQH/BAUwAwEBADAPBgNVHQ8BAf8EBQMDAIAB
MBUGA1UdJQEB/wQLMAkGB2eBBQUEZAkwCgYIKoZIzj0EAwIDSAAwRQIgOaC40+Pi
iJ3ldOeJ7zapgcbT8l9b1eoh3TvHObRz4ikCIQC98RzDQTsTHzjqftf1yQuAfYzD
yh2ISH4c6kkAxv8wmg==
-----END CERTIFICATE-----
haripriya@haripriyab:~$ openssl x509 -inform PEM -in l.pem -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            44:e2:18:b7:a0:92:51:0a:a9:3b:8e:fc:c0:82:be:4d:6c:ca:24:b4
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: C = US, ST = Illinois, L = Chicago, O = Google, OU = PSS, CN = Sree, emailAddress = srrevoori@google.com
        Validity
            Not Before: May 31 23:43:29 2023 GMT
            Not After : May 30 23:43:29 2024 GMT
        Subject: C = US, ST = Illinois, L = Chicago, O = Google, OU = PSS, CN = Sree, emailAddress = srrevoori@google.com
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:4e:0d:2f:e1:79:c3:ea:97:07:58:6f:d4:78:b7:
                    15:01:df:1a:3c:ea:f7:7b:9f:fe:b8:a6:77:b1:3c:
                    c7:21:34:3c:e6:96:16:5b:b3:7b:5c:82:76:cf:39:
                    93:50:ef:d6:ee:df:4d:49:b0:8e:fb:e7:56:37:2b:
                    2b:ed:7f:a1:80
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Subject Key Identifier: 
                69:6F:35:61:11:12:7B:C6:90:BD:E4:42:4C:29:4B:80:2C:E9:37:44
            X509v3 Authority Key Identifier: 
                69:6F:35:61:11:12:7B:C6:90:BD:E4:42:4C:29:4B:80:2C:E9:37:44
            X509v3 Basic Constraints: critical
                CA:TRUE
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        30:45:02:20:63:e9:7d:94:41:91:65:a8:ab:55:81:30:8d:b9:
        c7:16:5c:09:86:d5:c1:eb:01:c6:cb:01:0c:d9:25:43:76:57:
        02:21:00:84:0d:b5:96:8e:32:64:b4:b5:5f:39:00:7d:07:a3:
        ec:e2:56:71:8b:97:32:c0:69:9b:b4:b8:5a:64:d7:e1:9a
-----BEGIN CERTIFICATE-----
MIICYTCCAgegAwIBAgIUROIYt6CSUQqpO478wIK+TWzKJLQwCgYIKoZIzj0EAwIw
gYUxCzAJBgNVBAYTAlVTMREwDwYDVQQIDAhJbGxpbm9pczEQMA4GA1UEBwwHQ2hp
Y2FnbzEPMA0GA1UECgwGR29vZ2xlMQwwCgYDVQQLDANQU1MxDTALBgNVBAMMBFNy
ZWUxIzAhBgkqhkiG9w0BCQEWFHNycmV2b29yaUBnb29nbGUuY29tMB4XDTIzMDUz
MTIzNDMyOVoXDTI0MDUzMDIzNDMyOVowgYUxCzAJBgNVBAYTAlVTMREwDwYDVQQI
DAhJbGxpbm9pczEQMA4GA1UEBwwHQ2hpY2FnbzEPMA0GA1UECgwGR29vZ2xlMQww
CgYDVQQLDANQU1MxDTALBgNVBAMMBFNyZWUxIzAhBgkqhkiG9w0BCQEWFHNycmV2
b29yaUBnb29nbGUuY29tMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETg0v4XnD
6pcHWG/UeLcVAd8aPOr3e5/+uKZ3sTzHITQ85pYWW7N7XIJ2zzmTUO/W7t9NSbCO
++dWNysr7X+hgKNTMFEwHQYDVR0OBBYEFGlvNWEREnvGkL3kQkwpS4As6TdEMB8G
A1UdIwQYMBaAFGlvNWEREnvGkL3kQkwpS4As6TdEMA8GA1UdEwEB/wQFMAMBAf8w
CgYIKoZIzj0EAwIDSAAwRQIgY+l9lEGRZairVYEwjbnHFlwJhtXB6wHGywEM2SVD
dlcCIQCEDbWWjjJktLVfOQB9B6Ps4lZxi5cywGmbtLhaZNfhmg==
-----END CERTIFICATE-----

```

verification/certifyKey_test.go
```diff
@@ -3,61 +3,521 @@
...
 	if err != nil {
 		if err.Error() == "Requested support is not supported in the emulator" {
-			t.Skipf("Warning: Failed executing TestCertifyKey_SimulationMode command due to unsupported request. Hence, skipping the command execution")
+			t.Skipf("[WARNING]: Failed executing TestCertifyKey_SimulationMode command due to unsupported request. Hence, skipping the command execution")
 		} else {
-			log.Fatal(err)
+			log.Fatalf("[FATAL]: %s", err.Error())
 		}
 	}
 	testCertifyKey(instance, t)
```
jhand2, line 141: This doesn't appear to actually be testing in simulation mode (it just tests that the target supports simulation mode). To test this, we should initializing a simulation context and calling CertifyKey on it.

verification/certifyKey_test.go
```diff
@@ -3,61 +3,521 @@
...
+				t.Errorf("[ERROR]: Failed to unmarshal MultiTcbInfo field: %v", err)
+			} else {
+				t.Logf("[LOG]: MultiTcb extension %v", multiTcbInfo)
+			}
+			break
+		}
+	}
+}
+
+func parseMultiTcbInfo(der []byte) (*TcgMultiTcbInfo, error) {
```
jhand2, line 265: We should avoid manually parsing ASN.1. The Go ASN.1 package provides type annotations, for example

```go
type diceTCBInfo struct {
	Vendor     string           `asn1:"optional,implicit,tag:0,utf8"`
	Model      string           `asn1:"optional,implicit,tag:1,utf8"`
	Version    string           `asn1:"optional,implicit,tag:2,utf8"`
	SVN        int              `asn1:"optional,implicit,tag:3"`
	Layer      int              `asn1:"optional,implicit,tag:4"`
	Index      int              `asn1:"optional,implicit,tag:5"`
	FWIDs      []fwid           `asn1:"optional,implicit,tag:6"`
	Flags      operationalFlags `asn1:"optional,implicit,tag:7"`
	VendorInfo []byte           `asn1:"optional,implicit,tag:8"`
	Type       []byte           `asn1:"optional,implicit,tag:9"`
}

type multiTCBInfo struct {
	DICETCBInfoSeq []diceTCBInfo
}
```

Then you can parse it with `asn1.Unmarshal()`

verification/certifyKey_test.go
```diff
@@ -3,61 +3,521 @@
...
+	expectedKeyUsageName := ""
+	if c.IsCA {
+		expectedKeyUsage = OidExtensionTcgDiceKpEca
+		expectedKeyUsageName = "tcg-dice-kp-eca"
+	} else {
+		expectedKeyUsage = OidExtensionTcgDiceKpAttestLoc
+		expectedKeyUsageName = "tcg-dice-kp-attest-loc"
+	}
+
+	for _, oid := range extKeyUsage {
```
jhand2, line 426: can do `if slices.Contains(extKeyUsage, expectedKeyUsage)`

verification/certifyKey_test.go
```diff
@@ -118,21 +579,194 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		// chain of intermediate certificates
+		t.Logf("[LOG]: Intermediate certificates chain validation is done")
+	}
+}
+
+func buildVerifyOptions(t *testing.T, certChain []*x509.Certificate) x509.VerifyOptions {
+	roots := x509.NewCertPool()
+	intermediates := x509.NewCertPool()
+
+	t.Logf("[LOG]: Root certificate is expected to be in the beginning of the chain, the rest are expected to be intermediates.")
```
jhand2, line 704: nit: This can just be a comment, don't need to log this.

verification/certifyKey_test.go
```diff
@@ -118,21 +579,194 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+	if certChain[0].Subject.String() != certChain[0].Issuer.String() {
+		t.Errorf("[ERROR]: Found a non-root certificate in beginning of certificate chain returned by GetCertificateChain.")
+	} else {
```
jhand2, line 707: It is not a requirement that the root is self-signed. That just happens to be true for the current targets.

verification/certifyKey_test.go
```diff
@@ -118,21 +579,194 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		roots.AddCert(certChain[0])
+	}
+
+	for _, cert := range certChain[1:] {
+		if cert.Subject.String() == cert.Issuer.String() {
+			t.Errorf("[ERROR]: Found a Root certificate in middle of certificate chain returned by GetCertificateChain.")
```
jhand2, line 713: nit
```suggestion
			t.Errorf("[ERROR]: Found a self-signed certificate in middle of certificate chain returned by GetCertificateChain.")
```

verification/certifyKey_test.go
```diff
@@ -118,21 +579,194 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+				for i, cert := range chain {
+					t.Logf("%d %s", i, (*cert).Subject)
+				}
+			}
+		}
+
+		// This indicates that signature validation found no errors each cert
+		// chain of intermediate certificates
+		t.Logf("[LOG]: Intermediate certificates chain validation is done")
+	}
```
jhand2, line 697: I would prefer to move this check to the GetCertificateChain test so that the test code is easier to follow. Suggested refactor:

```go
func validateCertChain(...) {
	// Remove unhandled critical extensions reported by x509 but defined in spec
	removeTcgDiceCriticalExtensions(t, certsToProcess)

	// Remove unhandled extended key usages reported by x509 but defined in spec
	removeTcgDiceExtendedKeyUsages(t, certsToProcess)

	// Build verify options
	opts := buildVerifyOptions(t, certChain)
		
         // Certificate chain validation for each intermediate certificate
	for _, cert := range certChain {
		chains, err := cert.Verify(opts)
		if err != nil {
			t.Errorf("[ERROR]: Error in Certificate Chain of %s: %s", cert.Subject, err.Error())
		}
		// Log certificate chains linked to each cetificate in chain
		t.Logf("[LOG]: Chains of intermediate certificate.")
		for _, chain := range chains {
			for i, cert := range chain {
				t.Logf("%d %s", i, (*cert).Subject)
			}
		}
	}

	// This indicates that signature validation found no errors each cert
	// chain of intermediate certificates
	t.Logf("[LOG]: Intermediate certificates chain validation is done")
}
```

Then the code here can be specifically for the leaf key

verification/certifyKey_test.go
```diff
@@ -118,21 +579,194 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
+		// Certificate chain validation for leaf
+		chains, err := leafCert.Verify(opts)
+		if err != nil {
+			// Certificate chain cannot be built from leaf to root
+			t.Errorf("[ERROR]: Error in certificate chain %s: ", err.Error())
+		}
+
+		// Log certificate chains linked to leaf
+		t.Logf("[LOG]: Chains of DPE leaf certificate.")
+		for _, chain := range chains {
```
jhand2, line 667: Instead of looping over chains, can just check that `len(chains) == 1`

verification/certifyKey_test.go
hpya93: I started out with this method. I works great if the data is not nested and has universal tags. 
Since we have some context specific tags, and it always resulted in tag mismatch errors. I even debugged into asn1 lib to understand and fix this. 
I came across some blogs suggesting that x509.parsecertificate that uses cryptobyte library is more efficient.
https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/crypto/x509/parser.go;l=997 
Cryptobyte seemed more promising to deal with arrays and context-specific data types 
So switched to it. 

I see that using crypto byte provides more control to specify constructed, context specific tags.

hpya93: Do you still think this needs to be done with struct tags?

verification/certifyKey_test.go
steffy3494: Tried using slices, but it is throwing an error: 'encoding/asn1'.ObjectIdentifier does not satisfy comparability.'

verification/certifyKey_test.go
steffy3494: Addressed

verification/certifyKey_test.go
jhand2: Ok, this is fine to leave then

verification/certifyKey_test.go
jhand2: The latest commit just moves this whole function to GetCertificateChain. To clarify, we still want to verify the leave cert chains to the root. I just think they can be split into two functions. GetCertificateChain test has the above code. Then CertifyKey test has

```go
func validateLeafCert(...) {
    	// Remove unhandled critical extensions reported by x509 but defined in spec
	removeTcgDiceCriticalExtensions(t, certsToProcess)

	// Remove unhandled extended key usages reported by x509 but defined in spec
	removeTcgDiceExtendedKeyUsages(t, certsToProcess)

	// Build verify options
	opts := buildVerifyOptions(t, certChain)

	// Certificate chain validation for leaf
	chains, err := leafCert.Verify(opts)
	if err != nil {
		// Certificate chain cannot be built from leaf to root
		t.Errorf("[ERROR]: Error in certificate chain %s: ", err.Error())
	}
	
        // Log certificate chains linked to leaf
	if len(chains) == 0 {
		t.Errorf("[ERROR]: certificate chain is empty")
	}
}
```

verification/certifyKey_test.go
```diff
@@ -118,21 +579,148 @@ func testCertifyKey(d TestDPEInstance, t *testing.T) {
...
 	client, err := NewClient256(d)
 	if err != nil {
-		t.Fatalf("Could not initialize client: %v", err)
+		t.Fatalf("[FATAL]: Could not initialize client: %v", err)
+	}
+
+	var ctx ContextHandle
+	var initCtxResp *InitCtxResp
+
+	if d.GetSupport().Simulation {
```
jhand2, line 588: This still just checks if the DPE instance supports simulation contexts, not whether the test is expecting to use simulation contexts. I would expect `testCertifyKey` to accept a boolean `use_simulation`

verification/certifyKey_test.go
jhand2: The struct tagging mechanism supports implicit/explicit tags. I don't see what here can't be achieved with annotated structs.

The manual parsing code is a very large amount of code to maintain. I would greatly prefer the struct annotations.

verification/getCertificateChain_test.go
```diff
@@ -0,0 +1,184 @@
...
+		// Certificate chain validation for leaf
+		chains, err := leafCert.Verify(opts)
+		if err != nil {
+			// Certificate chain cannot be built from leaf to root
+			t.Errorf("[ERROR]: Error in certificate chain %s: ", err.Error())
+		}
+
+		// Log certificate chains linked to leaf
+		t.Logf("[LOG]: Chains of DPE leaf certificate.")
+		if len(chains) == 0 {
```
jhand2, line 158: It is also invalid if len(chains) > 1. The cert chain should only contain one chain from leaf to root.

verification/getCertificateChain_test.go
```diff
@@ -0,0 +1,184 @@
...
+		chains, err := leafCert.Verify(opts)
+		if err != nil {
+			// Certificate chain cannot be built from leaf to root
+			t.Errorf("[ERROR]: Error in certificate chain %s: ", err.Error())
+		}
+
+		// Log certificate chains linked to leaf
+		t.Logf("[LOG]: Chains of DPE leaf certificate.")
+		if len(chains) != 1 {
+			t.Errorf("[ERROR]: certificate chain is empty")
```
jhand2, line 159: nit: This log message should be different if `len(chains) > 1`

jhand2 (commented):
I think this is ready to be merged once the remaining open comments are addressed

verification/getCertificateChain_test.go
hpya93: Modified the log message as below
t.Errorf("[ERROR]: There are %d certificate chains but there should be exactly one certificate chain.", len(chains))

verification/certifyKey_test.go
hpya93: Hi Jordan, I have updated my code to rely only on struct tags.
It works fine now.
The issue came while defining the Multi Tcb Info improperly, it should have been an array not a struct it seems. 
Now it seems to be fixed. Thanks for insisting on this.

verification/certifyKey_test.go
hpya93: Have split the cert chain validation function as suggested.

verification/certifyKey_test.go
hpya93: This check is removed.

jhand2 (approved):

PR #180, opened 2023-09-12
Initial profile changes

premalathak12:
This Pr contains the initial changes for testing profile command.


verification/emulator.go
```diff
@@ -20,12 +20,12 @@ const (
 
 	DPE_EMULATOR_AUTO_INIT_LOCALITY    uint32  = 0
 	DPE_EMULATOR_OTHER_LOCALITY        uint32  = 0
-	DPE_EMULATOR_PROFILE               Profile = 0
-	DPE_EMULATOR_MAX_TCI_NODES         uint32  = 0
+	DPE_EMULATOR_PROFILE               Profile = ProfileP256SHA384
```
jhand2, line 23: This should be ProfileP384SHA384

verification/verification_test.go
```diff
@@ -99,13 +99,14 @@ func GetTestTarget(support_needed []string) (TestDPEInstance, error) {
...
-	value := reflect.ValueOf(DpeEmulator{}.supports)
+	/*value := reflect.ValueOf(DpeEmulator{}.supports)
 	for i := 0; i < len(support_needed); i++ {
 		support := reflect.Indirect(value).FieldByName(support_needed[i])
 		if !support.Bool() {
 			return nil, errors.New("Error in creating dpe instances - supported feature is not enabled in emulator")
 		}
-	}
+	}*/
+	// TODO : Get the supported modes from emulator and then check.
```
zhalvorsen, line 109: Would implementing this be a large enough effort to put in a separate PR? I think it might be okay to do as part of this PR.

zhalvorsen: I might be misunderstanding how this works. But could you do the same as `GetSimulatorTarget` below? i.e.

```Rust
	value := reflect.ValueOf(DpeEmulator{}.supports)
	fields := reflect.Indirect(value)
	fVal := reflect.New(reflect.TypeOf(DpeEmulator{}.supports))
	for i := 0; i < len(support_needed); i++ {
		for j := 0; j < value.NumField(); j++ {
			if fields.Type().Field(j).Name == support_needed[i] {
				fVal.Elem().Field(j).SetBool(true)
			}
		}
	}
```

jhand2: In general I agree we should be able to use Zach's code to do this in this PR. This will also require calling GetProfile, but I think that can be done as long as you construct the emulator and then do this check.

premalathak12: Updated the code, based on the comments provided.

jhand2 (changes requested):
Please don't duplicate huge portions of the test code just to support a different algorithm. It should be fairly simple to just refactor a couple spots to use different request/response structures.

verification/emulator.go
```diff
@@ -126,11 +127,17 @@ func (s *DpeEmulator) waitForPower(on bool) bool {
...
 	if err != nil {
 		return nil, err
 	}
 
+	var i int
+	for i = 0; i < len(buf); i++ {
+		fmt.Print(buf[i])
+		fmt.Print(",")
+	}
+
```
jhand2, line 140: nit: Should be removed

verification/verification_test.go
```diff
@@ -98,15 +101,36 @@ func GetTestTarget(support_needed []string) (TestDPEInstance, error) {
...
+
+	value := reflect.ValueOf(support.ToSupport(rsp.Flags))
 	for i := 0; i < len(support_needed); i++ {
 		support := reflect.Indirect(value).FieldByName(support_needed[i])
 		if !support.Bool() {
 			return nil, errors.New("Error in creating dpe instances - supported feature is not enabled in emulator")
 		}
 	}
-	var instance TestDPEInstance = &DpeEmulator{exe_path: *target_exe}
+	instance.SetSupport(support)
```
jhand2, line 133: `Support` cannot be set on the emulator. Either the target supports what is needed or it doesn't. So I think this line can be removed and then the SetSupport target API can be removed.

clundin25:
This looks very stale, going to close it

PR #181, opened 2023-09-15
Test fail cases defined in iRotProfile

steffy3494:
Here, Pulled the X509 changes(hpya93/caliptra-dpe/tree/main) and committed my changes above on that code.

**My changes:** Added fail cases - 10a85df2699b6a0dfc24930ba70353769d174564

**PR Description:** 

Verified the failure test cases for the commands other than the derivechild (which has not been implemented yet) and certifykey (which has dependencies on derivechild flags).

In initialize_context, I added a fail check to the below point, but I am unable to verify the case as the simulation contexts are always supported when the simulation flag is set true.

'_simulation contexts are not supported and the SIMULATION flag is set; this command shall fail with status DPE_STATUS_ARGUMENT_NOT_SUPPORTED_'

Also, added a failure check in the getcertificatechain command.

In the PR, I have added these two codes. Other fail checks are already present in the code which I have tested.

**Query:**

In fail cases of getcertificatechain command, spec says '_If OFFSET is larger than the full certificate chain, this API shall return DPE_STATUS_INVALID_ARGUMENT_'

But,

The full cert chain size is computed in runtime, by repeated calls to GetCertificateChain()

So, we can't check if offset > fullcertchainsize

This can be checked only if we pass a negative test input like

cases := []GetCertificateReq{
{size: 2048, offset: 100000},
{size: 2048, offset: 100000}, // Negative case.
}

So, should we write a negative test case like this?


sree-revoori1:
> In initialize_context, I added a fail check to the below point, but I am unable to verify the case as the simulation contexts are always supported when the simulation flag is set true.
> 
> '_simulation contexts are not supported and the SIMULATION flag is set; this command shall fail with status DPE_STATUS_ARGUMENT_NOT_SUPPORTED_'

The InitializeContext command takes in a Simulation flag. If simulation is not set in the DPE Instance's support and we call InitializeContext with the simulation flag set to true, we should expect a failure with error DPE_STATUS_ARGUMENT_NOT_SUPPORTED.

steffy3494:
Tried the below initialize_context command failure case approach based on @sree-revoori1 comment. 

'_simulation contexts are not supported, and the SIMULATION flag is set; this command shall fail with status DPE_STATUS_ARGUMENT_NOT_SUPPORTED_'

And now, I am able to verify it. Thanks.

steffy3494:
I have added details about other validation points below. Please take a look at it along with this PR.

1. Validate proper behavior to against the feature support matrix (exposed by GetProfile)

- 	For the above sow point, Validated the getprofile modes for both simulator and emulator using code from this PR:

	https://github.com/chipsalliance/caliptra-dpe/pull/180/files

2. Add tests to ensure localities are properly isolated from each other for all commands.

- 	For the above sow point, I have done testing on separate localities in simulator side (TestInitializeContext using multiple localities)

- 	The code for separating localities is already there in the command file. Just checked that the initialize context is called for each locality.		

- 	In emulator side, we are not setting any locality to the DPE instance when executing commands.

PR #183, opened 2023-09-17
Add functions to validate a DPE Instance

sree-revoori1:
This commit adds a function that validates a DPEInstance's context array, to ensure it represents a properly-formed tree. We also add functions to retrieve the current tci from a context, and get the index of a context whose locality and tci_type are equal to some parameters. These functions will be called from caliptra rt-fw when it validates the DPE instance loaded from SRAM upon a warm reset.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -389,6 +389,49 @@ impl DpeInstance {
 
         hasher.finish().map_err(|_| DpeErrorCode::HashError)
     }
+
+    /// Determines if the context array represents a valid tree by checking that
+    /// there is only 1 connected component and that all nodes lead up to
+    /// the root node.
+    ///
+    /// # Arguments
+    ///
+    /// * `root_idx` - The index of the root context
+    pub fn validate_context_tree(&self, root_idx: usize) -> bool {
+        let mut seen = [false; MAX_HANDLES];
+
+        // dfs from the root node and try to discover invalid subtrees
+        if self.detect_invalid_subtree(root_idx, &mut seen) {
+            return false;
+        }
+
+        for (i, node_visited) in seen.iter().enumerate().take(MAX_HANDLES) {
+            // If a node was not seen when doing a dfs from the root, there must be multiple
+            // connected components or the root is not actually the root
+            if i != root_idx && self.contexts[i].state != ContextState::Inactive && !node_visited {
+                return false;
+            }
+        }
+        true
+    }
+
+    fn detect_invalid_subtree(&self, curr_idx: usize, seen: &mut [bool; MAX_HANDLES]) -> bool {
+        // if the current node was already visited we have a cycle
+        if curr_idx >= MAX_HANDLES
+            || self.contexts[curr_idx].state == ContextState::Inactive
+            || seen[curr_idx]
+        {
+            return true;
+        }
+        seen[curr_idx] = true;
+        // dfs on all child nodes
+        for child_idx in flags_iter(self.contexts[curr_idx].children, MAX_HANDLES) {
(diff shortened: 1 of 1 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -389,6 +389,83 @@ impl DpeInstance {
...
+    /// # Arguments
+    ///
+    /// * `idx` - desired index of the root context
+    /// * `locality` - desired locality of the root context
+    pub fn get_root_idx(&self, tci_type: u32, locality: u32) -> Result<usize, DpeErrorCode> {
+        // find indices of all non-inactive contexts matching tci_type and locality
+        let mut tci_nodes = self.contexts.iter().enumerate().filter(|(_, context)| {
+            context.state != ContextState::Inactive
+                && context.locality == locality
+                && context.tci.tci_type == tci_type
```
jhand2, line 405: tci_type isn't guaranteed to be unique, so I don't think we can use this for searching for a specific node. What is the use-case for this function during validation?

dpe/src/dpe_instance.rs
```diff
@@ -389,6 +389,83 @@ impl DpeInstance {
...
+            // If a node was not seen when doing a dfs from the root, there must be multiple
+            // connected components or the root is not actually the root
+            if i != root_idx && self.contexts[i].state != ContextState::Inactive && !node_visited {
+                return false;
+            }
+        }
+        true
+    }
+
+    fn detect_cycle(&self, curr_idx: usize, seen: &mut [bool; MAX_HANDLES]) -> bool {
```
jhand2, line 452: naming nit: This doesn't just detect cycles, it also detects paths to inactive nodes. Maybe `detect_invalid_subtree`?

dpe/src/dpe_instance.rs
```diff
@@ -389,6 +389,83 @@ impl DpeInstance {
...
+    }
+
+    /// Determines if the context array represents a valid tree by checking that
+    /// there is only 1 connected component and that all nodes lead up to
+    /// the root node.
+    ///
+    /// # Arguments
+    ///
+    /// * `root_idx` - The index of the root context
+    pub fn validate_context_tree(&self, root_idx: usize) -> bool {
```
jhand2, line 434: Ideally we would test this on a large number of inputs for how the Context array is populated. But this is a hard function to fuzz because the test has no way to check whether the answer this function gives is the correct answer for the generated output.

I don't have a solution, and we don't have to solve in this PR, just something to think about for how we could improve testing here.

dpe/src/dpe_instance.rs
sree-revoori1: This is to find the root node with tci_type "RTJM". I think the root node is the only node with this tci_type.

dpe/src/dpe_instance.rs
sree-revoori1: Done.

dpe/src/dpe_instance.rs
sree-revoori1: On second thought, I think this function along with get_current_tci are pretty niche and would be better off on the caliptra side, so I deleted them from this PR.

dpe/src/dpe_instance.rs
sree-revoori1: Thanks for mentioning this, I made https://github.com/chipsalliance/caliptra-dpe/issues/204 to track.

dpe/src/dpe_instance.rs
jhand2: Ok, sounds good. The caller to DeriveChild can pick any value they want for the type, including "RTJM". I think another way we could find the root node is the find an Active node whose parent is 0xFF (no parent) and that there is only one such node.

jhand2 (approved):

PR #185, opened 2023-09-18
dpe/fuzz: Fix bit-rot

benjamindoron:


jhand2 (commented):
LGTM. Can we also build the fuzz target (no need to run it) in the CI script so we catch things like this before merging PRs? https://github.com/chipsalliance/caliptra-dpe/blob/d3b9102af176338292e76914eb2c56e87406cb79/ci.sh#L6

benjamindoron:
> LGTM. Can we also build the fuzz target (no need to run it) in the CI script so we catch things like this before merging PRs?
> 
> https://github.com/chipsalliance/caliptra-dpe/blob/d3b9102af176338292e76914eb2c56e87406cb79/ci.sh#L6

Oh. We have https://github.com/chipsalliance/caliptra-sw/pull/700, but that'll only trigger when someone bumps the submodule revision. That'd potentially be a lot of work if DPE went through many changes. Okay

jhand2 (approved):

ci.sh
```diff
@@ -12,6 +12,14 @@ set -ex
   cargo fmt -- --check
   cargo clippy -- --deny=warnings
 )
+( cd dpe/fuzz
+  rustup toolchain install nightly-2023-04-15
+  cargo +nightly-2023-04-15 install cargo-fuzz cargo-afl
+  cargo fmt --check
+  cargo clippy
```
jhand2, line 19: Looks like clippy needs one of the features enabled to succeed

benjamindoron: Oh, right. May as well check both of them, this crate includes different code depending on features.

jhand2 (approved):

PR #186, opened 2023-09-19
Refactor DPE commands to avoid mutating state before erroring

### LINKED ISSUE #63: Be able to cleanly handle errors that happen late in a command.
If an error happens late in a command, the state of the system could have undefined behavior.

Here is a single example, but there are several other cases which could have similar problems.

In the `DeriveChild` command, we generate a child, add it to the parent's list of children, activate it, and then add the TCI measurement given in the command. If the hashing engine fails while adding the TCI measurement, we immediately cascade the error upwards. This would leave a dangling child context that nothing has a handle to.

We should be able to handle errors such that the state of the DPE instance in unchanged. The easiest way to accomplish this would to copy out the `Context`, modify it, and then commit it back to the DPE instance. This will allow us to return at any point in the command without affecting the persistent state of the DPE instance.

### LINKED ISSUE #148: Check for InternalError in fuzz tests
In DPE, we use InternalError to check for cases that should never happen (e.g. context data is in a bad/unrecoverable state). We should check for this in fuzz tests and fail if we see this error.

### LINKED ISSUE #149: Check for state change in fuzz tests
Internal state in DpeInstance should only change if a command succeeds. We should have the fuzz tests check for cases where

* DPE command failed
* Any state in DpeInstance changed

And raise a fatal error.

sree-revoori1:
Also, adds checks to the fuzzer where we error if a DPE command returns an InternalError or if DPE state changes and a command returns any error.

Fixes #63 
Fixes #149 
Fixes #148


dpe/fuzz/src/fuzz_target_1.rs
```diff
@@ -82,6 +84,12 @@ fn harness(data: &[u8]) {
     };
     // There are a few vendor error codes starting at 0x1000, so this can be a 2 bytes.
     trace!("| Response Code {response_code:#06x}");
+    if dpe.contexts != prev_contexts && response_code != 0 {
+        trace!("Error: DPE state changes upon a failed DPE command.");
+    }
+    if response_code == DpeErrorCode::InternalError as u32 {
+        trace!("Error: DPE reached a state that should be unreachable.");
+    }
```
jhand2, line 92: I think we need to panic here (instead of trace) to make sure the fuzzer reports the error correctly.

dpe/src/commands/derive_child.rs
```diff
@@ -142,13 +142,7 @@ impl CommandExecution for DeriveChildCmd {
...
         };
 
-        if !self.retains_parent() {
-            dpe.contexts[parent_idx].state = ContextState::Retired;
-            dpe.contexts[parent_idx].handle = ContextHandle([0xff; ContextHandle::SIZE]);
-        } else if !dpe.contexts[parent_idx].handle.is_default() {
-            dpe.contexts[parent_idx].handle = dpe.generate_new_handle(env)?;
-        }
-
+        let unactivated_child_context = dpe.contexts[child_idx];
```
jhand2, line 145: I think it would be more clear what's happening here if we make a `Context::default`, populate it, and then at the end, do `dpe.contexts[child_idx] = tmp_context;`

This allows you to still use the `?` mechanism for returns.

dpe/src/commands/extend_tci.rs
```diff
@@ -29,10 +29,18 @@ impl CommandExecution for ExtendTciCmd {
         }
 
         let idx = dpe.get_active_context_pos(&self.handle, locality)?;
+        let tci_current = dpe.contexts[idx].tci.tci_current;
+        let tci_cumulative = dpe.contexts[idx].tci.tci_cumulative;
         dpe.add_tci_measurement(env, idx, &TciMeasurement(self.data), locality)?;
```
jhand2, line 34: As a general rule, it might be good to

* Make a copy of the context
* Modify it
* Set `contexts[idx] = tmp` at the end of the function

We can change `add_tci_measurement` to take a `&Context` instead of an index.

dpe/fuzz/src/fuzz_target_1.rs
sree-revoori1: Done

dpe/src/commands/extend_tci.rs
sree-revoori1: I took a slightly different approach, where I copy out the context, and then restore contexts[idx] = copy if we run into an error after mutating contexts[idx]. The reason for this is that if we mutate the copied context instead in extend_tci for example, we would also have to change roll_onetime_use_handle to take in a context instead of an idx, and I felt this would be too invasive since it would lead to changing all calls to roll_onetime_use_handle and add_tci_measurement.

I think this way it is equally clear, but the downside is that we can't just use the ? mechanism to return.

dpe/src/commands/derive_child.rs
sree-revoori1: Done.

dpe/src/commands/extend_tci.rs
sree-revoori1: Disregard what I previously said, I followed your approach. I didn't want to also make roll_onetime_use_handle take in a context instead of a handle, since then we would need to update all callers of this function, so I took another approach where I mutate everything but dpe.contexts[idx].handle after we know extend_tci is sure to succeed.

dpe/src/commands/derive_child.rs
```diff
@@ -159,7 +154,22 @@ impl CommandExecution for DeriveChildCmd {
...
+            dpe.contexts[parent_idx].handle = ContextHandle([0xff; ContextHandle::SIZE]);
+        } else if !dpe.contexts[parent_idx].handle.is_default() {
+            dpe.contexts[parent_idx].handle = dpe.generate_new_handle(env)?;
+        }
+
+        // at this point we cannot error out anymore, so it is safe to set the child context
+        dpe.contexts[child_idx] = tmp_context;
 
         // Add child to the parent's list of children.
         dpe.contexts[parent_idx].add_child(child_idx)?;
```
jhand2, line 175: If this fails, the parent_idx state is still modified. I think we need to make updating the parent `state`, `handle`, and `children` atomic.

sree-revoori1: Refactored it to also copy the parent context into a temporary context. Also, added the same logic to sign and destroy_context to avoid mutation upon InternalErrors.

jhand2 (approved):

dpe/src/commands/destroy_context.rs
```diff
@@ -54,16 +54,38 @@ impl CommandExecution for DestroyCtxCmd {
             1 << idx
         };
 
+        // contexts_to_destroy[i] == true implies dpe.contexts[i] must be destroyed
+        let mut contexts_to_destroy = [false; MAX_HANDLES];
+        // updated_children[i] holds i's non-destroyed children.
+        let mut updated_children = [0u32; MAX_HANDLES];
```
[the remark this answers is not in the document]
sree-revoori1: Done

jhand2 (approved):

PR #198, opened 2023-09-20
Remove infinite loop possibility from ChildtoRoot iter

### LINKED ISSUE #192: ChildToRootIter Infinite Loop
NCC report page 27

sree-revoori1:
Previously, if we used the ChildToRootIter with a leaf index >= MAX_HANDLES, we would infinite loop since the done flag
is not set. This change sets the done flag so that we can terminate if we see an index out of bounds.

Fixes #192

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -262,6 +262,7 @@ impl<'a> Iterator for ChildToRootIter<'a> {
             return Some(Err(DpeErrorCode::MaxTcis));
         }
         if self.idx >= self.contexts.len() {
+            self.done = true;
             return Some(Err(DpeErrorCode::InternalError));
         }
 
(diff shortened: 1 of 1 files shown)


zhalvorsen (approved):

dpe/src/context.rs
```diff
@@ -382,6 +383,19 @@ mod tests {
...
 
+    #[test]
+    fn test_child_to_root_iter_infinite_loop() {
+        let contexts = [CONTEXT_INITIALIZER; MAX_HANDLES];
+        let mut i = 0;
+        for _ in ChildToRootIter::new(30, &contexts) {
+            i += 1;
+            // fail test if we iterate over all nodes without terminating
+            if i > MAX_HANDLES {
+                assert!(false);
```
jhand2, line 394: nit: Prefer `panic!("Some useful message")`. This makes debugging test failures easier.

sree-revoori1: Done

jhand2 (approved):

PR #199, opened 2023-09-20
Use constant_time_eq to compare ContextHandles

### LINKED ISSUE #188: Timing Side-Channel When Comparing Context Handles
See NCC report, page 19

sree-revoori1:
Currently, we are using a non-constant time comparator for ContextHandles which is a security issue since attackers can determine the correct content handle in a byte-by-byte manner by repeatedly observing the execution time of get_active_context_pos with varying inputs.

Fixes #188

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -12,6 +12,7 @@ use crate::{
     tci::{TciMeasurement, TciNodeData},
     U8Bool, DPE_PROFILE, INTERNAL_INPUT_INFO_SIZE, MAX_HANDLES,
 };
+use constant_time_eq::constant_time_eq;
 use crypto::{Crypto, Digest, Hasher};
 use platform::{Platform, MAX_CHUNK_SIZE};
 use zerocopy::{AsBytes, FromBytes};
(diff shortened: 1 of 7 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -151,27 +152,27 @@ impl DpeInstance {
...
             .peekable();
-        if valid_handles.peek().is_none() {
-            return Err(DpeErrorCode::InvalidHandle);
+        if valid_localities.peek().is_none() {
+            return Err(DpeErrorCode::InvalidLocality);
         }
-        let mut valid_handles_and_localities = valid_handles
-            .filter(|(_, context)| context.locality == locality)
+        let mut valid_handles_and_localities = valid_localities
+            .filter(|(_, context)| constant_time_eq(&context.handle.0, &handle.0))
```
zhalvorsen, line 167: Just for curiosity's sake, did you switch the order here to make it harder for an attacker to see how many total active contexts there are? The most they could see is how many active contexts are in their locality?

zhalvorsen (commented):
LGTM pending any comments on Jordan's chat

dpe/src/dpe_instance.rs
```diff
@@ -151,27 +152,27 @@ impl DpeInstance {
...
         locality: u32,
     ) -> Result<usize, DpeErrorCode> {
-        let mut valid_handles = self
+        let mut valid_localities = self
             .contexts
             .iter()
             .enumerate()
             .filter(|(_, context)| {
-                context.state == ContextState::Active && &context.handle == handle
+                context.state == ContextState::Active && context.locality == locality
```
jhand2, line 160: I think it should be ok to just filter on `context.state == ContextState::Active && context.locality == locality && constant_time_eq(&context.handle.0, &handle.0))` right? And then if `valid_handles` size is != 1, we can return an error that the context state is corrupted.

sree-revoori1: The reason I initially split this into two iterators was to differentiate the errors better. Note how we return InvalidLocality/ InvalidHandle based on which filter renders an empty iterator.

dpe/src/dpe_instance.rs
sree-revoori1: I'm not totally sure about the reasoning, it was just one of NCC's recommendations.

dpe/src/dpe_instance.rs
jhand2: Ok, that's a good reason. In that case, I would just suggest adding comments to each block explaining what checks are being done in each part and why they're separated.

sree-revoori1: Done

dpe/src/dpe_instance.rs
jhand2: This does reduce the number of handles that you need to do a handle comparison on, so I think it's more of a performance optimization than a security optimization.

jhand2 (dismissed):

jhand2 (approved):

PR #200, opened 2023-09-20
Make destroy context clear destroyed context's parent's child bit

sree-revoori1:
Also add tests for destroy context.

### DIFF
--- a/dpe/src/commands/destroy_context.rs
+++ b/dpe/src/commands/destroy_context.rs
@@ -1,7 +1,7 @@
 // Licensed under the Apache-2.0 license.
 use super::CommandExecution;
 use crate::{
-    context::ContextHandle,
+    context::{Context, ContextHandle},
     dpe_instance::{flags_iter, DpeEnv, DpeInstance, DpeTypes},
     response::{DpeErrorCode, Response, ResponseHdr},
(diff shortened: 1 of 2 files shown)


dpe/src/commands/destroy_context.rs
```diff
@@ -91,4 +99,144 @@ mod tests {
...
+            }
+            .execute(&mut dpe, &mut env, TEST_LOCALITIES[0])
+        );
+        assert_eq!(dpe.contexts[0].state, ContextState::Inactive);
+
+        activate_dummy_context(
+            &mut dpe,
+            0,
+            Context::ROOT_INDEX,
+            &ContextHandle([0; ContextHandle::SIZE]),
```
zhalvorsen, line 158: nit: it might be worthwhile to use `ContextHandle::default()` here to make it obvious which one is being destroyed later.

zhalvorsen (approved):

dpe/src/commands/destroy_context.rs
sree-revoori1: Done.

zhalvorsen (approved):

jhand2 (approved):

PR #201, opened 2023-09-21
Remove usage of unsafe Rust in CryptoBuf

### LINKED ISSUE #196: CryptoBuf Can Be Partially Initialized
NCC report page 33

sree-revoori1:
Previously, we were unsafely using ArrayVec::set_len in the CryptoBuf implementation. We were also not checking that the size of the input slice bytes == algs.size() which resulted in partially initialized CryptoBufs. This change fixes both of these issues.

Now, there are no more usages of unsafe Rust in DPE.

Fixes #196

### DIFF
--- a/crypto/src/signer.rs
+++ b/crypto/src/signer.rs
@@ -28,16 +28,16 @@ impl EcdsaPub {
 pub type HmacSig = CryptoBuf;
 
 /// A common base struct that can be used for all digests, signatures, and keys.
+#[derive(Debug, PartialEq, Eq)]
 pub struct CryptoBuf(ArrayVec<u8, { Self::MAX_SIZE }>);
 
 impl CryptoBuf {
     pub const MAX_SIZE: usize = AlgLen::MAX_ALG_LEN_BYTES;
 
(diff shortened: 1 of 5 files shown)


crypto/src/signer.rs
```diff
@@ -37,16 +37,20 @@ impl EcdsaPub {
 pub type HmacSig = CryptoBuf;
 
 /// A common base struct that can be used for all digests, signatures, and keys.
+#[derive(Debug, PartialEq, Eq)]
 pub struct CryptoBuf(ArrayVec<u8, { Self::MAX_SIZE }>);
 
 impl CryptoBuf {
     pub const MAX_SIZE: usize = AlgLen::MAX_ALG_LEN_BYTES;
 
     pub fn new(bytes: &[u8], algs: AlgLen) -> Result<CryptoBuf, CryptoError> {
```
jhand2, line 46: It's a little odd that `algs` is passed into this function at all. If `bytes` is less than the max size of the CryptoBuf, I'm wondering if we should just create it. All off the callsites that use it need to check the size anyway. What do you think?

sree-revoori1: SGTM. Done.

jhand2 (approved):

PR #202, opened 2023-09-21
Set internal input flags on child contexts instead of parent

### LINKED ISSUE #187: Changes in Context Tree Affecting Behaviour in Other Branches
See NCC Caliptra Security Report, page 16.

sree-revoori1:
Previously we were setting the internal_dice and internal_info flags on the parent context in derive child. This is a bug since attestation could fail under the following scenario:

1. Assume context[0] is auto-initialized to the default context
2. Call DeriveChild with retain parent
3. Call Sign on the default context (context[0])
4. Call DeriveChild with retain parent and internal_input_info.
5. Call CertifyKey on the default context (context[0])

Attestation fails here (we cannot verify the signature with the public key given to us by CertifyKey) since step 2 does not set the internal_input_flag so in step 3, when computing CDIs we do not mix in internal input info. However, in step 4, we would then incorrectly set context[0]'s internal_input_flag so when we certify key in step 5 and derive CDIs, the internal input info is mixed in, leading to a signature verification failure.

Now we make derive child set the internal_input flags on the child context in order to avoid this problem. I have also added a test for this scenario.

We also make deterministic rand_bytes use a seed so that successive calls return different results. This helps for testing.

Fixes #187

### DIFF
--- a/dpe/src/commands/derive_child.rs
+++ b/dpe/src/commands/derive_child.rs
@@ -120,9 +120,6 @@ impl CommandExecution for DeriveChildCmd {
             .get_next_inactive_context_pos()
             .ok_or(DpeErrorCode::MaxTcis)?;
 
-        dpe.contexts[parent_idx].uses_internal_input_info = self.uses_internal_info_input().into();
-        dpe.contexts[parent_idx].uses_internal_input_dice = self.uses_internal_dice_input().into();
-
         let target_locality = if !self.changes_locality() {
             locality
         } else {
@@ -167,6 +164,9 @@ impl CommandExecution for DeriveChildCmd {
         // Add child to the parent's list of children.
         dpe.contexts[parent_idx].add_child(child_idx)?;
 
+        dpe.contexts[child_idx].uses_internal_input_info = self.uses_internal_info_input().into();
+        dpe.contexts[child_idx].uses_internal_input_dice = self.uses_internal_dice_input().into();
+
         Ok(Response::DeriveChild(DeriveChildResp {
             handle: child_handle,
             parent_handle: dpe.contexts[parent_idx].handle,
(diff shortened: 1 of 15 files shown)


dpe/src/commands/derive_child.rs
```diff
@@ -373,6 +378,80 @@ mod tests {
...
+
+            EcdsaSig::from_private_components(
+                BigNum::from_slice(&resp.sig_r_or_hmac).unwrap(),
+                BigNum::from_slice(&resp.sig_s).unwrap(),
+            )
+            .unwrap()
+        };
+
+        // necessary so that generate_new_handle doesn't fail due to non-deterministic random bytes
+        dpe.contexts[1].handle = ContextHandle([0x2; ContextHandle::SIZE]);
```
jhand2, line 423: I don't understand what this is for. If you return the output handle from each of the responses, shouldn't it just work to use the handle that DPE generated?

sree-revoori1: When we call derive_child a second time without the make_default flag, it will call generate_new_handle which will generate another buffer of 1s (the first buffer of 1s was generated in the first call to derive_child and is located at contexts[1].) We can't have duplicate buffers, so I did this to make generate_new_handle work for the second derive_child call. For non-deterministic rand_bytes, this isn't an issue, and we wouldn't have had to do this. 

As for the test overall, I am trying to derive new contexts with contexts[0] as the parent, and then just pass the default context to certify key and sign to check that internal_input_info isn't getting included unecessarily. This is why using the output handle isn't that helpful.

jhand2: oh, because the RNG is fake. Maybe we should hash or increment the previous value every time for the test RNG. This will give us better testing properties, but will still be deterministic.

sree-revoori1: We already do increment the previous value. The issue is the buffer passed to generate_new_handle is always the buffer of 0s, so we still get a buffer of 1s. Even if we hashed it, we would get the same problem

jhand2: Actually maybe a better option: Use the rng crate and initialize it `from_seed` (https://docs.rs/rand/latest/rand/trait.SeedableRng.html#tymethod.from_seed). This will be deterministic but still give reasonable RNG properties if called more than once.

jhand2: For clarity, I mean call Rng::from_seed when creating OpensslCrypto and saving it as a member (since calling it in `rand_bytes` would lead to the same problem)

sree-revoori1: Done.

jhand2 (approved):

PR #203, opened 2023-09-25
Return error when attempting to retain default context

### LINKED ISSUE #197: DeriveChild Permits Context Handle to Coexist With the Default Context
NCC report page 35

sree-revoori1:
According to the spec, "DPE must not allow simultaneous use of a
default context and context handles within the same session: these
are mutually exclusive." If we retain the default context and
generate a new context within the same locality, we violate this.
This commit addresses this behavior and returns an InvalidArgument
error.

Further, we also slightly change the behavior of rotate_context
to align with the spec. If we try to rotate to a default context
and there are other random contexts within the same locality,
we return InvalidArgument since this is also a violation of the
aforementioned part of the spec, since we cannot have both default
and non-default contexts within the same locality.

Fixes #197


dpe/src/commands/derive_child.rs
```diff
@@ -147,6 +147,9 @@ impl CommandExecution for DeriveChildCmd {
             dpe.contexts[parent_idx].handle = ContextHandle([0xff; ContextHandle::SIZE]);
         } else if !dpe.contexts[parent_idx].handle.is_default() {
             dpe.contexts[parent_idx].handle = dpe.generate_new_handle(env)?;
+        } else if locality == target_locality {
+            // The default context cannot be retained.
```
jhand2, line 151: nit: I would recommend a little more detail here. Maybe something like

```
// The default context cannot be retained because a locality cannot
// support the default context and random context handles at the same time.
```

jhand2 (dismissed):

zhalvorsen:
It looks like this quote also requires us to update RotateContext. Instead of just checking if there is already default context for the locality. We will need to check if there are any valid contexts in the locality. If there are, it needs to fail otherwise it would once again allow simultaneous use of a default and non-default contexts.

dpe/src/commands/derive_child.rs
sree-revoori1: Done.

sree-revoori1:
@zhalvorsen Thanks for pointing this out. I have updated the PR with the rotate_context fix.

dpe/src/commands/rotate_context.rs
```diff
@@ -51,7 +69,9 @@ impl CommandExecution for RotateCtxCmd {
         if self.uses_target_is_default() {
             let default_context_idx =
                 dpe.get_active_context_pos(&ContextHandle::default(), locality);
-            if default_context_idx.is_ok() {
+            if default_context_idx.is_ok()
+                || self.non_default_valid_handles_exist(dpe, locality, idx)
+            {
```
zhalvorsen, line 74: If there is a default context there shouldn't be any other contexts in the locality,  so in this scenario if `default_context_idx` is valid and `idx` is valid, they must be the same index. So the request is to turn the default context into the default context. I'm leaning toward still considering this an error, but what do you think?

sree-revoori1: I agree especially since the spec says "If TARGET_IS_DEFAULT is set, NEW_CONTEXT_HANDLE SHALL be made the default in the caller's locality. If the default context handle is already defined in the caller's locality, this command SHALL fail.", but we are already doing this since default_context_idx.is_ok() would be true, right?

zhalvorsen: Okay cool LGTM!

dpe/src/commands/derive_child.rs
```diff
@@ -400,52 +406,68 @@ mod tests {
...
 
-        DeriveChildCmd {
+        let Response::RotateCtx(NewHandleResp { handle, .. }) = RotateCtxCmd {
             handle: ContextHandle::default(),
+            flags: RotateCtxFlags::empty(),
+        }
+        .execute(&mut dpe, &mut env, TEST_LOCALITIES[0])
+        .unwrap() else {
+            panic!("Rotate Context failed");
+        };
```
jhand2, line 420: nit (throughout this test): just calling `unwrap` with no else will give more information (like the error code). So I think that's prefferable.

dpe/src/commands/rotate_context.rs
```diff
@@ -33,6 +32,25 @@ impl RotateCtxCmd {
...
+    fn non_default_valid_handles_exist(
+        &self,
+        dpe: &mut DpeInstance,
+        locality: u32,
+        target_idx: usize,
+    ) -> bool {
+        dpe.contexts
+            .iter()
+            .enumerate()
+            .filter(|(idx, context)| {
```
jhand2, line 45: nit: You can use [any](https://doc.rust-lang.org/beta/core/iter/struct.Enumerate.html#method.any) instead of filter. Just a little more clear and stops after finding the first instance.

dpe/src/commands/rotate_context.rs
jhand2: Just reopening this comment until the PR is updated

dpe/src/commands/rotate_context.rs
```diff
@@ -22,9 +22,8 @@ bitflags! {
 #[derive(Debug, PartialEq, Eq, zerocopy::FromBytes)]
 #[cfg_attr(test, derive(zerocopy::AsBytes))]
 pub struct RotateCtxCmd {
-    handle: ContextHandle,
-    flags: RotateCtxFlags,
-    target_locality: u32,
```
jhand2, line 27: Why was `target_locality` removed?

sree-revoori1: Looks like we don't use it anywhere in rotate context. Also not in the spec

dpe/src/commands/derive_child.rs
sree-revoori1: Done

dpe/src/commands/rotate_context.rs
sree-revoori1: Done

dpe/src/commands/rotate_context.rs
jhand2: Never mind, I misread this.

dpe/src/commands/derive_child.rs
jhand2: Oh, ok. In that case, it might be good to do something like (this probably doesn't compile, but close enough)

```rust
match execute() {
  Ok(RotateContextResp(resp)) => something,
  Ok(_) => panic!("Invalid enum variant"),
  Err(e) => panic!("{}", e),
}
```

sree-revoori1: Done

dpe/src/commands/rotate_context.rs
jhand2: Hmm. The upstream spec supports it, but we could disallow it in the profile. I think it's preferred from a security perspective if handles can't be moved between localities.

So fine to leave this as-is. We just need to update the profile spec to say that the `target_locality` parameter is not supported and that handles will always be rotated into the same locality.

jhand2 (approved):

PR #208, opened 2023-10-03
Add functionality to Platform trait to print from DPE

### LINKED ISSUE #205: It is not currently possible to log to UART from DPE code
Adding the ability to print things from caliptra-dpe code would help a lot for debugging. We should provide a way for integrations to implement an API which will log to UART.

The main thing that makes this challenging is that we'll want to be able to use format strings. The most straightforward way to do this would be to implement a macro similar to https://github.com/chipsalliance/caliptra-sw/blob/850d604ef0b7485d6fdc803d64003e3a1b1222e4/drivers/src/printer.rs#L48-L52.

But I don't know how to do this while letting the integration specify the macro implementation (similar to implementing a trait, but for macros).

sree-revoori1:
Make use of the ufmt crate to create a macro plat_println which prints things from the platform. This will be useful for debugging and allows the implementation to define the printing functionality. In the default implementation, we simply use print!.

Fixes #205

### DIFF
--- a/platform/Cargo.toml
+++ b/platform/Cargo.toml
@@ -13,3 +13,4 @@ dpe_profile_p384_sha384 = []
 
 [dependencies]
 openssl = {version = "0.10", optional = true}
+ufmt = { git = "https://github.com/korran/ufmt.git", rev = "1d0743c1ffffc68bc05ca8eeb81c166192863f33", features = ["inline"] }
--- a/platform/src/lib.rs
+++ b/platform/src/lib.rs
@@ -11,6 +11,8 @@ pub use openssl::x509::X509;
 #[cfg(feature = "openssl")]
 pub mod default;
 
+pub mod printer;
+
 pub const MAX_CHUNK_SIZE: usize = 2048;
 
 #[derive(Debug)]
@@ -18,6 +20,7 @@ pub enum PlatformError {
     CertificateChainError,
     NotImplemented,
     IssuerNameError,
+    PrintError,
 }
 
 pub trait Platform {
@@ -47,4 +50,6 @@ pub trait Platform {
     fn get_vendor_sku(&mut self) -> Result<u32, PlatformError>;
 
     fn get_auto_init_locality(&mut self) -> Result<u32, PlatformError>;
+
+    fn write_str(&mut self, str: &str) -> Result<(), PlatformError>;
 }
--- a/platform/src/printer.rs
+++ b/platform/src/printer.rs
@@ -1,0 +1,30 @@
+// Licensed under the Apache-2.0 license
+
+use ufmt::uWrite;
+
+use crate::{Platform, PlatformError};
+
+pub struct Printer<'a> {
+    pub platform: &'a mut dyn Platform,
+}
+
+impl<'a> uWrite for Printer<'a> {
+    type Error = PlatformError;
+
+    fn write_str(&mut self, str: &str) -> Result<(), Self::Error> {
+        self.platform.write_str(str)
+    }
+}
+
+impl<'a> Printer<'a> {
+    pub fn new(platform: &'a mut dyn Platform) -> Self {
+        Self { platform }
+    }
+}
+
+#[macro_export]
+macro_rules! plat_println {
+    ($platform:expr, $($tt:tt)*) => {{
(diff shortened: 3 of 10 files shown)


dpe/src/printer.rs
```diff
@@ -0,0 +1,33 @@
...
+pub struct Printer<'a, T: DpeTypes + 'a> {
+    pub platform: T::Platform<'a>,
+}
+
+impl<'a, T: DpeTypes> uWrite for Printer<'a, T> {
+    type Error = PlatformError;
+
+    fn write_str(&mut self, str: &str) -> Result<(), Self::Error> {
+        self.platform.write_str(str)?;
+        Ok(())
```
zhalvorsen, line 18: This can be simplified

```suggestion
        self.platform.write_str(str)
```

dpe/src/printer.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+impl<'a, T: DpeTypes> Printer<'a, T> {
+    pub fn new(platform: T::Platform<'a>) -> Self {
```
zhalvorsen, line 23: fyi: you might need to add this

```suggestion
    #[allow(dead_code)]
    pub fn new(platform: T::Platform<'a>) -> Self {
```

dpe/src/printer.rs
sree-revoori1: Done.

dpe/src/printer.rs
sree-revoori1: Changed to _new instead.

zhalvorsen: That can be a little awkward when we do use it though. Or do we always expect to use `plat_println`?

sree-revoori1: I think we will always use plat_println! but when we do use plat_println! for the first time, _new will no longer be considered unused by the compiler and we can change it back to new

zhalvorsen: I guess I was picturing it as more of a debugging capability than a regular addition. How often do we expect to commit code with `plat_println!`s? I'm good with the `_new` in that case then.

sree-revoori1: I think you're right, we don't really expect to use it in production code. Changed to #[allow(dead_code)].

zhalvorsen: Sorry to bring this back, but if we make the printer module public, because this is a public function it doesn't need #[allow(dead_code)]

platform/src/printer.rs
```diff
@@ -0,0 +1,31 @@
...
+impl<'a> Printer<'a> {
+    #[allow(dead_code)]
+    pub fn new(platform: &'a mut dyn Platform) -> Self {
+        Self { platform }
+    }
+}
+
+#[macro_export]
+macro_rules! plat_println {
+    ($platform:ident, $($tt:tt)*) => {{
```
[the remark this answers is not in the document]
sree-revoori1: Nice, thanks for finding this. Done

platform/src/lib.rs
```diff
@@ -11,13 +11,16 @@ pub use openssl::x509::X509;
 #[cfg(feature = "openssl")]
 pub mod default;
 
+mod printer;
```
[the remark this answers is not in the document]
sree-revoori1: Done

dpe/src/printer.rs
sree-revoori1: Done

zhalvorsen (dismissed):

platform/Cargo.toml
```diff
@@ -12,4 +12,5 @@ dpe_profile_p256_sha256 = []
 dpe_profile_p384_sha384 = []
 
 [dependencies]
-openssl = {version = "0.10", optional = true}
\ No newline at end of file
+openssl = {version = "0.10", optional = true}
+ufmt = { git = "https://github.com/korran/ufmt.git", rev = "1d0743c1ffffc68bc05ca8eeb81c166192863f33", features = ["inline"] }
```
zhalvorsen, line 16: fyi: I also had to add `ufmt` to dpe/Cargo.toml

sree-revoori1: Done

jhand2 (approved):

PR #209, opened 2023-10-03
DPE command validation and DeriveChild changes

hpya93:
This contains-

_•Test all failure cases defined in the DPE iRoT Profile_

- Added a failure check in the initialize context and getcertificatechain command. Other fail checks are already present in the code which I have tested.

_•For all input enums/flags for all DPE commands, ensure there is a test case for each permutation._

_•Add tests to ensure localities are properly isolated from each other for all commands_

- I have done testing on separate localities in simulator side (TestInitializeContext using multiple localities)

- The code for separating localities is already there in the command file. Just checked that the initialize context is called for each locality.

- In emulator side, we are not setting any locality to the DPE instance when executing commands.

_•Validate proper behavior to against the feature support matrix (exposed by GetProfile)_

- Validated the getprofile modes for both simulator and emulator using code from this 

_•DeriveChild  - Test limits on number of DeriveChild calls_
- Have checked whether proper error is reported when number of derived child exceed MAX TCI nodes

_•Test that DeriveChild privileges (like ALLOW_CA) are correctly bounded and that a DPE context cannot escalate its privileges_
 - Multi Tcb validation for matching TCITYPE passed to Derive child request   
 - Have checked whether the proper error is returned when feature flags are set in derived child context but not supported by parent context  

**Query:**

In fail cases of getcertificatechain command, spec says 'If OFFSET is larger than the full certificate chain, this API shall return DPE_STATUS_INVALID_ARGUMENT'

But,

The full cert chain size is computed in runtime, by repeated calls to GetCertificateChain()

So, we can't check if offset > fullcertchainsize

This can be checked only if we pass a negative test input like

cases := []GetCertificateReq{
{size: 2048, offset: 100000},
{size: 2048, offset: 100000}, // Negative case.
}

So, should we write a negative test case like this?

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -55,6 +55,16 @@ impl CryptoError {
         // field, so we can read the discriminant without offsetting the pointer.
         unsafe { *<*const _>::from(self).cast::<u16>() }
     }
+
+    pub fn get_error_detail(&self) -> Option<u32> {
+        match self {
+            CryptoError::AbstractionLayer(code) => Some(*code),
+            CryptoError::CryptoLibError(code) => Some(*code),
+            CryptoError::Size => None,
+            CryptoError::NotImplemented => None,
+            CryptoError::HashError(code) => Some(*code),
+        }
+    }
 }
 
 pub trait Hasher: Sized {
--- a/dpe/src/response.rs
+++ b/dpe/src/response.rs
@@ -206,19 +206,8 @@ impl DpeErrorCode {
     /// Reporting of detailed error information is platform-defined.
     pub fn get_error_detail(&self) -> Option<u32> {
         match self {
-            DpeErrorCode::Platform(e) => match e {
-                PlatformError::CertificateChainError => None,
-                PlatformError::NotImplemented => None,
-                PlatformError::IssuerNameError(code) => Some(*code),
-                PlatformError::PrintError(code) => Some(*code),
-            },
-            DpeErrorCode::Crypto(e) => match e {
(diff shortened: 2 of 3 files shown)


verification/client.go
```diff
@@ -236,6 +239,18 @@ func (c *Client[_, _]) GetCertificateChain() (*GetCertificateChainResp, error) {
 	return &certs, nil
 }
 
+// DeriveChild() calls the DPE CertifyKey command.
```
sree-revoori1, line 242: *calls the DPE DeriveChild command

verification/deriveChild_test.go
```diff
@@ -0,0 +1,347 @@
...
+			log.Fatal(err)
+		}
+		defer d.PowerOff()
+	}
+	client, err := NewClient256(d)
+	if err != nil {
+		t.Fatalf("Could not initialize client: %v", err)
+	}
+
+	tciNodeCount := uint32(1)
```
sree-revoori1, line 37: can you add a comment here that the initial tciNodeCount is 1 due to auto-init?

verification/client.go
hpya93: Hi @sree-revoori1  , seems the code flow is modified few days back and this tree needs significant changes to incorporate the new code. I shall commit the changes after rebasing the code. 

Thank you!

hpya93:
Have incorporated latest pattern changes in PR

hpya93:
@sree-revoori1 Can you please review and let us know the comments on this PR?

PR #216, opened 2023-10-06
Implement CertifyKey with CSRs and add encodings of CMS/CSRs

### LINKED ISSUE #83: Implement DPE Command CertifyCsr
See https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.md#certifycsr for specification.

### LINKED ISSUE #237: build.rs is not conditionally compiling correctly
Since build.rs runs before compilation, it doesn't know about features. Thus, The default value of dpe_profile_p256_sha256 is always getting used. This causes issues with the openssl implementation of ecdsa_sign_with_alias since we end up always signing with a 32 bit private key.

sree-revoori1:
Also, rework build.rs to write both the 256 bit and 384 bit private keys. This is necessary because feature compilation doesn't work in build.rs. Then when we read the alias key in OpensslCrypto::ecdsa_sign_with_alias, we can read from the proper file based on the value of the algorithm size. 

fixes #237 
fixes #83

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -5,7 +5,7 @@ use crate::{
     dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
     response::{CertifyKeyResp, DpeErrorCode, Response, ResponseHdr},
     tci::TciNodeData,
-    x509::{DirectoryString, MeasurementData, Name, X509CertWriter},
+    x509::{CertWriter, DirectoryString, MeasurementData, Name},
     DPE_PROFILE, MAX_CERT_SIZE, MAX_HANDLES,
 };
 use bitflags::bitflags;
@@ -77,7 +77,7 @@ impl CommandExecution for CertifyKeyCmd {
             .crypto
             .derive_cdi(DPE_PROFILE.alg_len(), &digest, b"DPE")
             .map_err(|_| DpeErrorCode::CryptoError)?;
-        let (_, pub_key) = env
+        let (priv_key, pub_key) = env
             .crypto
             .derive_key_pair(algs, &cdi, &self.label, b"ECC")
             .map_err(|_| DpeErrorCode::CryptoError)?;
@@ -119,7 +119,7 @@ impl CommandExecution for CertifyKeyCmd {
         let cert_size = match self.format {
             Self::FORMAT_X509 => {
                 let mut tbs_buffer = [0u8; MAX_CERT_SIZE];
-                let mut tbs_writer = X509CertWriter::new(&mut tbs_buffer, true);
+                let mut tbs_writer = CertWriter::new(&mut tbs_buffer, true);
                 if issuer_len > MAX_CHUNK_SIZE {
                     return Err(DpeErrorCode::InternalError);
                 }
@@ -144,7 +144,7 @@ impl CommandExecution for CertifyKeyCmd {
                     .ecdsa_sign_with_alias(DPE_PROFILE.alg_len(), &tbs_digest)
                     .map_err(|_| DpeErrorCode::CryptoError)?;
 
-                let mut cert_writer = X509CertWriter::new(&mut cert, true);
+                let mut cert_writer = CertWriter::new(&mut cert, true);
                 bytes_written =
                     cert_writer.encode_ecdsa_certificate(&tbs_buffer[..bytes_written], &sig)?;
                 u32::try_from(bytes_written).map_err(|_| DpeErrorCode::InternalError)?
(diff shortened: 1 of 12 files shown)


dpe/src/x509.rs
```diff
@@ -1030,6 +1160,149 @@ impl X509CertWriter<'_> {
...
 
+    /// Encode a SignedData
+    ///
+    /// SignedData ::= SEQUENCE {
+    /// version CMSVersion,
+    /// digestAlgorithms DigestAlgorithmIdentifiers,
+    /// encapContentInfo EncapsulatedContentInfo,
+    /// certificates [0] IMPLICIT CertificateSet OPTIONAL,
+    /// crls [1] IMPLICIT RevocationInfoChoices OPTIONAL,
+    /// signerInfos SignerInfos }
```
jhand2, line 1171: minor nit: Can you indent these fields?

dpe/src/x509.rs
```diff
@@ -1030,6 +1160,149 @@ impl X509CertWriter<'_> {
         Ok(bytes_written)
     }
 
+    /// Encode a SignedData
```
jhand2, line 1163: nit: Might be good to add a comment that we don't populate the `certificates` or `crls` fields

dpe/src/x509.rs
```diff
@@ -1030,6 +1160,149 @@ impl X509CertWriter<'_> {
...
+
+        // signerInfos
+        bytes_written += self.encode_signer_infos(serial_number, issuer_name, sig)?;
+
+        Ok(bytes_written)
+    }
+
+    /// Encode a SET OF SignerInfo
+    ///
+    /// SignerInfo ::= SEQUENCE {
```
jhand2, line 1211: nit: same comment about indentation

dpe/src/commands/certify_key.rs
```diff
@@ -150,7 +150,58 @@ impl CommandExecution for CertifyKeyCmd {
...
+                    .crypto
+                    .hash(
+                        DPE_PROFILE.alg_len(),
+                        &cert_req_info_buffer[..bytes_written],
+                    )
+                    .map_err(|_| DpeErrorCode::HashError)?;
+                let cert_req_info_sig = env
+                    .crypto
+                    .ecdsa_sign_with_alias(DPE_PROFILE.alg_len(), &cert_req_info_digest)
+                    .map_err(|_| DpeErrorCode::CryptoError)?;
```
jhand2, line 178: PKCS#10 CSRs should be self-signed (signed the the private key returned by `derive_key_pair`). This proves ownership of the key (this is also why we wrap the CSR in a CMS structure. So we can sign the whole thing with a different key).

dpe/src/x509.rs
```diff
@@ -1135,6 +1408,118 @@ impl X509CertWriter<'_> {
...
+
+        Ok(bytes_written)
+    }
+
+    /// Encode a CMS ContentInfo message
+    ///
+    /// ContentInfo ::= SEQUENCE {
+    /// contentType ContentType,
+    /// content [0] EXPLICIT ANY DEFINED BY contentType }
+    pub fn encode_cms(
```
jhand2, line 1497: Can you add some unit tests to make sure the certs can be parsed?

The certification_request module from x509_parser should work for the PKCS#10 stuff: https://docs.rs/x509-parser/latest/x509_parser/certification_request/index.html

And then the CMS crate you added for parsing the outer CMS part.

dpe/src/x509.rs
```diff
@@ -1135,6 +1408,118 @@ impl X509CertWriter<'_> {
...
+        bytes_written += self.encode_version()?;
+
+        // subject
+        bytes_written += self.encode_rdn(subject_name)?;
+
+        // subjectPublicKeyInfo
+        bytes_written += self.encode_ecdsa_subject_pubkey_info(pub_key)?;
+
+        // extensions
+        bytes_written += self.encode_extensions(measurements)?;
```
jhand2, line 1454: I think there needs to be a wrapper struct around the extensions. There is an attribute "Extension Request" defined in RFC 2985 that should wrap these extensions.

dpe/src/commands/certify_key.rs
sree-revoori1: Done

dpe/src/x509.rs
sree-revoori1: Done

dpe/src/x509.rs
sree-revoori1: Done

dpe/src/x509.rs
sree-revoori1: Done

dpe/src/x509.rs
sree-revoori1: Done

dpe/src/x509.rs
sree-revoori1: Done

dpe/src/commands/certify_key.rs
```diff
@@ -329,4 +402,214 @@ mod tests {
...
+        };
+        env.crypto
+            .get_pubkey_serial(DPE_PROFILE.alg_len(), &pub_key, &mut subj_serial)
+            .unwrap();
+        match sid {
+            SignerIdentifier::IssuerAndSerialNumber(issuer_and_serial_number) => {
+                let cert_serial_number = &issuer_and_serial_number.serial_number;
+                assert_eq!(&subj_serial[..20], cert_serial_number.as_bytes());
+                // note: the cms crate does not let us convert issuer_and_serial_number.issuer
+                // into bytes so we unfortunately cannot test the issuer name here.
```
jhand2, line 506: I don't quite understand. Isn't there an as_bytes function here? https://docs.rs/x509-cert/0.2.3/x509_cert/serial_number/struct.SerialNumber.html#method.as_bytes

dpe/src/commands/certify_key.rs
```diff
@@ -329,4 +402,214 @@ mod tests {
...
+        let pub_key_der = ec_point.data();
+        // skip first 0x04 der encoding byte
+        let x = BigNum::from_slice(&pub_key_der[1..DPE_PROFILE.get_ecc_int_size() + 1]).unwrap();
+        let y = BigNum::from_slice(&pub_key_der[DPE_PROFILE.get_ecc_int_size() + 1..]).unwrap();
+        EcKey::from_public_key_affine_coordinates(group, &x, &y).unwrap();
+
+        // note: ideally here, we would verify the signature with the public key.
+        // However, the x509_parser library does not expose a function to
+        // convert the certification request info into bytes so we have no way to
+        // get the digest.
```
jhand2, line 565: I think you can use the `raw` field here: https://docs.rs/x509-parser/latest/x509_parser/certification_request/struct.X509CertificationRequestInfo.html

dpe/src/x509.rs
```diff
@@ -1088,6 +1299,251 @@ impl X509CertWriter<'_> {
...
+    ///
+    /// Attributes ::= SET OF Attribute
+    ///
+    /// Attribute ::= SEQUENCE {
+    ///    attrType OBJECT IDENTIFIER,
+    ///    attrValues SET OF AttributeValue
+    /// }
+    ///
+    /// AttributeValue ::= ANY -- Defined by attribute type
+    #[allow(clippy::identity_op)]
```
jhand2, line 1401: Why is this needed? Because of the `| 0x0`?

dpe/src/x509.rs
```diff
@@ -1088,6 +1299,251 @@ impl X509CertWriter<'_> {
         Ok(bytes_written)
     }
 
+    fn encode_implicit_extensions(
```
jhand2, line 1302: I wonder if we can avoid duplicating the code with a parameter here. Something like

```rust
enum ConstructedTag {
    Implicit(u8),
    Explicit(u8),
}

fn encode_implicit_extensions(
    &mut self,
    measurements: &MeasurementData,
    constructed: Option<ConstructedTag>,
    match constructed {
        Some(Implicit(tag_num)) => // Do implicit tag
        Some(Explicit(tag_num)) => // Do explicit tag
        None => // Use normal tag
    }
}
```

dpe/src/x509.rs
sree-revoori1: Yes

dpe/src/commands/certify_key.rs
sree-revoori1: I am using the cms crate to parse the wrapping ContentInfo  - https://docs.rs/cms/0.2.2/cms/cert/struct.IssuerAndSerialNumber.html

sree-revoori1: i was able to use Name::to_der to parse it actually.

dpe/src/commands/certify_key.rs
sree-revoori1: Thanks, this worked.

dpe/src/x509.rs
sree-revoori1: Done, except I just passed a bool to encode_extensions since there's only two cases.

jhand2 (approved):

PR #220, opened 2023-10-09
Add a wrapper around the generics-based DPE clients

jhand2:
To allow DPE tests to be less opinionated about which DPE profile the target uses, add a wrapper around the DPE Client (changed to DPEABIClient).

This leaves the ABI client in place for callers who do not want a more permissive interface.

### DIFF
--- a/verification/abi.go
+++ b/verification/abi.go
@@ -2,6 +2,12 @@
 
 package verification
 
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
 const (
 	CmdMagic  uint32 = 0x44504543
 	RespMagic uint32 = 0x44504552
@@ -11,6 +17,20 @@ const (
 )
 
 type CommandCode uint32
+
+type Support struct {
+	Simulation    bool
+	ExtendTci     bool
+	AutoInit      bool
+	Tagging       bool
+	RotateContext bool
+	X509          bool
+	Csr           bool
+	IsSymmetric   bool
+	InternalInfo  bool
+	InternalDice  bool
+	IsCA          bool
+}
 
 const (
 	CommandGetProfile          CommandCode = 0x1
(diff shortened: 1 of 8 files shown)


verification/abi.go
```diff
@@ -126,3 +154,370 @@ type GetTaggedTCIResp[Digest DigestAlgorithm] struct {
...
+// DPEABI384 is a client that implements DPE_PROFILE_IROT_P384_SHA384
+type DPEABI384 = DPEABI[NISTP384Parameter, SHA384Digest]
+
+// dpeProfileImplementsTypeConstraints checks that the requested DPEABI type constraints are compatible with the DPE profile.
+func dpeProfileImplementsTypeConstraints[C Curve, D DigestAlgorithm](profile Profile) error {
+	// Test that the expected value types produced by each DPE profile can be assigned to variables of type C and D
+	var c C
+	var d D
+	switch profile {
+	case ProfileP256SHA256:
```
chrisfenner, line 182: nit/highly optional: just thinking about the error messages here. If you derive the expected profile based on C and D, you can compare that derived profile with the actual one, returning either:

* "Client requested (Curve C, Digest D), which is not a valid DPE profile"
* "Client requested DPE Profile X, but DPE implements Y"

verification/abi.go
```diff
@@ -77,6 +101,10 @@ type GetProfileResp struct {
 
 type CertifyKeyFlags uint32
 
+const (
+	CertifyAddIsCA CertifyKeyFlags = 1 << 30
```
zhalvorsen, line 105: Is the `Add` part here a typo? I'm thinking this should either be `CertifyKeyIsCA` or simply `IsCA`.

verification/abi.go
```diff
@@ -126,3 +154,370 @@ type GetTaggedTCIResp[Digest DigestAlgorithm] struct {
...
+		if _, ok := any(d).(SHA384Digest); !ok {
+			return fmt.Errorf("an incorrect digest type was passed to a DPE implementing DPE_PROFILE_IROT_P384_SHA384")
+		}
+		return nil
+	}
+	return fmt.Errorf("unsupported DPE profile: %v", profile)
+}
+
+// NewDPEABI initializes a new DPE client.
+func NewDPEABI[C Curve, D DigestAlgorithm](t Transport) (*DPEABI[C, D], error) {
```
chrisfenner, line 205: optional: something to think about (but not for too long, this is really an optional comment): do you expect callers to actually call `NewDPEABI` or `NewDPEABI256` or `NewDPEABI384`, or do you expect them to call `NewClient` and then perform a type assertion that they got e.g., a `DPEABI256`? (Maybe you expect some will do the first thing, some will do the second thing).

Depending on how opinionated you feel about what callers should do, you can hide some of these symbols:

- the actual parameterized DPEABI type, renaming to dpeABI (leaving the aliases `DPEABI256` and `DPEABI384` exported for callers who want to write code for *their* expected profile)
- the `NewDPEABI*` factories (requiring the caller to ask the client library which DPE ABI it's got)

In general if this library gets some (non-external, non-test) callers, it'd be easier to make private things public rather than the other way around, so I'd suggest making any of this stuff private that's not needed to write the tests.

chrisfenner (approved):
LGTM I just have minor, hopefully helpful, comments

verification/abi.go
jhand2: Technically the flag in the ABI is `ADD_IS_CA` which is short for "Additional Input: IS_CA". This is because the mainline DPE spec defines a parameter "Additional Input", which is additional data used to populate the cert. And then it is up to the profile to decide what additional input is supported.

But this is probably minutiae the caller doesn't care about, so I'll remove the add from here.

verification/abi.go
jhand2: I could see a caller which really does want a 256 client and wants to fail if that's not what the RoT implements. So I'll leave the NewDPEABI* factories. But I changed the type-parameterized DPEABI to be private.

verification/abi.go
zhalvorsen: Oh, I missed that in the ABI. You can leave it as is.

verification/abi.go
jhand2: Done.

zhalvorsen (dismissed):

zhalvorsen (approved):

PR #227, opened 2023-10-10
Minor fixes to make DPE tests easier to run against Caliptra

jhand2:
1. Make InitializeContext tests use simulation when the test asks for it instead of when the target supports it.
2. TagTCI assumes the default context TCI is all zeros, but this only works if the target does not support auto initialization. Disable this assertion until the test can use DeriveChild to specify the input TCI.
3. For tagets whose support vector isn't configurable, the GetProfile tests were just running the same test repeatedly. Update this to just run the test once. For simulator configuration, this testing belongs in simulator unit tests.

### DIFF
--- a/verification/verification.go
+++ b/verification/verification.go
@@ -19,6 +19,9 @@ var TestCases = []TestCase{
 	TestCase{
 		"InitializeContext", TestInitializeContext, []string{},
 	},
+	TestCase{
+		"InitializeContextSimulation", TestInitializeSimulation, []string{"Simulation"},
+	},
 	// CertifyKey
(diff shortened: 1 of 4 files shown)


verification/verification.go
```diff
@@ -38,46 +41,4 @@ var TestCases = []TestCase{
 	TestCase{
 		"GetProfile", TestGetProfile, []string{},
 	},
```
zhalvorsen, line 43: optional: Should we put `GetProfile` first because almost all of the other tests depend on it working correctly?

zhalvorsen (approved):

PR #232, opened 2023-10-11
Remove child from parent context correctly in destroy_context

sree-revoori1:
Previously, we were always setting the children of a parent context to 0 in destroy_context, which is incorrect: If P is the parent of A and B and we destroy_context on A, then B would cease to be the child of P.

### DIFF
--- a/dpe/src/commands/destroy_context.rs
+++ b/dpe/src/commands/destroy_context.rs
@@ -2,9 +2,8 @@
 use super::CommandExecution;
 use crate::{
     context::ContextHandle,
-    dpe_instance::{flags_iter, DpeEnv, DpeInstance, DpeTypes},
+    dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
     response::{DpeErrorCode, Response, ResponseHdr},
-    MAX_HANDLES,
 };
 
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


dpe/src/commands/destroy_context.rs
```diff
@@ -57,8 +57,8 @@ impl CommandExecution for DestroyCtxCmd {
             .take(contexts_to_destroy.len())
         {
             if *context_to_destroy {
-                dpe.contexts[idx].destroy();
                 let parent = dpe.contexts[idx].parent_idx as usize;
+                dpe.contexts[idx].destroy();
```
jhand2, line 61: nit: in Context::destroy should we set parent_idx to 0xFF to protect against this code being moved around incorrectly in the future?

dpe/src/commands/destroy_context.rs
```diff
@@ -33,7 +33,7 @@ impl CommandExecution for DestroyCtxCmd {
         // contexts_to_destroy[i] == true implies dpe.contexts[i] must be destroyed
         let mut contexts_to_destroy = [false; MAX_HANDLES];
         // updated_children[i] holds i's non-destroyed children.
-        let mut updated_children = [0u32; MAX_HANDLES];
+        let mut updated_children = [u32::MAX; MAX_HANDLES];
```
jhand2, line 36: I think this code works correctly, but is a bit hard to follow, which could lead to issues being introduced in the future. I'll think about this a little more, but maybe we could try to refactor this to simplify it.

dpe/src/commands/destroy_context.rs
jhand2, line 61: Instead of `dpe.contexts[idx]` can we use `context_to_destroy`?

dpe/src/commands/destroy_context.rs
jhand2: Here's a straw proposal for how this could maybe be changed:

```rust
// TODO: Is there a way to be extra sure this is being copied and not moved?
// Like calling a .copy function?
let mut tmp_contexts = dpe.contexts;

let to_destroy = (1 << idx) | dpe.get_descendants(context)?;

for idx, c in tmp_contexts {
    // Clears all the to_destroy bits in the children of every context
    c.children &= ~to_destroy;
    if to_destroy & (1 << idx) != 0 {
        c.destroy();
    }
}

dpe.contexts = tmp_contexts;
```

I _think_ this does the same thing as the existing code.

Looking at this, I'm not even sure we need to do the copying because I don't think any of these calls can fail.

dpe/src/commands/destroy_context.rs
sree-revoori1: Done

sree-revoori1: context_to_destroy is just a bool, so how would we use it?

jhand2: Oh, never mind, I misread and thought it was the actual context.

dpe/src/commands/destroy_context.rs
sree-revoori1: Thanks, this is way simpler. Also removed the copying.

jhand2 (approved):

PR #235, opened 2023-10-11
Refactor logic for `DeriveChild`.

zhalvorsen:
There are quite a few edge cases in `DeriveChild` that need to be considered. This mainly hitches on if a locality has a default context, it MUST be the only context in the locality. Which also means, default and non-default contexts can never be together. This refactors this logic to be more explicit about when a `DeriveChild` command will be able to satisfy this requirement.

### DIFF
--- a/dpe/src/commands/derive_child.rs
+++ b/dpe/src/commands/derive_child.rs
@@ -64,27 +64,91 @@ impl DeriveChildCmd {
         self.flags.contains(DeriveChildFlags::INPUT_ALLOW_X509)
     }
 
-    /// Check if this will result in two default contexts in the same locality.
+    /// Whether it is okay to make a default context.
     ///
-    /// There can only be one default context in each locality at any given time. A default context
-    /// has a specific handle. If there are multiple, the DPE instance will not be able to
-    /// differentiate between them.
+    /// When a default context is in a locality, it MUST be the only context in the locality. This
+    /// checks that the operation will result in the locality will only have the newly created
+    /// default context and no others.
     ///
     /// # Arguments
     ///
     /// * `parent_idx` - Index of the soon-to-be parent.
-    /// * `default_context_idx` - Index of the current default context, if there is one.
+    /// * `default_context_idx` - Index of the target locality's default context, if there is one.
+    /// * `num_contexts_in_locality` - Number of contexts already in the locality.
     fn safe_to_make_default(
         &self,
         parent_idx: usize,
-        default_context_idx: Result<usize, DpeErrorCode>,
+        default_context_idx: Option<usize>,
+        num_contexts_in_locality: usize,
     ) -> bool {
-        if let Ok(default_idx) = default_context_idx {
-            if default_idx != parent_idx || self.retains_parent() {
-                return false;
+        match (num_contexts_in_locality, default_context_idx) {
+            // No other contexts in the locality.
+            (0, None) => true,
+            // There is only one context, but that context is the parent.
+            (1, Some(default_idx)) if default_idx == parent_idx => {
+                // It is okay if the parent is about to be retired. The Child can be the default
+                // because the parent is the only other context in the locality and it is about to
+                // be retired.
+                !self.retains_parent()
             }
+            // In all other scenarios, there will be a combination of default and non-default
+            // contexts
+            _ => false,
         }
-        true
+    }
+
+    /// Whether it is okay to make a NON-default context.
+    ///
+    /// There can never be a mixture of default and non-default contexts within the same locality.
+    /// This checks to make sure the operation will not result in a mixture.
+    ///
+    /// # Arguments
+    ///
+    /// * `parent_idx` - Index of the soon-to-be parent.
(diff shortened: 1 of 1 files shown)


dpe/src/commands/derive_child.rs
```diff
@@ -612,4 +673,37 @@ mod tests {
             Err(DpeErrorCode::InvalidArgument)
         );
     }
+
+    #[test]
+    fn test_make_default_in_other_locality_that_has_non_default() {
```
zhalvorsen, line 678: This tests the scenario called out by NCC

dpe/src/commands/derive_child.rs
```diff
@@ -64,27 +64,86 @@ impl DeriveChildCmd {
...
+    /// When a default context is in a locality, it MUST be the only context in the locality. There
+    /// can never be a mixture of default and non-default contexts. This checks that the operation
+    /// will not violate either statement.
+    ///
+    /// # Arguments
+    ///
+    /// * `parent_idx` - Index of the soon-to-be parent.
+    /// * `default_context_idx` - Index of the current default context, if there is one.
+    /// * `num_contexts_in_locality` - Number of contexts already in the locality.
+    fn safe_to_make_child(
```
jhand2, line 131: nit: There is another piece of code below that is also relevant to whether it's safe to create the child:

```rust
        // [jhand editorial] This statement is only hit if parent_idx handle is default
        } else if locality == target_locality {
            // The default context cannot be retained because a locality cannot
            // support the default context and random context handles at the same time.
            return Err(DpeErrorCode::InvalidArgument);
        }
```

I'm wondering if it would make sense to move that here so that all of these cases are handled in one place. But it is a little different (it's about whether it's safe to retain the parent). So maybe it's ok to handle it separately.

jhand2 (dismissed):

dpe/src/commands/derive_child.rs
zhalvorsen: I was picturing that the parent context always rotated, even if the parent was a default. Fixed logic and removed the section mentioned.

dpe/src/commands/derive_child.rs
```diff
@@ -64,27 +64,89 @@ impl DeriveChildCmd {
...
-        if let Ok(default_idx) = default_context_idx {
-            if default_idx != parent_idx || self.retains_parent() {
-                return false;
+        match (num_contexts_in_locality, default_context_idx) {
+            // No other contexts in the locality.
+            (0, None) => true,
+            // There is only one context, but that context is the parent.
+            (1, Some(default_idx)) if default_idx == parent_idx => {
+                // It is okay if the parent is about to be retired.
+                !self.retains_parent()
```
jhand2, line 90: I think whether or not this is ok still depends on whether the parent and child contexts are in the same locality, right?

zhalvorsen: Yeah that's true, default_context_idx is the index of the default context in the target locality so it should be covered. Or am I missing something?

dpe/src/commands/derive_child.rs
```diff
@@ -64,27 +64,89 @@ impl DeriveChildCmd {
...
+    /// Whether it is okay to create a child in the given environment.
+    ///
+    /// When a default context is in a locality, it MUST be the only context in the locality. There
+    /// can never be a mixture of default and non-default contexts. This checks that the operation
+    /// will not violate either statement.
+    ///
+    /// # Arguments
+    ///
+    /// * `parent_idx` - Index of the soon-to-be parent.
+    /// * `default_context_idx` - Index of the current default context, if there is one.
```
sree-revoori1, line 132: I think the argument descriptions need to be updated here

dpe/src/commands/derive_child.rs
```diff
@@ -64,27 +64,89 @@ impl DeriveChildCmd {
...
+            // In all other scenarios, there will be a combination of default and non-default
+            // contexts
+            _ => false,
         }
-        true
+    }
+
+    /// Whether it is okay to make a NON-default context.
+    ///
+    /// There can never be a mixture of default and non-default contexts. This checks to make sure
```
sree-revoori1, line 100: nit: Maybe specify that there can never be a mix of default/non-default contexts within the same locality

sree-revoori1:
nit: could you run the test_panic_missing test within the caliptra runtime? I highly doubt this change introduces panic, but just want to be sure before merging.

dpe/src/commands/derive_child.rs
zhalvorsen: Done

dpe/src/commands/derive_child.rs
zhalvorsen: Done

dpe/src/commands/derive_child.rs
zhalvorsen: Done

zhalvorsen:
Passing `test_panic_missing` tests from caliptra-sw

sree-revoori1 (approved):

Issue #226, opened 2023-10-10
[verification] Reboot target less in tests


jhand2:
Every DPE verification test reboots the target. When run on a real server, this can take a very long time. We should refactor tests so they don't need to be rebooted so often. Ideally, most tests could be run without a reboot.

One way to make this happen is to hold onto the root DPE context node and then delete it after every test.

sree-revoori1:
Fixed by #241

PR #241, opened 2023-10-12
Reboot the verification tests less often.

zhalvorsen:
With a simulator it is very easy to reboot to reset the state and try different configurations, this is not true for real platforms. This refactors the golang verification test suite to reboot less often, if at all.

#226

### DIFF
--- a/verification/verification_test.go
+++ b/verification/verification_test.go
@@ -8,40 +8,20 @@ import (
 	"testing"
 )
 
-var target_exe *string
 var testTargetType string
 
 // This will be called before running tests, and it assigns the socket path based on command line flag.
 func TestMain(m *testing.M) {
-	target_exe = flag.String("sim", "../simulator/target/debug/simulator", "path to simulator executable")
+	TargetExe = flag.String("sim", "../simulator/target/debug/simulator", "path to simulator executable")
 
 	exitVal := m.Run()
 	os.Exit(exitVal)
 }
 
 func TestRunAll(t *testing.T) {
-	for _, test := range TestCases {
-		t.Run(test.Name, func(t *testing.T) {
-			d, err := GetTestTarget(test.SupportNeeded)
-			if err != nil {
-				t.Errorf("Unable to get test target: %v", err)
-			}
+	targets := GetSimulatorTargets()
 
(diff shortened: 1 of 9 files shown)


verification/verification_test.go
```diff
@@ -4,44 +4,58 @@ package verification
...
-		})
+	for _, target := range targets {
+		runTargetTestCases(target, t)
 	}
 }
 
-// Get the test target for simulator/emulator
-func GetTestTarget(supportNeeded []string) (TestDPEInstance, error) {
-	instance, err := GetSimulatorTarget(supportNeeded, *target_exe)
+func runTargetTestCases(target TestTarget, t *testing.T) {
```
jhand2, line 30: One possible issue here is that we're running tests differently depending on whether the target supports power control, which could lead to failures that happen on other targets but don't happen here.

I think there are two types of tests: Ones that need a reboot to succeed and ones that don't. Ideally we would always reboot the ones that really need it and don't reboot the others.

If a target doesn't support reboot control (or doesn't want to because of time constraints), they can return false from `HasPowerControl`. A test which doesn't work w/o power control can add

```
if !d.HasPowerControl() {
    t.Skipf("Some reason...")
}
```

As you mentioned in the PR description, I actually don't think we have any such tests today.

zhalvorsen: We discussed this offline. For posterity, here is what we talked about.

Now it tries to power on once per `TestTarget`. The `HasPowerControl` here check is to make sure the test device is powered before executing the test suite.

`TestTarget` can be thought of as either different platforms or separate test devices. Platforms that integrate with this library would probably only have one set of `Support` options. So there would only be one entry for that platform. Something like this could be added to `TestRunAll`.

```go
func TestRunAll(t *testing.T) {
    targets := GetSimulatorTargets()
    targets += {
        "CoolNewPlatform",
        CoolNewPlatform(),
        AllTestCases,
    }

    for _, target := range targets {
        runTargetTestCases(target, t)
    }
}
```

If a platform doesn't want to reboot, the test device could be left powered and then mark `HasPowerControl` as false.

jlmahowa-amd (approved):

jhand2 (approved):

PR #246, opened 2023-10-14
Exclude root node from certs if it doesn't contain any measurements

### LINKED ISSUE #244: Exclude root node from the cert
The root node is not a real node, so it's measurements should not be in the certs.

sree-revoori1:
There are two ways a context can get measurements - if it got created in derive child as a child context or if it's tci was extended. Thus, the root node can only have measurements if it's tci was extended. Otherwise, we should exclude adding it's measurements to the certs since they are default buffers of 0s.

This PR also adds support for extend_tci in the go verification tests so that we can call it before certifying key. This is necessary since we do not have derive_child implemented yet in the verification tests, so we can only ever have the root node to certify. If we skip adding the root node to the certs, then certify key would fail as there are no measurements to add to the certs, so we call extend_tci to create measurements in the root node.

fixes #244

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -42,8 +42,11 @@ pub struct DpeInstance {
     /// `InitializeContext(simulation=false)`) once per reset cycle.
     pub(crate) has_initialized: U8Bool,
 
+    /// Flag used to determine if the root node can be used in CDI/Key derivation
+    pub(crate) root_has_measurement: U8Bool,
+
     // unused buffer added to make DpeInstance word aligned and remove padding
-    reserved: [u8; 3],
+    reserved: [u8; 2],
 }
 
 impl DpeInstance {
@@ -64,7 +67,8 @@ impl DpeInstance {
             contexts: [CONTEXT_INITIALIZER; MAX_HANDLES],
             support,
             has_initialized: false.into(),
-            reserved: [0u8; 3],
+            root_has_measurement: false.into(),
+            reserved: [0u8; 2],
         };
 
         if dpe.support.auto_init() {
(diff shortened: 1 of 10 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -264,9 +268,11 @@ impl DpeInstance {
                 return Err(DpeErrorCode::InternalError);
             }
 
-            // TODO: The root node isn't a real node with measurements and
-            // shouldn't be in the cert. But we don't support DeriveChild yet,
-            // so this is the only node we can create to test cert creation.
+            // Skip adding the root node if it doesn't contain any measurements.
+            if curr.parent_idx == Context::ROOT_INDEX && !self.root_has_measurement.get() {
+                break;
+            }
```
jhand2, line 274: I'm wondering if it is relevant to always include the root measurement in the key derivation. Here's why:

Tree 1:

```
root(0x00) -> DeriveChild(0x00) -> DeriveChild(0x01)
                 ^
                 |
            ExtendTCI(0xAA
```

Tree 2:
```
root(0x00) -> DeriveChild(0x01)
     ^
     |
ExtendTCI(0xAA)
```

These will have the same key derived and the same certificate, but they technically took different boot paths.

When we talked about this previously, I wasn't thinking about the fact that you could technically call ExtendTCI on the root node.

sree-revoori1: Yeah that makes sense to me. But maybe we could still exclude the root measurement if the DPE does not support ExtendsTci?

verification/verification.go
```diff
@@ -25,10 +25,10 @@ var InitializeContextSimulationTestCase = TestCase{
 	"InitializeContextSimulation", TestInitializeSimulation, []string{"Simulation"},
 }
 var CertifyKeyTestCase = TestCase{
-	"CertifyKey", TestCertifyKey, []string{"AutoInit", "X509", "IsCA"},
+	"CertifyKey", TestCertifyKey, []string{"AutoInit", "X509", "IsCA", "ExtendTci"},
 }
 var CertifyKeySimulationTestCase = TestCase{
-	"CertifyKeySimulation", TestCertifyKey_SimulationMode, []string{"AutoInit", "Simulation", "X509", "IsCA"},
+	"CertifyKeySimulation", TestCertifyKey_SimulationMode, []string{"AutoInit", "Simulation", "X509", "IsCA", "ExtendTci"},
 }
```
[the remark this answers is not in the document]
sree-revoori1: The problem is, we need ExtendTci here since otherwise we can only call certify key with the root, but key derivation would fail since there is no measurement to include. The other option is to wait until derive child is supported in the verification tests

PR #247, opened 2023-10-24
Preserve platform error codes in DpeErrorCode

jhand2:
The DPE Platform and Crypto APIs both implement error enums, but they may mask more detailed platform-specific error codes. Make the DPE error enums more expressive so they can suface these details errors.

Ultimately, it is up to the platform to decide how to surface this extended error information.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -36,13 +36,25 @@ impl AlgLen {
     }
 }
 
+// For errors which come from lower layers, include the error code returned
+// from platform libraries.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(u16)]
 pub enum CryptoError {
-    AbstractionLayer,
-    CryptoLibError,
-    Size,
-    NotImplemented,
-    HashError,
+    AbstractionLayer(u32) = 0x1,
+    CryptoLibError(u32) = 0x2,
+    Size = 0x3,
+    NotImplemented = 0x4,
+    HashError(u32) = 0x5,
+}
+
+impl CryptoError {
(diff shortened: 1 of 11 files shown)


dpe/src/response.rs
```diff
@@ -161,8 +164,50 @@ pub enum DpeErrorCode {
...
+            DpeErrorCode::Crypto(e) => match e {
+                CryptoError::AbstractionLayer(code) => Some(*code),
+                CryptoError::CryptoLibError(code) => Some(*code),
+                CryptoError::Size => None,
+                CryptoError::NotImplemented => None,
+                CryptoError::HashError(code) => Some(*code),
+            },
+            _ => None,
+        }
+    }
```
zhalvorsen, line 212: optional: it might be worth adding a couple `From`s here as well to simplify some more of the conversions

crypto/src/lib.rs
```diff
@@ -36,13 +36,25 @@ impl AlgLen {
     }
 }
 
+// For errors which come from lower layers, include the error code returned
+// from platform libraries.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(u16)]
```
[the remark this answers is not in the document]
jhand2: Copy/paste error. I mean to make this u16 so it is safe to or it with the `Platform` error code (which sets only high bits) to get a full error code.

dpe/src/response.rs
jhand2: Good idea, that's much nicer :)

zhalvorsen (approved):

PR #251, opened 2023-11-02
Sign support

hpya93:
Hi @jhand2 
Test for Sign command is added separately.
The following are tested in this PR
- Testcase for Sign command in Simulation context
- Testcase for Symmetric Signing
- Testcase for Asymmetric signing and verification

Update:
- Bug in Rust side is fixed, Instead of setting Flag bit 30, flag bit 31 is set for symmetric signing.   (merged in PR 255)
- These items were covered in a common testcases (merged in PR 255)
     - Sign command in Default context
     - Wrong Locality
     - Context handle


verification/abi.go
```diff
@@ -404,6 +423,18 @@ func (c *dpeABI[CurveParameter, Digest]) CertifyKeyABI(cmd *CertifyKeyReq[Digest
 	}, nil
 }
 
+// Sign calls the DPE Sign command.
+func (c *dpeABI[_, Digest]) SignABI(cmd *SignReq[Digest]) (*SignResp[Digest], error) {
```
jhand2, line 427: I think this needs to be rebased with main

hpya93: > I think this needs to be rebased with main

This is rebased can you please review the rebased code?

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+	// Get digest size
+	profile, err := GetTransportProfile(d)
+	if err != nil {
+		t.Fatalf("Could not get profile: %v", err)
+	}
+
+	digestLen := profile.GetDigestSize()
+	if digestLen != 32 && digestLen != 48 {
+		t.Fatalf("[FATAL]: Unsupported digest size.")
+	}
```
jhand2, line 28: nit: probably not necessary. The client will already fail if the lengths are wrong.

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		t.Fatalf("Could not get profile: %v", err)
+	}
+
+	digestLen := profile.GetDigestSize()
+	if digestLen != 32 && digestLen != 48 {
+		t.Fatal("[FATAL]: Unsupported digest size.")
+	}
+
+	if _, err := c.Sign(handle, make([]byte, digestLen), SignFlags(0), make([]byte, digestLen)); err == nil {
```
jhand2, line 52: nit: Should we also check that Symmetric sign fails?

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Fatalf("[FATAL]: Error while signing %v", err)
+	}
+
+	// Rerun with same label and compare signature emitted.
+	signedDataWithSameLabel, err := c.Sign(handle, label, SignFlags(IsSymmetric), tbs)
+	if err != nil {
+		t.Fatalf("[FATAL]: Error while signing %v", err)
+	}
+
+	// It is sufficient to check signatureR, signatureS is all zeroes
```
jhand2, line 103:
```suggestion
	// Symmetric sign only populates HmacOrSignatureR. SignatureS is all zeroes.
```

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if bytes.Equal(signedDataWithDiffLabel.HmacOrSignatureR, signedData.HmacOrSignatureR) {
+		t.Errorf("[ERROR]: Signature must vary for different label despite having same toBeSigned content, want new signature but got old %v", signedData.HmacOrSignatureR)
+	}
+}
+
+// Checks whether the digital signature returned by Sign command can be verified
+// using public key in signing key certificate returned by CertifyKey command.
+// Inspite of the DPE profile supporting symmetric key, for symmetric signing it must be enabled
+// explicitly in Sign command. Else asymmetric signing is used as default.
```
jhand2, line 127:
```suggestion
// explicitly in Sign command flags. Else asymmetric signing is used as default.
```

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+	x := new(big.Int).SetBytes(certifiedKey.Pub.X)
+	y := new(big.Int).SetBytes(certifiedKey.Pub.Y)
+
+	if digestLen == 32 {
+		ec = elliptic.P256()
+	} else if digestLen == 48 {
+		ec = elliptic.P384()
+	}
+
+	publicKey := ecdsa.PublicKey{Curve: ec, X: x, Y: y}
```
jhand2, line 170: Can you add a check that `publicKey` and `leafCert.PublicKey` are the same key? (Would also be fine if you'd rather make this check in the CertifyKey test instead)

verification/sign.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Error("Signature Verification failed")
+	}
+
+	// Validate certificate chain validation establish the CA of signing key certificate
+	certChainBytes, err := c.GetCertificateChain()
+	if err != nil {
+		t.Fatalf("[FATAL]: Could not get Certificate Chain: %v", err)
+	}
+	certChain := checkCertificateChain(t, certChainBytes)
+	validateLeafCertChain(t, certChain, leafCert)
```
jhand2, line 188: nit: Probably not necessary. This is already tested in the CertifyKey test

verification/tagTCI.go
jhand2, line 1: Can you put the TagTCI tests in a different PR? Since Sign tests are already quite a bit of code, it's easier to review them separately.

hpya93: Hi, I created this PR https://github.com/chipsalliance/caliptra-dpe/pull/253 has ExtendTCI commands alone.

verification/sign.go
hpya93: Have added this in CertifyKey.go, but yet to raise as PR.
I will be raising as a separate PR, as I have considerable amount of changes in this file. 
Adding in this PR will again increase the changes and make review difficult.

I have tested in that file as below
```
// Checks CertifyKey command response against public key extracted from certificate returned in response
func checkCertifyKeyResponse(t *testing.T, x509Cert *x509.Certificate, response CertifiedKey, hashAlg asn1.ObjectIdentifier) {
	var err error

	publicKeyDer, err := x509.MarshalPKIXPublicKey(x509Cert.PublicKey)
	if err != nil {
		t.Fatalf("[FATAL]: Could not marshal pub key: %v", err)
	}

	// Parse the DER-encoded public key
	pubKey, err := x509.ParsePKIXPublicKey(publicKeyDer)
	if err != nil {
		t.Fatalf("[FATAL]: Failed to parse DER-encoded public key: %v", err)
	}

	if _, ok := pubKey.(*ecdsa.PublicKey); !ok {
		t.Fatal("[FATAL]: Public key is not a ecdsa key")
	}

	var pubKeyInResponse ecdsa.PublicKey

	if hashAlg.Equal(OidSHA384) {
		pubKeyInResponse = ecdsa.PublicKey{
			Curve: elliptic.P384(),
			X:     new(big.Int).SetBytes(response.Pub.X),
			Y:     new(big.Int).SetBytes(response.Pub.Y),
		}
	} else if hashAlg.Equal(OidSHA256) {
		pubKeyInResponse = ecdsa.PublicKey{
			Curve: elliptic.P256(),
			X:     new(big.Int).SetBytes(response.Pub.X),
			Y:     new(big.Int).SetBytes(response.Pub.Y),
		}
	} else {
		t.Errorf("[ERROR]: Unsupported hash algorithm.")
		return
	}

	if !(pubKeyInResponse.Equal(pubKey)) {
		t.Errorf("[ERROR]: Public key returned in response must match the Public Key Info in the leaf certificate.")
	}
}

```

verification/sign.go
hpya93: Changed this

verification/sign.go
hpya93: Changed this

verification/sign.go
hpya93: Ok, removed

verification/sign.go
hpya93: ok removed

verification/certifyKey.go
```diff
@@ -433,9 +433,11 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
...
 func testCertifyKey(d TestDPEInstance, client DPEClient, t *testing.T, simulation bool) {
 	ctx := getInitialContextHandle(d, client, t, simulation)
-	if simulation {
-		defer client.DestroyContext(ctx, 0)
-	}
+	defer func() {
+		if simulation {
+			client.DestroyContext(ctx, 0)
+		}
+	}()
```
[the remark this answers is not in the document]
hpya93: - Yes, they are different especially in Simulation mode. 
- If defer "function call" is used instead of defer "function block", it causes error and subsequent Go tests break with "Invalid handle" error.
- This is because some "older" context handle is passed to DestroyContext() instead of "newest" context handle.  

#### **Here is why:**
    - In simulation mode, the "ctx" value keeps changing unlike default context (because CertifyKey in default context will return default handle only.)
    - The problem arises due to incorrect evaluation & assignment to "ctx" variable.
    - When we use defer, we expect that "ctx" must hold the new context handle.
    - But, it behaves differently for "function call" and "anonymous function call" as below.
    - **In case of deferred function call,** even though the DestroyContext() is call is deferred, the value of "ctx" is assigned beforehand. So, DestroyContext() tries to delete an older context (that doesn't even exist)  - This breaks subsequent tests with "Invalid handle error"
    - But **in case of deferred anonymous function call,**   the anonymous function block is evaluated at the end only. "ctx" variable has the "newest" context handle. DestroyContext() works as expected.
    https://stackoverflow.com/a/47321051

verification/tagTCI.go
jhand2: I would strongly recommend doing non-overlapping work in separate branches with different commits. This will make it much easier to merge your PRs independently (i.e. it doesn't matter what order they get merged in).

verification/sign.go
jhand2: I don't see this change. Do you need to push to the branch?

hpya93: I see the changes in line:48, do you expect some other change? Am I misunderstanding something?

hpya93: image

verification/tagTCI.go
hpya93: TagTci.go is the overlapping file that is causing issue.
TagTci.go was already present but without much validation.
I had to add new tests in existing file and this was interfering with changes made for other commands.

hpya93:
Synced branch with PR258, PR254 merge

verification/tagTCI.go
hpya93: I have created separate file for extendTCI tests and is now merged with main branch

verification/certifyKey.go
```diff
@@ -630,3 +655,58 @@ func getKeyUsageNames(keyUsage x509.KeyUsage) []string {
 
 	return keyUsageNames
 }
+
+// Checks CertifyKey command response against public key extracted from certificate returned in response
+func checkCertifyKeyResponse(t *testing.T, x509Cert *x509.Certificate, response CertifiedKey, hashAlg asn1.ObjectIdentifier) {
+	var err error
```
hpya93, line 661: Check that verifies if key returned in CertifyKey response and in the Leaf certificate are the same.

verification/certifyKey.go
```diff
@@ -630,3 +655,58 @@ func getKeyUsageNames(keyUsage x509.KeyUsage) []string {
...
+
+	if !(pubKeyInResponse.Equal(pubKeyInCert)) {
+		t.Errorf("[ERROR]: Public key returned in response must match the Public Key Info in the certificate.")
+	}
+}
+
+// Checks whether the context handle is unchanged after certifyKey command when default context handle is used.
+func checkCertifyKeyRespHandle(res CertifiedKey, t *testing.T, handle *ContextHandle) {
+	if *handle != DefaultContextHandle {
+		t.Logf("[LOG]: Handle is not default context, skipping check...")
```
hpya93, line 705: A simple check that esures default context handle is unaffected by CertifyKey command

verification/simulator.go
```diff
@@ -230,7 +230,7 @@ func GetSimulatorTargets() []TestTarget {
 		},
 		{
 			"DefaultSupport",
-			getTestTarget([]string{"AutoInit", "Simulation", "X509", "IsCA", "Tagging", "RotateContext", "ExtendTci"}),
+			getTestTarget([]string{"AutoInit", "Simulation", "X509", "IsCA", "Tagging", "RotateContext", "ExtendTci", "IsSymmetric"}),
```
hpya93, line 233: Required to test symmetric signing

verification/verification.go
```diff
@@ -30,7 +30,7 @@ var CertifyKeyTestCase = TestCase{
 	"CertifyKey", TestCertifyKey, []string{"AutoInit", "X509", "IsCA"},
 }
 var CertifyKeySimulationTestCase = TestCase{
-	"CertifyKeySimulation", TestCertifyKey_SimulationMode, []string{"AutoInit", "Simulation", "X509", "IsCA"},
+	"CertifyKeySimulation", TestCertifyKeySimulation, []string{"AutoInit", "Simulation", "X509", "IsCA"},
 }
```
hpya93, line 34: For consistent naming

verification/certifyKey.go
```diff
@@ -120,18 +123,23 @@ const (
 
 type TcgMultiTcbInfo = []DiceTcbInfo
 
+type CertifyKeyParams struct {
+	Label []byte
+	Flags CertifyKeyFlags
+}
+
```
hpya93, line 130: This struct declaration is moved out of testCertifyKey function to reduce the function size.

verification/certifyKey.go
```diff
@@ -362,25 +375,25 @@ func checkCertifyKeyExtensions(t *testing.T, c *x509.Certificate) {
...
 }
 
-// Validate basic constraints in certificate returned by CertifyKey command
+// Validates basic constraints in certificate returned by CertifyKey command
 // against the flag set for input parameter.
 // The BasicConstraints extension MUST be included
 // If CertifyKey AddIsCA is set, IsCA MUST be set to true.
 // If CertifyKey AddIsCA is NOT set, IsCA MUST be set to false
-func checkCertifyKeyBasicConstraints(t *testing.T, c *x509.Certificate, flags uint32) {
+func checkCertifyKeyBasicConstraints(t *testing.T, c *x509.Certificate, flags CertifyKeyFlags) {
```
hpya93, line 383: unnecessary type casting

verification/certifyKey.go
```diff
@@ -474,44 +488,48 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
 	return x509Cert
 }
 
-func testCertifyKey(d TestDPEInstance, client DPEClient, t *testing.T, simulation bool) {
-	ctx := getInitialContextHandle(d, client, t, simulation)
+func testCertifyKey(d TestDPEInstance, c DPEClient, t *testing.T, simulation bool) {
+	handle := getInitialContextHandle(d, c, t, simulation)
```
hpya93, line 492: just a consistent naming

verification/certifyKey.go
```diff
@@ -474,44 +488,48 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
...
 
+	var hashAlg asn1.ObjectIdentifier
+	if digestLen == 32 {
+		hashAlg = OidSHA256
+	} else if digestLen == 48 {
+		hashAlg = OidSHA384
+	} else {
+		t.Fatal("Unknown Hash Algorithm")
+	}
+
```
hpya93, line 513: This is required to build pub key from CertifyKey response which in turn is validated

verification/certifyKey.go
```diff
@@ -474,44 +488,48 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
...
 	seqLabel := make([]byte, digestLen)
 	for i := range seqLabel {
 		seqLabel[i] = byte(i)
 	}
 
-	certifyKeyParams := []Params{
+	certifyKeyParams := []CertifyKeyParams{
```
hpya93, line 519: Descriptive name

verification/sign.go
```diff
@@ -0,0 +1,170 @@
...
+	valid := ecdsa.Verify(&publicKey, tbs, r, s)
+	if !valid {
+		t.Error("Signature Verification failed")
+	}
+}
+
+// Check command fails in simulated context because this context does not allow signing.
+// This is because simulation context does not allow using context's private key.
+func TestSignSimulation(d TestDPEInstance, c DPEClient, t *testing.T) {
+	useSimulation := true
```
hpya93, line 87: Checks to ensure that Simulation mode does not allow Symmetric/Asymetric signing.

verification/sign.go
```diff
@@ -0,0 +1,170 @@
...
+	"math/big"
+	"testing"
+)
+
+// Obtain and validate signature of asymmetric signing.
+// Check whether the digital signature returned by Sign command can be verified
+// using public key in signing key certificate returned by CertifyKey command.
+// Inspite of the DPE profile supporting symmetric key, for symmetric signing it must be enabled
+// explicitly in Sign command flags. Else asymmetric signing is used as default.
+func TestAsymmetricSigning(d TestDPEInstance, c DPEClient, t *testing.T) {
```
hpya93, line 20: Checks to ensure that the data signed by DPE is verifiable through ecdsa.verify()

verification/sign.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Fatalf("[FATAL]: Incorrect error type. Should return %q, but returned %q", StatusInvalidArgument, err)
+	}
+}
+
+// Obtain HMAC (symmetric signature) generated and compare for varying label inputs.
+// Signature created is deterministic and depends on label passed to command.
+// This is because label is used by DPE in symmetric key derivation.
+// Invoking Sign command multiple times with same label and same content (TBS) should return same signature
+// but it should return different signatures for different labels despite having the same content (To Be Signed content).
+func TestSymmetricSigning(d TestDPEInstance, c DPEClient, t *testing.T) {
```
hpya93, line 119: Checks to ensure that signature of TBS changes when input label is changed. There is no way to verify HMAC and encryption key is protected by DPE and is never exposed as it is a symmetric key.

verification/extendTCI.go
```diff
@@ -50,10 +51,108 @@ func TestExtendTCI(d TestDPEInstance, c DPEClient, t *testing.T) {
...
 
 	// Cross-check current and cumulative measurement by CertifyKey
-	verifyMeasurementsByCertifyKey(c, t, handle, defaultTci, tciValue, hasher)
+	verifyMeasurementsByCertifyKey(c, t, handle, defaultTci, tciValue, false, hasher)
 }
 
-func verifyMeasurementsByCertifyKey(c DPEClient, t *testing.T, handle *ContextHandle, label []byte, tciValue []byte, hasher hash.Hash) {
+// Check whether the ExtendTCI command with derived child context.
+func TestExtendTciOnDerivedContexts(d TestDPEInstance, c DPEClient, t *testing.T) {
+	var err error
```
hpya93, line 59: This tests ExtendTCI on derived context using CertifyKey instead of GettaggedTCI (which was removed in PR 257).

jhand2: Please restrict this PR to only sign tests

hpya93:
Synced code

jhand2 (approved):

verification/extendTCI.go
hpya93: Removed the other changes, synced this branch with latest as it had conflicts.

jhand2 (approved):

PR #255, opened 2023-11-03
Common client code and few bug fixes

hpya93:
Hi @jhand2 

Have added
- **Files: abi.go, client.go**: Code to issue RotateContextHandle, DeriveChild, ExtendTCI and Sign commands
- **File- sign.rs**: Bug fix for IS_SYMMETRIC flag in sign.rs file
- **File- sign.rs**: Bug fix  for DPE error code in sign.rs. It must be "ArgumentNotSupported" instead of "InvalidArgument" when DPE profile does not support a command flag.
- **File - CertifyKey.go, TagTci.go**:  Bug fix, the context handle created for Simulation mode is not used in code. The contexthandle is stored in variable "handle", but variable "ctx" is used
- **File - InitializeContext.go**: Separated out code for simulated context handle creation and put it as a func in InitializeContext.go file. This block was repeated many times.
-  **Files - negativeCases.go, simulator.go, verification.go**: Added a negativeCases.go - it will test InvalidHandle, wrongLocality, Unsupported DPE command, Unsupported command flags. Tried to address your comment in TagTCI PR by writing a separate function for checking WrongLocality and other negative cases..   

When this is merged I shall change other PRs in tune to the main branch with these changes. Kindly let us know the comments.

### DIFF
--- a/dpe/src/commands/sign.rs
+++ b/dpe/src/commands/sign.rs
@@ -15,7 +15,7 @@ pub struct SignFlags(u32);
 
 bitflags! {
     impl SignFlags: u32 {
-        const IS_SYMMETRIC = 1u32 << 31;
+        const IS_SYMMETRIC = 1u32 << 30;
     }
 }
 
@@ -83,7 +83,7 @@ impl CommandExecution for SignCmd {
     ) -> Result<Response, DpeErrorCode> {
         // Make sure the operation is supported.
         if !dpe.support.is_symmetric() && self.uses_symmetric() {
-            return Err(DpeErrorCode::InvalidArgument);
+            return Err(DpeErrorCode::ArgumentNotSupported);
         }
 
         let idx = dpe.get_active_context_pos(&self.handle, locality)?;
@@ -190,7 +190,7 @@ mod tests {
 
         // Bad argument
         assert_eq!(
-            Err(DpeErrorCode::InvalidArgument),
+            Err(DpeErrorCode::ArgumentNotSupported),
             SignCmd {
                 handle: ContextHandle([0xff; ContextHandle::SIZE]),
                 label: TEST_LABEL,
--- a/verification/abi.go
+++ b/verification/abi.go
@@ -3,10 +3,11 @@
 package verification
 
 import (
-	"errors"
 	"fmt"
 	"reflect"
 )
+
+var DefaultContextHandle = ContextHandle{0}
 
 const (
 	CmdMagic  uint32 = 0x44504543
@@ -35,9 +36,13 @@ type Support struct {
 const (
 	CommandGetProfile          CommandCode = 0x1
 	CommandInitializeContext   CommandCode = 0x7
+	CommandDeriveChild         CommandCode = 0x8
 	CommandCertifyKey          CommandCode = 0x9
+	CommandSign                CommandCode = 0xa
+	CommandRotateContextHandle CommandCode = 0xe
 	CommandDestroyContext      CommandCode = 0xf
 	CommandGetCertificateChain CommandCode = 0x80
+	CommandExtendTCI           CommandCode = 0x81
 	CommandTagTCI              CommandCode = 0x82
 	CommandGetTaggedTCI        CommandCode = 0x83
 )
@@ -156,6 +161,74 @@ type GetTaggedTCIResp[Digest DigestAlgorithm] struct {
 	CurrentTCI    Digest
 }
 
+type RotateContextHandleFlags uint32
+
+const (
+	TargetIsDefault RotateContextHandleFlags = 1 << 31
+)
+
+type RotateContextHandleCmd struct {
+	Handle ContextHandle
+	Flags  RotateContextHandleFlags
+}
+
+type RotatedContextHandle struct {
+	NewContextHandle ContextHandle
+}
+
+type DeriveChildFlags uint32
+
+const (
+	InternalInputInfo DeriveChildFlags = 1 << 31
+	InternalInputDice DeriveChildFlags = 1 << 30
+	RetainParent      DeriveChildFlags = 1 << 29
+	MakeDefault       DeriveChildFlags = 1 << 28
+	ChangeLocality    DeriveChildFlags = 1 << 27
+	InputAllowCA      DeriveChildFlags = 1 << 26
+	InputAllowX509    DeriveChildFlags = 1 << 25
+)
+
+type DeriveChildReq[Digest DigestAlgorithm] struct {
+	ContextHandle  ContextHandle
+	InputData      Digest
+	Flags          DeriveChildFlags
+	TciType        uint32
+	TargetLocality uint32
+}
+
+type DeriveChildResp struct {
+	NewContextHandle    ContextHandle
+	ParentContextHandle ContextHandle
+}
+
+type SignFlags uint32
+
+const (
+	IsSymmetric SignFlags = 1 << 30
+)
+
+type SignReq[Digest DigestAlgorithm] struct {
+	ContextHandle ContextHandle
+	Label         Digest
+	Flags         SignFlags
+	ToBeSigned    Digest
+}
+
+type SignResp[Digest DigestAlgorithm] struct {
+	NewContextHandle ContextHandle
+	HmacOrSignatureR Digest
+	SignatureS       Digest
+}
+
+type ExtendTCIReq[Digest DigestAlgorithm] struct {
+	ContextHandle ContextHandle
+	InputData     Digest
+}
+
+type ExtendTCIResp struct {
+	NewContextHandle ContextHandle
+}
+
 // dpeABI is a connection to a DPE instance, parameterized by hash algorithm and ECC curve.
 type dpeABI[CurveParameter Curve, Digest DigestAlgorithm] struct {
 	transport    Transport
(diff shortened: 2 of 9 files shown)


verification/abi.go
```diff
@@ -3,11 +3,12 @@
 package verification
 
 import (
-	"errors"
 	"fmt"
 	"reflect"
 )
 
+var DefaultContextHandle = ContextHandle{0}
+
```
hpya93, line 11: Couldnt declare as constant as arrays are not allowed to be constants.

verification/certifyKey.go
```diff
@@ -431,26 +431,10 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
...
-		} else {
-			t.Errorf("[ERROR]:  DPE instance doesn't support simulation contexts.")
-		}
-	} else {
-		//default context
-		ctx = [16]byte{0}
+func testCertifyKey(d TestDPEInstance, client DPEClient, t *testing.T, simulation bool) {
+	ctx := getContextHandle(d, client, t, simulation)
+	if simulation {
+		defer client.DestroyContext(ctx, 0)
```
hpya93, line 437: Have put  this code block as a function, since it is getting repeated for each command using Simulation mode.

verification/negativeCases.go
```diff
@@ -0,0 +1,275 @@
+// Licensed under the Apache-2.0 license
```
hpya93, line 1: File that tests WrongLoclaity, invalid handle, Unsupported command and UnsupportedCommand flags

verification/certifyKey.go
```diff
@@ -500,6 +484,11 @@ func testCertifyKey(d TestDPEInstance, client DPEClient, t *testing.T, use_simul
 		// This also checks certificate lifetime, signatures as part of cert chain validation
 		validateLeafCertChain(t, certChain, leafCert)
 
+		// Reassign handle for simulation mode.
+		// However, this does not impact in default mode because
+		// same default context handle is returned in default mode.
+		ctx = &certifyKeyResp.Handle
+
```
hpya93, line 491: This is required, else SimulationMode test fails after properly assigning simulation context handle

verification/tagTCI.go
```diff
@@ -8,16 +8,14 @@ import (
...
-			t.Fatalf("Failed to initialize default context: %v", err)
-		}
-		defer client.DestroyContext(handle, DestroyDescendants)
+	testTagTCI(d, client, t, false)
+}
+
+func testTagTCI(d TestDPEInstance, client DPEClient, t *testing.T, simulation bool) {
+	ctx := getContextHandle(d, client, t, simulation)
+	if simulation {
+		defer client.DestroyContext(ctx, DestroyDescendants)
```
hpya93, line 18: same problem here, handle is initialiaed but never used, also its scope is only within the IF block

verification/verification.go
```diff
@@ -39,6 +41,18 @@ var TagTCITestCase = TestCase{
 var GetProfileTestCase = TestCase{
 	"GetProfile", TestGetProfile, []string{},
 }
+var InvalidHandleTestCase = TestCase{
+	"CheckInvalidHandle", TestInvalidHandle, []string{"Simulation", "RotateContext", "ExtendTci", "Tagging"},
+}
+var WrongLocalityTestCase = TestCase{
+	"CheckWrongLocality", TestWrongLocality, []string{"AutoInit", "RotateContext", "ExtendTci", "Tagging"},
+}
+var UnsupportedCommand = TestCase{
```
hpya93, line 50: Have added limited DPE support to test failure

dpe/src/commands/sign.rs
```diff
@@ -15,7 +15,7 @@ pub struct SignFlags(u32);
 
 bitflags! {
     impl SignFlags: u32 {
-        const IS_SYMMETRIC = 1u32 << 31;
+        const IS_SYMMETRIC = 1u32 << 30;
```
hpya93, line 18: A minor deviation from irot spec

dpe/src/commands/sign.rs
```diff
@@ -83,7 +83,7 @@ impl CommandExecution for SignCmd {
     ) -> Result<Response, DpeErrorCode> {
         // Make sure the operation is supported.
         if !dpe.support.is_symmetric() && self.uses_symmetric() {
-            return Err(DpeErrorCode::InvalidArgument);
+            return Err(DpeErrorCode::ArgumentNotSupported);
```
hpya93, line 86: All other commands return ArgumentNotSupported. Only Sign comamnd is returning a different error code, so changed it.

dpe/src/commands/sign.rs
```diff
@@ -190,7 +190,7 @@ mod tests {
 
         // Bad argument
         assert_eq!(
-            Err(DpeErrorCode::InvalidArgument),
+            Err(DpeErrorCode::ArgumentNotSupported),
```
hpya93, line 193: corresponding change added to rust testcase

verification/initializeContext.go
```diff
@@ -67,3 +67,27 @@ func testInitContext(d TestDPEInstance, client DPEClient, t *testing.T, simulati
 		}
 	}
 }
+
+func getContextHandle(d TestDPEInstance, c DPEClient, t *testing.T, simulation bool) *ContextHandle {
```
jhand2, line 71: nit: rename to `getInitialContextHandle`? Also would be good to have a function level comment explaining what this function is for.

verification/initializeContext.go
```diff
@@ -67,3 +67,27 @@ func testInitContext(d TestDPEInstance, client DPEClient, t *testing.T, simulati
...
+	var handle *ContextHandle
+	var err error
+	if simulation {
+		if d.GetSupport().Simulation {
+			handle, err = c.InitializeContext(InitIsSimulation)
+			if err != nil {
+				t.Fatal("The instance should be able to create a simulation context.")
+			}
+			// Could prove difficult to prove it is a cryptographically secure random.
```
jhand2, line 80: That's fine, I don't think it's necessary. Can remove this comment.

verification/negativeCases.go
```diff
@@ -0,0 +1,275 @@
...
+var InvalidHandle = ContextHandle{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}
+
+func TestInvalidHandle(d TestDPEInstance, c DPEClient, t *testing.T) {
+	testInvalidHandle(d, c, t)
+}
+
+// Checks whether error is reported when non-existent handle is passed as input to DPE commands.
+// Exceptions are - GetProfile, InitializeContext, GetCertificateChain, GetTaggedTCI commands
+// which do not need context handle as input parameter.
+func testInvalidHandle(d TestDPEInstance, c DPEClient, t *testing.T) {
```
jhand2, line 19: All this code can just be in the public function instead of calling a separate helper (same for all of these tests)

verification/tagTCI.go
jhand2, line 18: This doesn't need to be in a helper funciton

verification/verification.go
```diff
@@ -39,6 +41,18 @@ var TagTCITestCase = TestCase{
...
+	"CheckSupportForCommand", TestUnsupportedCommand, []string{"AutoInit"},
+}
+var UnsupportedCommandFlag = TestCase{
+	"CheckSupportForCommmandFlag", TestUnsupportedCommandFlag, []string{"AutoInit", "RotateContext", "ExtendTci", "Tagging"},
+}
```
jhand2, line 55: Can you add all of these to the `AllTestCases` array?

verification/initializeContext.go
hpya93: Yes renamed and added a description.

verification/initializeContext.go
hpya93: This was there already in original code, but removed now.

verification/negativeCases.go
hpya93: Removed the helper, put the logic under public function

verification/tagTCI.go
hpya93: Removed the helper, put the logic under public function

verification/verification.go
hpya93: **I have some problems with adding.**
1. The AllTestCases has a broad DPE suppport features, so it will not cause/catch errors in  UnsupportedCommand and UnsupportedCommandFlag  testcases, so left as it is.
2. However, I could add the Invalidhandle and Wrong locality test cases under AllTestCases without issues.

jhand2: I don't understand. AllTestCases is just a list if all the test cases. If the target doesn't have the support the test is requesting, the test will be skipped.

hpya93: Adding some more information for context:
1. AllTestCases is list of all testcases that is run with DefaultSupport test target which has more support features.
```
		{
			"DefaultSupport",
			getTestTarget([]string{"AutoInit", "Simulation", "X509", "IsCA", "Tagging", "RotateContext", "ExtendTci"}),
			AllTestCases,
		},
```
2. For **UnSupportedCommand testcase**, we **purposefully exclude** these support features to test for the "UnsupportedCommand" error code - especially **Tagging, RotateContext, ExtendTCI**
4. Similarly, for **UnSupportedCommandFlag testcase**, we **purposefully exclude** these support features to test for the "UnSupportedArgument" error code - especially **Simulation, IsCA, X509, ** 
5. Adding these tests in AllTestCases will ultimately test these two negative tests will Default support test target.
6. **If this happens, some cases in "UnSupportedCommand " and "UnSupportedCommandFlag" will fail as expected error will not be returned.**
7. Here is how it fails
```
    --- FAIL: TestRunAll/DefaultSupport-CheckSupportForCommand (0.00s)
    --- FAIL: TestRunAll/DefaultSupport-CheckSupportForCommmandFlag (0.01s)

```
- It gives many Error messages like so are given when testing ....
```
negativeCases.go:248: [ERROR]: Should return "argument is not supported by this profile, implementation, or integration", but returned no error

```

jhand2: Thanks for the explanation. I have opened https://github.com/chipsalliance/caliptra-dpe/issues/256 to track supporting this. We can merge this PR without this feature though.

jhand2 (approved):

PR #273, opened 2023-12-01
Add CertifyKey CSR Verfication Test

jhand2:
Add a verification test for running CertifyKey in CSR mode.

### DIFF
--- a/platform/src/default.rs
+++ b/platform/src/default.rs
@@ -1,6 +1,6 @@
 // Licensed under the Apache-2.0 license
 
-use crate::{Platform, PlatformError, MAX_CHUNK_SIZE};
+use crate::{Platform, PlatformError, MAX_CHUNK_SIZE, MAX_SN_SIZE};
 use core::cmp::min;
 use openssl::x509::X509;
 
@@ -16,6 +16,12 @@ pub const TEST_CERT_CHAIN: &[u8] = include_bytes!("test_data/cert_256.der");
 
 #[cfg(feature = "dpe_profile_p384_sha384")]
 pub const TEST_CERT_CHAIN: &[u8] = include_bytes!("test_data/cert_384.der");
+
+#[cfg(feature = "dpe_profile_p256_sha256")]
+pub const TEST_CERT_PEM: &[u8] = include_bytes!("test_data/cert_256.pem");
+
+#[cfg(feature = "dpe_profile_p384_sha384")]
+pub const TEST_CERT_PEM: &[u8] = include_bytes!("test_data/cert_384.pem");
 
 impl Platform for DefaultPlatform {
     fn get_certificate_chain(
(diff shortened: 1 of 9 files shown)


verification/certifyKey.go
```diff
@@ -133,6 +118,37 @@ func TestCertifyKeySimulation(d TestDPEInstance, c DPEClient, t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("[FATAL]: Could not certify key: %v", err)
+	}
+
+	wrappedCSR, err := pkcs7.Parse(certifyKeyResp.Certificate)
+	if err != nil {
+		t.Fatalf("[FATAL]: Could not unmarshal CSR CMS message: %v", err)
+	}
+
+	csr, err := x509.ParseCertificateRequest(wrappedCSR.Content)
```
sree-revoori1, line 144: Currently it looks like we are parsing the CMS, the CSR, and the extensions within the CRI in the CSR. Do you think it would be good to parse and validate some more fields? Probably overkill to do all of them but a few things I have in mind: 
- version in certification_request_info
- subject name in certification_request_info
- public key
- signature alg/signature value

jhand2: Done. One thing I found is that in SerialNumberAndIssuer, I think the Serial Number is supposed to be the issuer's serial number rather than the subject serial number.

I can't 100% tell from the spec, but otherwise the cert verification fails.

platform/src/default.rs
```diff
@@ -53,6 +59,20 @@ impl Platform for DefaultPlatform {
         Ok(issuer_name.len())
     }
 
+    fn get_issuer_sn(&mut self, out: &mut [u8; MAX_SN_SIZE]) -> Result<usize, PlatformError> {
+        let sn = X509::from_pem(TEST_CERT_PEM)
```
sree-revoori1, line 63: nit: maybe we should make a const value for the X509 representation of TEST_CERT_PEM so we don't have to repeat work in get_issuer_name and get_issuer_sn when converting the PEM to an X509?

sree-revoori1 (approved):

sree-revoori1:
Could you add something to https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.md#requirements-for-csrs to clarify what the IssuerAndSerialNumber field should be?

verification/certifyKey.go
sree-revoori1: Yeah I think you're right. Thanks for catching this.

jhand2:
> Could you add something to https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.md#requirements-for-csrs to clarify what the IssuerAndSerialNumber field should be?

Opened a PR at https://github.com/TrustedComputingGroup/Server-Internal/pull/142

PR #278, opened 2023-12-07
Allow DPE to be initialized with a measurement

sree-revoori1:
Add a function new_auto_init that allows DPE to be auto-initialized with some input data and tci type.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -74,6 +74,33 @@ impl DpeInstance {
         Ok(dpe)
     }
 
+    pub fn new_auto_init(
+        env: &mut DpeEnv<impl DpeTypes>,
+        support: Support,
+        tci_type: u32,
+        auto_init_measurement: [u8; DPE_PROFILE.get_hash_size()],
+    ) -> Result<DpeInstance, DpeErrorCode> {
+        // auto-init must be supported to add an auto init measurement
+        if !support.auto_init() {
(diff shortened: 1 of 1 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -74,6 +74,33 @@ impl DpeInstance {
...
+        let mut tmp_context = dpe.contexts[idx];
+        // add measurement to auto-initialized context
+        dpe.add_tci_measurement(
+            env,
+            &mut tmp_context,
+            &TciMeasurement(auto_init_measurement),
+            locality,
+        )?;
+        dpe.contexts[idx] = tmp_context;
+        dpe.contexts[idx].tci.tci_type = tci_type;
```
jhand2, line 100: I think once this is done we need to set has_initialized

dpe/src/dpe_instance.rs
```diff
@@ -871,4 +898,28 @@ pub mod tests {
         // validation fails on a graph with multiple connected components
         assert_eq!(dpe.validate_context_tree(0), false);
     }
+
+    #[test]
+    fn test_new_auto_init() {
```
jhand2, line 903: can you add a test to ensure `initialize_context` fails if new_auto_init was used.

sree-revoori1: Done

jhand2 (approved):

PR #285, opened 2023-12-11
Destroy dangling retired parent contexts in destroy_ctx

### LINKED ISSUE #283: Retired contexts not getting destroyed upon all children being destroyed
This could lead to a DOS by filling the context array with retired contexts.

sree-revoori1:
When DestroyContext is called, it is possible that the parents of the context to be destroyed is a chain of retired contexts. These should be destroyed, since otherwise it is impossible to clean them up since their handles are invalid. This could prevent a potential DOS attack.

Fixes #283

### DIFF
--- a/dpe/src/commands/destroy_context.rs
+++ b/dpe/src/commands/destroy_context.rs
@@ -1,9 +1,10 @@
 // Licensed under the Apache-2.0 license.
 use super::CommandExecution;
 use crate::{
-    context::ContextHandle,
-    dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
+    context::{Context, ContextHandle, ContextState},
+    dpe_instance::{flags_iter, DpeEnv, DpeInstance, DpeTypes},
     response::{DpeErrorCode, Response, ResponseHdr},
+    MAX_HANDLES,
 };
 
 #[repr(C)]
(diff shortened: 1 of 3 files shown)


dpe/src/commands/destroy_context.rs
```diff
@@ -227,6 +249,73 @@ mod tests {
...
+
+        DestroyCtxCmd { handle: handle_3 }
+            .execute(&mut dpe, &mut env, TEST_LOCALITIES[1])
+            .unwrap();
+
+        // only the auto-initialized context should remain, context[1] and context[2] should be
+        // destroyed since they are in the chain of consecutive retired parents of the destroyed
+        // context.
+        assert_eq!(
+            dpe.count_contexts(|ctx| ctx.state != ContextState::Inactive)
```
zhalvorsen, line 313: Should this also verify that it isn't in the retired state?

zhalvorsen (commented):
LGTM

dpe/src/commands/destroy_context.rs
```diff
@@ -26,7 +26,28 @@ impl CommandExecution for DestroyCtxCmd {
...
+        let mut retired_contexts = 0u32;
+        let mut parent_idx = context.parent_idx as usize;
+        loop {
+            if parent_idx == Context::ROOT_INDEX as usize {
+                break;
+            } else if parent_idx >= dpe.contexts.len() {
+                return Err(DpeErrorCode::InternalError);
+            }
+            let parent_context = dpe.contexts[parent_idx];
+            if parent_context.state == ContextState::Retired {
```
jhand2, line 39: Take the following sequence of DPE commands:

```
parent_handle = 0xAB;

child_a = DeriveChild(parent, retain_parent = true);
child_b = DeriveChild(parent, retain_parent = true);
child_c = DeriveChild(parent, retain_parent = false);

DestroyContext(child_a);
```

The 3rd DeriveChild call would retire `parent`. This DestroyContext call would destroy `parent`, but it is still a parent of both `child_a` and `child_b`.

So I think we can only destroy it if it has no other children that aren't being destroyed.

dpe/src/commands/destroy_context.rs
sree-revoori1: Done.

dpe/src/commands/destroy_context.rs
sree-revoori1: Good point, done.

jhand2 (approved):

PR #286, opened 2023-12-11
Implement full DPE validation

### LINKED ISSUE #279: Allow DPE to check for illegal state
Illegal state we need to check for: 

- Inactive contexts with children
- Inactive context with parent != Root context
- Inactive contexts with measurements
- Default/non-default contexts in the same locality
- Multiple default contexts in the same locality
- Duplicate handles in locality
- Dangling retired contexts
- Retired contexts with multiple children
- Children that don't exist
- Parents that don't exist
- TciNodeData locality != Context locality
- uses_internal_input_info or uses_internal_input_dice == true for inactive context
- Simulation context with supports_simulation = false
- Context having non-Normal, non-Simulation ContextType
- Context having non-Active, non-Inactive, non-Retired ContextState

### LINKED ISSUE #284: Multiple connected components bug in DPE Validation
We do not consider simulation contexts in DPE validation. DPE allows initialize context to be called on simulation contexts, even after the DPE has been initialized, meaning we could get multiple connected components in the context tree structure. However, DPE validation fails if the context tree has multiple connected components. 

My proposal: disable simulation contexts in Caliptra and then in validation, allow multiple connected components. This way, Caliptra's DPE context structure will always be a tree. We can still check that all the connected components represent valid trees, and make permitting multiple connected components dependent on a flag passed to the validator.

sree-revoori1:
DPE validation does two things:
1. Checks for illegal state in DPE
2. Checks that the DPE context tree is well-formed

2 was done in a previous commit, but this commit improves upon that implementation in a few ways:

1. Previously we were returning a bool from the context tree validation function. Now we return an error code so the caller knows exactly what is wrong with the shape of the DPE context tree
2. We weren't considering simulation contexts before. Simulation contexts forming a different connected component is not an error.

fixes #279 
fixes #284

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -96,6 +96,8 @@ impl Context {
         self.state = ContextState::Inactive;
         self.uses_internal_input_info = false.into();
         self.uses_internal_input_dice = false.into();
+        self.allow_ca = false.into();
+        self.allow_x509 = false.into();
         self.parent_idx = Self::ROOT_INDEX;
     }
 
--- a/dpe/src/validation.rs
+++ b/dpe/src/validation.rs
@@ -1,0 +1,567 @@
+// Licensed under the Apache-2.0 license.
+
+use crate::{
+    context::{Context, ContextState, ContextType},
+    dpe_instance::flags_iter,
+    response::DpeErrorCode,
+    tci::TciNodeData,
+    DpeInstance, MAX_HANDLES,
+};
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(u16)]
+/// It is possible that there are multiple issues with the DPE state. At most one will be found.
+/// There is no priority on which error will be found first if there are multiple.
+pub enum ValidationError {
+    MultipleNormalConnectedComponents = 0x0,
+    CyclesInTree = 0x1,
+    InactiveContextInvalidParent = 0x2,
+    InactiveContextWithChildren = 0x3,
+    BadContextState = 0x4,
+    BadContextType = 0x5,
+    InactiveContextWithMeasurement = 0x6,
+    MixedContextLocality = 0x7,
+    MultipleDefaultContexts = 0x8,
+    SimulationNotSupported = 0x9,
+    ParentDoesNotExist = 0xA,
+    InternalDiceNotSupported = 0xB,
+    InternalInfoNotSupported = 0xC,
+    ChildDoesNotExist = 0xD,
+    InactiveContextWithFlagSet = 0xE,
+    LocalityMismatch = 0xF,
+    DanglingRetiredContext = 0x10,
+    MixedContextTypeConnectedComponents = 0x11,
+    ChildWithMultipleParents = 0x12,
+    ParentChildLinksCorrupted = 0x13,
+    AllowCaNotSupported = 0x14,
+    AllowX509NotSupported = 0x15,
+    InactiveParent = 0x16,
+    InactiveChild = 0x17,
+    DpeNotMarkedInitialized = 0x18,
+}
+
+impl ValidationError {
+    pub fn discriminant(&self) -> u16 {
+        *self as u16
+    }
+}
+
+pub struct DpeValidator<'a> {
+    pub dpe: &'a mut DpeInstance,
+}
+
+impl<'a> DpeValidator<'a> {
+    /// Validates that the shape of the DPE instance is well-formed and that
+    /// there is no illegal state present within the DPE.
+    pub fn validate_dpe(&self) -> Result<(), DpeErrorCode> {
+        self.validate_dpe_state()
+            .map_err(DpeErrorCode::Validation)?;
+        self.validate_context_forest()
+            .map_err(DpeErrorCode::Validation)
+    }
+
+    /// Returns an error if there is any illegal state or inconsistencies
+    /// present within the DPE instance.
+    fn validate_dpe_state(&self) -> Result<(), ValidationError> {
+        for i in 0..MAX_HANDLES {
+            let context = &self.dpe.contexts[i];
+
+            self.check_support(context)?;
+
+            match context.state {
+                ContextState::Inactive => {
+                    self.validate_inactive_context(context)?;
+                }
+                ContextState::Active => {
+                    // has_initialized must be true if there is a normal, active context
+                    if context.context_type == ContextType::Normal && !self.dpe.has_initialized() {
+                        return Err(ValidationError::DpeNotMarkedInitialized);
+                    }
+                    self.check_children_and_parent(i)?;
+                }
+                ContextState::Retired => {
+                    self.check_children_and_parent(i)?;
+                    // retired contexts must have at least one child context
+                    if flags_iter(context.children, MAX_HANDLES).count() == 0 {
+                        return Err(ValidationError::DanglingRetiredContext);
+                    }
+                }
+                _ => {
+                    return Err(ValidationError::BadContextState);
+                }
+            }
+
+            if context.context_type != ContextType::Normal
+                && context.context_type != ContextType::Simulation
+            {
+                return Err(ValidationError::BadContextType);
+            }
+
+            if context.locality != context.tci.locality {
+                return Err(ValidationError::LocalityMismatch);
+            }
+        }
+
+        self.check_context_handles_per_locality()?;
+
+        Ok(())
+    }
+
+    /// Checks that the context fields do not violate supported flags
+    fn check_support(&self, context: &Context) -> Result<(), ValidationError> {
+        if !self.dpe.support.simulation() && context.context_type == ContextType::Simulation {
+            return Err(ValidationError::SimulationNotSupported);
+        }
+        if !self.dpe.support.internal_dice() && context.uses_internal_input_dice() {
+            return Err(ValidationError::InternalDiceNotSupported);
+        }
+        if !self.dpe.support.internal_info() && context.uses_internal_input_info() {
+            return Err(ValidationError::InternalInfoNotSupported);
+        }
+        // initialized contexts will always have parent = Context::ROOT_INDEX and their allow_ca and allow_x509
+        // fields will always be true regardless of support
+        if context.parent_idx != Context::ROOT_INDEX {
+            if !self.dpe.support.csr() && context.allow_ca() {
+                return Err(ValidationError::AllowCaNotSupported);
+            }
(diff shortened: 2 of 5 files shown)


sree-revoori1:
Tests in progress. Please suggest other validations to check if you have any ideas.

sree-revoori1:
Tests are done and PR is ready for review.

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+            let context = &self.dpe.contexts[i];
+
+            self.check_support(context)?;
+
+            match context.state {
+                ContextState::Inactive => {
+                    self.validate_inactive_context(context)?;
+                }
+                ContextState::Active => {
+                    self.check_children_and_parent(i)?;
```
zhalvorsen, line 74: We should check that if there is an active non-simulation context `self.dpe.has_initialized` is true, otherwise we could have a second initialized tree by accident.

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+
+    /// Checks if there are multiple active default contexts or a mix of default
+    /// and non-default contexts within the same locality.
+    fn check_context_handles_per_locality(&self) -> Result<(), ValidationError> {
+        for i in 0..MAX_HANDLES {
+            let mut default_count = 0;
+            let mut non_default_count = 0;
+            let locality = self.dpe.contexts[i].locality;
+            for j in 0..MAX_HANDLES {
+                let context = &self.dpe.contexts[j];
```
zhalvorsen, line 171: nit:

```suggestion
        for locality in self.dpe.contexts.iter().map(|context| context.locality) {
            let mut default_count = 0;
            let mut non_default_count = 0;
            for context in self.dpe.contexts.iter() {
```

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    /// Determines if the context array represents a valid collection of disjoint
+    /// directed connnected acyclic graphs (forest) using depth-first search.
+    fn validate_context_forest(&self) -> Result<(), ValidationError> {
+        let mut seen = [false; MAX_HANDLES];
+        let mut in_degree = [0; MAX_HANDLES];
+
+        // count in degree of each node
+        for i in 0..MAX_HANDLES {
```
zhalvorsen, line 198: nit:

```suggestion
        for context in self.dpe.contexts.iter() {
```

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+            for child in flags_iter(self.dpe.contexts[i].children, MAX_HANDLES) {
+                if child >= MAX_HANDLES {
+                    return Err(ValidationError::ChildDoesNotExist);
+                }
+                in_degree[child] += 1;
+            }
+        }
+
+        for node_in_degree in in_degree.iter().take(MAX_HANDLES) {
```
zhalvorsen, line 207: nit:
```suggestion
        for node_in_degree in in_degree {
```

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+            // all nodes must have only one parent
+            if *node_in_degree > 1 {
+                return Err(ValidationError::ChildWithMultipleParents);
+            }
+        }
+
+        let mut normal_tree_count = 0;
+        for (i, node_in_degree) in in_degree.iter().enumerate().take(MAX_HANDLES) {
```
zhalvorsen, line 215: nit:
```suggestion
        for (i, (context, node_in_degree)) in self.dpe.contexts.iter().zip(in_degree).enumerate() {
```

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+                self.detect_invalid_subtree(i, &mut seen, context_type)?;
+            }
+        }
+        // there can be at most one tree of contexts with ContextType::Normal
+        if normal_tree_count > 1 {
+            return Err(ValidationError::MultipleNormalConnectedComponents);
+        }
+
+        // if any node is undiscovered the graph must have a simple cycle
+        for (i, node_visited) in seen.iter().enumerate().take(MAX_HANDLES) {
```
zhalvorsen, line 231: nit:
```suggestion
        for (context, node_visited) in self.dpe.contexts.iter().zip(seen) {
```

dpe/src/validation.rs
```diff
@@ -0,0 +1,306 @@
...
+use crate::{
+    context::{Context, ContextState, ContextType},
+    dpe_instance::flags_iter,
+    response::DpeErrorCode,
+    tci::TciNodeData,
+    DpeInstance, MAX_HANDLES,
+};
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[repr(u32)]
```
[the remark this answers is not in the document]
sree-revoori1: Done

dpe/src/validation.rs
```diff
@@ -0,0 +1,306 @@
...
+                    || context.uses_internal_input_info()
+                {
+                    return Err(ValidationError::InactiveContextWithFlagSet);
+                }
+            } else if context.state == ContextState::Active {
+                self.check_children_and_parent(i)?;
+            } else if context.state == ContextState::Retired {
+                self.check_children_and_parent(i)?;
+                // retired contexts must have exactly one child context
+                if flags_iter(context.children, MAX_HANDLES).count() == 0 {
```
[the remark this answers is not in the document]
sree-revoori1: Retired contexts can have more than 1 child under the following sequence of calls - https://github.com/chipsalliance/caliptra-dpe/pull/285#discussion_r1422823876

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+            .map_err(DpeErrorCode::Validation)?;
+        self.validate_context_forest()
+            .map_err(DpeErrorCode::Validation)
+    }
+
+    /// Returns an error if there is any illegal state or inconsistencies
+    /// present within the DPE instance.
+    fn validate_dpe_state(&self) -> Result<(), ValidationError> {
+        for i in 0..MAX_HANDLES {
+            let context = &self.dpe.contexts[i];
```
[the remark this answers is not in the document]
sree-revoori1: Done

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+        }
+        // Check if any children do not exist
+        for child in flags_iter(context.children, 32) {
+            if child >= MAX_HANDLES {
+                return Err(ValidationError::ChildDoesNotExist);
+            }
+            // Check that each child's parent is idx
+            if self.dpe.contexts[child].parent_idx as usize != idx {
+                return Err(ValidationError::ParentChildLinksCorrupted);
+            }
```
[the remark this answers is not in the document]
sree-revoori1: This function is only called for active or retired contexts so I think its uneccessary

dpe/src/validation.rs
sree-revoori1: Done

dpe/src/validation.rs
sree-revoori1: Done

dpe/src/validation.rs
sree-revoori1: Done

dpe/src/validation.rs
sree-revoori1: Done

dpe/src/validation.rs
sree-revoori1: Done

dpe/src/validation.rs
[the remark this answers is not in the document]
sree-revoori1: oh yeah didn't even notice. Done.

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+    fn check_support(&self, context: &Context) -> Result<(), ValidationError> {
+        if !self.dpe.support.simulation() && context.context_type == ContextType::Simulation {
+            return Err(ValidationError::SimulationNotSupported);
+        }
+        if !self.dpe.support.internal_dice() && context.uses_internal_input_dice() {
+            return Err(ValidationError::InternalDiceNotSupported);
+        }
+        if !self.dpe.support.internal_info() && context.uses_internal_input_info() {
+            return Err(ValidationError::InternalInfoNotSupported);
+        }
```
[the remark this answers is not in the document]
sree-revoori1: Thanks for finding, I added the first 2 but supporting csr isn't related to allow_x509 field in Context so didn't add.

dpe/src/validation.rs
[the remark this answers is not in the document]
sree-revoori1: Done, misunderstood that you were talking about the child context here.

dpe/src/validation.rs
```diff
@@ -0,0 +1,505 @@
...
+    ChildWithMultipleParents = 0x12,
+    ParentChildLinksCorrupted = 0x13,
+}
+
+impl ValidationError {
+    pub fn discriminant(&self) -> u16 {
+        // SAFETY: Because `Self` is marked `repr(u16)`, its layout is a `repr(C)` `union`
+        // between `repr(C)` structs, each of which has the `u16` discriminant as its first
+        // field, so we can read the discriminant without offsetting the pointer.
+        unsafe { *<*const _>::from(self).cast::<u16>() }
```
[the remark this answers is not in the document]
sree-revoori1: Done

dpe/src/context.rs
```diff
@@ -96,6 +96,8 @@ impl Context {
         self.state = ContextState::Inactive;
         self.uses_internal_input_info = false.into();
         self.uses_internal_input_dice = false.into();
+        self.allow_ca = false.into();
+        self.allow_x509 = false.into();
```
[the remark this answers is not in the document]
sree-revoori1: We could just do *self = Context::new(); but a lot of the fields are kind of unecessary to set which is why I think we followed this approach in the first place.

dpe/src/validation.rs
sree-revoori1: Now checking that if there is an active, normal ctx, has_initialized must be set to true.

dpe/src/validation.rs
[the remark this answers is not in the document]
sree-revoori1: Actually keeping this as is, since otherwise it breaks the caliptra panic test since the compiler can't infer that i is < MAX_HANDLES. We could add an if-statement to protect against this, but there isn't really a good error code for this.

zhalvorsen (approved):

jhand2 (approved):

PR #292, opened 2023-12-19
Allow platform to choose Signer Identifier and Cert validity

### LINKED ISSUE #280: Fix IssuerAndSerialNumber in CMS Content message
From @jhand2: "I looked more into PKCS7 and I do think we are using IssuerAndSerialNumber wrong. The purpose of that field is to identify who signed the CMS message. They way IssuerAndSerialNumber identifies that signer is by (1) who issuer the signer's certificate and (2) what serial number that issuer have to the signer. Because SerialNumbers are supposed to be unique under a given CA.
This is a bit annoying, because now not only does DPE need to know the issuer's name, but it also needs to know the issuer's issuer.

The alternative is we could also use SubjectKeyIdentifier. But this does have the additional constraint that we would be requiring the signer to use the SubjectKeyIdentifier extension in their cert.
"

### LINKED ISSUE #291: Let platform choose cert validity
The platform should be able to provide the validto/validfrom fields in the DPE leaf cert

sree-revoori1:
fixes #291 
fixes #280

### DIFF
--- a/platform/src/default.rs
+++ b/platform/src/default.rs
@@ -1,6 +1,7 @@
 // Licensed under the Apache-2.0 license
 
-use crate::{Platform, PlatformError, MAX_CHUNK_SIZE, MAX_SN_SIZE};
+use crate::{CertValidity, Platform, PlatformError, SignerIdentifier, MAX_CHUNK_SIZE};
+use arrayvec::ArrayVec;
 use cfg_if::cfg_if;
 use core::cmp::min;
 
@@ -9,6 +10,8 @@ pub struct DefaultPlatform;
 pub const AUTO_INIT_LOCALITY: u32 = 0;
 pub const VENDOR_ID: u32 = 0;
 pub const VENDOR_SKU: u32 = 0;
+pub const NOT_BEFORE: &str = "20230227000000Z";
+pub const NOT_AFTER: &str = "99991231235959Z";
 
 // Run ./generate.sh to generate all test certs and test private keys
 #[cfg(feature = "dpe_profile_p256_sha256")]
@@ -33,7 +36,7 @@ cfg_if! {
                 pub fn parse_issuer_name() -> Vec<u8> {
                     X509::from_pem(TEST_CERT_PEM)
                         .unwrap()
-                        .subject_name()
+                        .issuer_name()
                         .to_der()
                         .unwrap()
                 }
@@ -60,7 +63,7 @@ cfg_if! {
                     Certificate::from_pem(TEST_CERT_PEM)
                         .unwrap()
                         .tbs_certificate
-                        .subject
+                        .issuer
                         .to_der()
                         .unwrap()
                 }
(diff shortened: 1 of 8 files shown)


platform/src/default.rs
```diff
@@ -134,4 +143,10 @@ impl Platform for DefaultPlatform {
         print!("{str}");
         Ok(())
     }
+
+    fn get_cert_validity(&mut self) -> Result<(&str, &str), PlatformError> {
+        let not_before = "20230227000000Z";
+        let not_after = "99991231235959Z";
```
jhand2, line 149: nit: Would be more clear to make these constants at the top of the Platform impl

platform/src/lib.rs
```diff
@@ -65,11 +73,15 @@ pub trait Platform {
...
     fn get_issuer_name(&mut self, out: &mut [u8; MAX_CHUNK_SIZE]) -> Result<usize, PlatformError>;
 
-    /// Retrives the issuer's Serial Number
+    /// Retrieves a CMS Content Info's signer identifier
     ///
-    /// The issuer serial number is a big-endian integer which is at-most 20
-    /// bytes. It must adhere to all the requirements of an ASN.1 integer.
-    fn get_issuer_sn(&mut self, out: &mut [u8; MAX_SN_SIZE]) -> Result<usize, PlatformError>;
+    /// This function can simply return an error if the DPE does not support CSRs
+    fn get_signer_identifier<'a>(
```
jhand2, line 79: nit: Can you explain in the comment that the platform can choose either SubjectKeyIdentifier or IssuerAndSerialNumber

platform/src/lib.rs
```diff
@@ -16,6 +16,14 @@ pub mod printer;
...
 pub const MAX_SN_SIZE: usize = 20;
 
+pub enum SignerIdentifier<'a> {
+    IssuerAndSerialNumber {
+        issuer_name: &'a [u8],
+        serial_number: &'a [u8],
+    },
+    SubjectKeyIdentifier(&'a [u8]),
+}
+
```
jhand2, line 26: If you make all the values here ArrayVecs, you can simplify the `get_signer_identifier` function signature a bit. Either way, I think the values will get copied into fixed-size temporary stack buffers.

platform/src/default.rs
```diff
@@ -109,13 +109,22 @@ impl Platform for DefaultPlatform {
...
     }
 
-    fn get_issuer_sn(&mut self, out: &mut [u8; MAX_SN_SIZE]) -> Result<usize, PlatformError> {
+    fn get_signer_identifier<'a>(
+        &'a mut self,
+        issuer_name: &'a mut [u8; MAX_CHUNK_SIZE],
+        serial_number: &'a mut [u8; MAX_SN_SIZE],
+        _subject_key_identifier: &'a mut [u8; MAX_CHUNK_SIZE],
+    ) -> Result<SignerIdentifier, PlatformError> {
+        let issuer_len = self.get_issuer_name(issuer_name)?;
```
jhand2, line 118: Take the cert chain `Root CA -> Intermediate -> Wrapped CSR`. For IssuerAndSerialNumber,

1. Issuer Name should be the the issuer name from `Intermediate`. So the `CA` subject name.
2. Serial number should be `Intermdiate` serial number (which is correct here).

(1) is only correct here because `Root` and `Intermediate` are the same cert in this case (self-signed). For clarity we should probably pull the Issuer Name from the signer (whereas in `get_issuer_name` we get `subject_name` from the issuer.

platform/src/lib.rs
```diff
@@ -78,4 +90,12 @@ pub trait Platform {
...
 
     fn write_str(&mut self, str: &str) -> Result<(), PlatformError>;
+
+    /// Retrieves the DPE certificate's validity period
+    ///
+    /// Each output string should represent a valid ISO 8601 date and time
+    /// in the yyyyMMddHHmmss format followed by a timezone.
+    ///
+    /// Example: 99991231235959Z is December 31st, 9999 23:59:59 UTC
+    fn get_cert_validity(&mut self) -> Result<(&str, &str), PlatformError>;
```
jhand2, line 100: nit: It might be fore clear to return a struct

```rust
struct CertValidity<'a> {
  valid_from: &'a str,
  valid_to: &'a str,
};
```

platform/src/lib.rs
sree-revoori1: Done

platform/src/default.rs
sree-revoori1: Done

platform/src/lib.rs
sree-revoori1: Done

platform/src/lib.rs
sree-revoori1: Done

platform/src/default.rs

jhand2 (commented):
Looks good! Does the verification test succeed now if you remove this? 
https://github.com/chipsalliance/caliptra-dpe/blob/2e4110b0efcd39d08afd8536ee956b15ca004021/verification/testing/certifyKey.go#L173-L178

sree-revoori1:
> Looks good! Does the verification test succeed now if you remove this?
> 
> https://github.com/chipsalliance/caliptra-dpe/blob/2e4110b0efcd39d08afd8536ee956b15ca004021/verification/testing/certifyKey.go#L173-L178

Yes, it passes now.

jhand2 (approved):

PR #293, opened 2023-12-19
Spdm support

hpya93:
Adds support to SPDM E2E testing.
The Verifier (Go test) requests data from SPDM-requester which in turn requests from SPDM responder.
SPDM responder gets the data from Caliptra DPE.


clundin25:
This is quite stale, closing to clear the PR queue

PR #303, opened 2024-01-22
Make MAX_HANDLES configurable at runtime

### LINKED ISSUE #298: Make DPE features compile-time disableable
The current DPE code is runtime configurable via the `Support` struct and allows enabling/disabling certain features. Be should also add an option to disable features at compile time (cargo feature flags). This will make the binary smaller for an implementation which doesn't use all the profile-defined features.

### LINKED ISSUE #299: Make MAX_TCI_NODES configurable
Today the caliptra-dpe code hard-codes MAX_TCI_NODES to 24. We should make this compile-time configurable so it is possible to reduce the memory usage of Caliptra for lower-resource systems.

sree-revoori1:
To build DPE with a non-default number of MAX_HANDLES, you can run ARBITRARY_MAX_HANDLES=10 cargo build --features arbitrary_max_handles. If the ARBITRARY_MAX_HANDLES env var is not provided or is badly formatted, the default value of 24 will be used.

Some of the tests will fail for small numbers of max handles, but this is expected since the tests are intended for a DPE with 24 handles. In the future, we should add tests for DPEs where MAX_HANDLES=1, 2, etc.

Also, make support features disableable at compile time. Both of these changes will allow us to reduce the size of the DPE binary. 

fixes #299 
fixes #298

### DIFF
--- a/dpe/Cargo.toml
+++ b/dpe/Cargo.toml
@@ -9,6 +9,18 @@ edition = "2021"
 default = ["dpe_profile_p256_sha256"]
 dpe_profile_p256_sha256 = ["platform/dpe_profile_p256_sha256"]
 dpe_profile_p384_sha384 = ["platform/dpe_profile_p384_sha384"]
+# Run ARBITRARY_MAX_HANDLES=n cargo build --features arbitrary_max_handles to use this feature
+arbitrary_max_handles = []
+disable_simulation = []
+disable_extend_tci = []
+disable_auto_init = []
+disable_rotate_context = []
+disable_x509 = []
+disable_csr = []
+disable_is_symmetric = []
+disable_internal_info = []
+disable_internal_dice = []
+disable_is_ca = []
 
 [dependencies]
 bitflags = "2.4.0"
@@ -28,3 +40,4 @@ platform = {path = "../platform", default-features = false, features = ["openssl
 cms = "0.2.2"
 der = "0.7.8"
 spki = "0.7.2"
+rand = "0.8.5"
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -257,22 +257,24 @@ impl<'a> Iterator for ChildToRootIter<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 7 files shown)


dpe/Cargo.toml
```diff
@@ -9,6 +9,17 @@ edition = "2021"
 default = ["dpe_profile_p256_sha256"]
 dpe_profile_p256_sha256 = ["platform/dpe_profile_p256_sha256"]
 dpe_profile_p384_sha384 = ["platform/dpe_profile_p384_sha384"]
+arbitrary_max_handles = []
```
jhand2, line 12: nit: can you add a comment here explaining how to use this? (Basically just what you put in the PR description)

dpe/src/context.rs
```diff
@@ -372,32 +374,13 @@ mod tests {
...
-        19,
-        20,
-        21,
-        22,
-    ];
+    /// So `dpe.contexts[chain_indices[0]]` is the parent of `dpe.contexts[chain_indices[1]]` which is the parent of
+    /// `dpe.contexts[chain_indices[2]]` etc. The chain_indices vector is a random permutation of numbers in [0, MAX_HANDLES).
+    fn get_chain_indices() -> Vec<usize> {
+        let mut chain_indices: Vec<usize> = (0..MAX_HANDLES).collect();
+        let mut rng = rand::thread_rng();
```
jhand2, line 381: nit: I think it would be better us use https://docs.rs/rand/0.8.5/rand/trait.SeedableRng.html instead. And then choose a constant as the seed. This will make the tests deterministic. It looks like StdRng in this crate implements `from_seed`: https://docs.rs/rand/0.8.5/rand/rngs/struct.StdRng.html

dpe/Cargo.toml
sree-revoori1: Done

dpe/src/context.rs
sree-revoori1: Done

jhand2 (approved):

PR #304, opened 2024-01-24
Remove ExtendTci and rename DeriveChild to DeriveContext

sree-revoori1:


dpe/src/commands/derive_context.rs
```diff
@@ -202,6 +211,18 @@ impl CommandExecution for DeriveChildCmd {
...
+        if self.is_recursive() {
+            allow_ca = dpe.contexts[parent_idx].allow_ca();
+            allow_x509 = dpe.contexts[parent_idx].allow_x509();
+            uses_internal_input_info = dpe.contexts[parent_idx].uses_internal_input_info();
+            uses_internal_input_dice = dpe.contexts[parent_idx].uses_internal_input_dice();
+        }
+
         // Create a temporary context to mutate so that we avoid mutating internal state upon an error.
         let mut tmp_child_context = Context::new();
         tmp_child_context.activate(&ActiveContextArgs {
```
jhand2, line 228: I think if `is_recursive`, we should be basically doing the `extend_tci` behavior (update parent rather than making a new child context)?

dpe/src/commands/derive_context.rs
```diff
@@ -392,11 +473,11 @@ mod tests {
...
         let mut dpe = DpeInstance::new(&mut env, Support::AUTO_INIT).unwrap();
 
-        DeriveChildCmd {
+        DeriveContextCmd {
             handle: ContextHandle::default(),
             data: [0; DPE_PROFILE.get_tci_size()],
-            flags: DeriveChildFlags::MAKE_DEFAULT | DeriveChildFlags::CHANGE_LOCALITY,
-            tci_type: 7,
+            flags: DeriveContextFlags::MAKE_DEFAULT | DeriveContextFlags::CHANGE_LOCALITY,
+            tci_type: 0,
```
jhand2, line 480: nit: not a big deal, but why change this?

dpe/src/commands/derive_context.rs
```diff
@@ -704,30 +798,87 @@ mod tests {
...
+        .unwrap();
+
+        // Make sure the parent handle is non-sense when not retaining.
+        assert_eq!(
+            Ok(Response::DeriveContext(DeriveContextResp {
+                handle: ContextHandle::default(),
+                parent_handle: ContextHandle([0xff; ContextHandle::SIZE]),
+                resp_hdr: ResponseHdr::new(DpeErrorCode::NoError),
+            })),
+            DeriveContextCmd {
```
jhand2, line 855: I think it would be worth calling twice in the test. In the current implementation, I think every recursive call will blow away the previous measurement.

verification/testing/extendTCI.go
jhand2, line 1: Should we convert these to be `DeriveContext(recursive=true)` tests rather than deleting them?

dpe/src/commands/derive_context.rs
sree-revoori1: It's because I added the check `dpe.contexts[parent_idx].tci.tci_type != self.tci_type`. I see now that this is only for the recursive case so changing it back.

dpe/src/commands/derive_context.rs
sree-revoori1: Done.

dpe/src/commands/derive_context.rs
sree-revoori1: Done

verification/testing/extendTCI.go
jhand2: Sounds good. Thanks!

jhand2 (approved):

PR #306, opened 2024-01-26
Support different command IDs for different profile versions

jhand2:
Older versions of this profile have different command IDs. Support using slightly different command IDs based on profile version.

### DIFF
--- a/verification/client/abi.go
+++ b/verification/client/abi.go
@@ -16,7 +16,7 @@ const (
 	RespMagic uint32 = 0x44504552
 
 	CurrentProfileMajorVersion uint16 = 0
-	CurrentProfileMinorVersion uint16 = 8
+	CurrentProfileMinorVersion uint16 = 9
 )
 
 // CommandCode is a DPE command code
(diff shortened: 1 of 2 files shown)


verification/client/abi.go
```diff
@@ -37,17 +37,67 @@ type Support struct {
...
+type profileCommandCodes struct {
+	GetProfile          CommandCode
+	InitializeContext   CommandCode
+	DeriveContext       CommandCode
+	CertifyKey          CommandCode
+	Sign                CommandCode
+	RotateContextHandle CommandCode
+	DestroyContext      CommandCode
+	GetCertificateChain CommandCode
+	ExtendTCI           CommandCode
```
sree-revoori1, line 51: We don't need ExtendTci anymore right?

verification/client/abi.go
```diff
@@ -37,17 +37,67 @@ type Support struct {
...
+}
+
+// profileInfo holds constants defined in a specific version of the DPE iRoT
+// profile.
+type profileInfo struct {
```
sree-revoori1, line 56: Should we also include max_tci_nodes in here? It could be helpful for future PRs where we support a non-default number of max_tci_nodes.

verification/client/abi.go
```diff
@@ -37,17 +37,67 @@ type Support struct {
...
+			RotateContextHandle: 0xe,
+			DestroyContext:      0xf,
+			GetCertificateChain: 0x10,
+		},
+		MajorVersion: 0,
+		MinorVersion: 9,
+	}
+}
+
+func getProfileInfoLatest() profileInfo {
```
sree-revoori1, line 98: Maybe add a comment here saying that this function must be updated whenever a new version of the profile is released

verification/client/abi.go
```diff
@@ -333,7 +394,9 @@ func getProfile(t Transport) (*GetProfileResp, error) {
 		Flags        uint32
 	}{}
 
-	respHdr, err := execCommand(t, CommandGetProfile, 0, cmd, &respStruct)
+	// GetProfile command code is 1 in all revisions of the spec
+	getProfile := CommandCode(0x1)
```
sree-revoori1, line 398: Even though it is 0x1 in all versions of the spec, shouldn't we do c.constants.Codes.GetProfile for clarity?

jhand2: The problem is that this function is used to get to profile and then figure out which type of DPEABI to construct. So until we call this command we don't know which profileInfo to use.

sree-revoori1: Ah right that makes sense.

verification/client/abi.go
jhand2: Max tci nodes is a field returned by GetProfile, so I don't think we need to hard-code that. We can read it from the profile response.

verification/client/abi.go
jhand2: I just removed this function. I ended up choosing which profileInfo to use based on the version returned by GetProfile.

verification/client/abi.go
jhand2: Done

sree-revoori1 (approved):

PR #319, opened 2024-02-22
Update zerocopy to 0.6.6

jhand2:


dpe/Cargo.toml
```diff
@@ -32,7 +32,7 @@ constant_time_eq = "0.3.0"
 crypto = {path = "../crypto", default-features = false}
 platform = {path = "../platform", default-features = false}
 ufmt = { git = "https://github.com/korran/ufmt.git", rev = "1d0743c1ffffc68bc05ca8eeb81c166192863f33", features = ["inline"] }
-zerocopy = "0.6.1"
+zerocopy = "0.6.6"
```
sree-revoori1, line 35: Should we pull this out to the package level Cargo.toml as a workspace dependency? Then we would only have to bump the version in one place.

jhand2: Done

sree-revoori1 (approved):

PR #323, opened 2024-02-27
Fix some cert lint errors

sree-revoori1:
1. Truncate SerialNumber to 64 bits
2. Count unused bits in KeyUsage

The remaining lint errors related to key identifiers will be handled in a future PR.

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -109,10 +109,12 @@ impl CommandExecution for CertifyKeyCmd {
         let mut subj_serial = [0u8; DPE_PROFILE.get_hash_size() * 2];
         env.crypto
             .get_pubkey_serial(DPE_PROFILE.alg_len(), &pub_key, &mut subj_serial)?;
+        // The serial number of the subject can be at most 64 bytes
+        let truncated_subj_serial = &subj_serial[..64];
 
         let subject_name = Name {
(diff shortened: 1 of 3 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -109,10 +109,12 @@ impl CommandExecution for CertifyKeyCmd {
         let mut subj_serial = [0u8; DPE_PROFILE.get_hash_size() * 2];
         env.crypto
             .get_pubkey_serial(DPE_PROFILE.alg_len(), &pub_key, &mut subj_serial)?;
+        // The serial number of the subject can be at most 64 bytes
+        let truncated_subj_serial = &subj_serial[..64];
```
jhand2, line 113: When we go to enable these tests in caliptra-sw, I think all the other certs will fail this check: https://github.com/chipsalliance/caliptra-sw/blob/b305b01e4cfd843e7f48151726f813b8b5d3e91e/x509/build/csr.rs#L105-L107

I wonder if we should just explicitly disable this check in the linter since the ROM cert chain won't comply.

We could open an issue in caliptra-sw to potentially fix for a 1.x release

sree-revoori1: Added this check to the exclusion list.

sree-revoori1: Opened https://github.com/chipsalliance/caliptra-sw/issues/1372 to track

PR #331, opened 2024-03-12
Update DPE to latest CFI revision

sree-revoori1:


verification/testing/certifyKey.go
```diff
@@ -495,10 +495,6 @@ func checkCertificateStructure(t *testing.T, certBytes []byte) *x509.Certificate
 			// strictly worse and mixing the two formats does not lend itself well
 			// to fixed-sized X.509 templating.
 			"e_wrong_time_format_pre2050",
-			// Certs in the Caliptra cert chain fail this lint currently.
-			// We will need to truncate the serial numbers for those certs and
-			// then enable this lint.
-			"e_subject_dn_serial_number_max_length",
```
jhand2, line 501: Does this need to stay? I don't think the Caliptra certs comply with this still (also it's unrelated to the rest of this change)

sree-revoori1: No, I found that in caliptra we are encoding the serial number of the subject name as the hex encoding of the sha256 hash of the subject key so it will always be 64 bytes. I agree its not related so if youd like I can move it to a new PR

jhand2: Got it. It's fine to leave it as long as it works in caliptra-sw.

jhand2 (approved):

PR #333, opened 2024-03-13
Derive ZeroizeOnDrop everywhere

sree-revoori1:
To do this, we have to underive Copy on all types deriving ZeroizeOnDrop since a type can't be copyable and droppable.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -22,7 +22,7 @@ use constant_time_eq::constant_time_eq;
 use crypto::{Crypto, Digest, Hasher};
 use platform::{Platform, MAX_CHUNK_SIZE};
 use zerocopy::{AsBytes, FromBytes};
-use zeroize::Zeroize;
+use zeroize::ZeroizeOnDrop;
 
 pub trait DpeTypes {
     type Crypto<'a>: Crypto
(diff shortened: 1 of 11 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -39,7 +39,7 @@ pub struct DpeEnv<'a, T: DpeTypes + 'a> {
 }
 
 #[repr(C, align(4))]
-#[derive(AsBytes, FromBytes, Zeroize)]
+#[derive(AsBytes, FromBytes, ZeroizeOnDrop)]
```
jhand2, line 42: I want to be careful with DpeInstance since it in DCCM. Really it should never get dropped since we only ever have borrows of it.

dpe/src/tci.rs
```diff
@@ -1,10 +1,10 @@
 // Licensed under the Apache-2.0 license.
 use crate::DPE_PROFILE;
 use zerocopy::{AsBytes, FromBytes};
-use zeroize::Zeroize;
+use zeroize::ZeroizeOnDrop;
 
 #[repr(C, align(4))]
-#[derive(Default, Copy, Clone, AsBytes, FromBytes, PartialEq, Eq, Zeroize)]
+#[derive(Default, Clone, AsBytes, FromBytes, PartialEq, Eq, ZeroizeOnDrop)]
 pub struct TciNodeData {
```
jhand2, line 8: I don't think there's anything sensitive in this type. I think the most important type to ZeroizeOnDrop is CryptoBuf, which I believe we already do.

dpe/src/dpe_instance.rs
sree-revoori1: Hmm yeah, I guess we don't really need this PR then. Abandoning.

PR #334, opened 2024-03-15
Fix SKI computation in verification tests

sree-revoori1:
Currently, we are computing the subject key identifier as the hash of 0x04 || pubkey.x || pubkey.y and checking that it matches what's in the cert in the go verification tests. The pubkey is also pulled from the cert and we get its byte representation through big.int::Bytes.

For our purposes, this function is error-prone as it returns the absolute value of the pubkey coordinate. If the coordinate's first bit is 1 (i.e. the coordinate is negative), we would get the wrong byte representation and our hash would be incorrect. So instead of pulling the pubkey from the cert, in this PR we just pull it from the certifyKeyResponse since it's guaranteed to always have the correct pubkey representation. This should fix this flaky test failures in Caliptra.

### DIFF
--- a/verification/testing/certifyKey.go
+++ b/verification/testing/certifyKey.go
@@ -197,7 +197,7 @@ func TestCertifyKeyCsr(d client.TestDPEInstance, c client.DPEClient, t *testing.
 	}
 
 	// Check fields and extensions in the CSR
-	checkCertifyKeyExtensions(t, csr.Extensions, flags, label, csr.PublicKey, false, certChain[len(certChain)-1].SubjectKeyId)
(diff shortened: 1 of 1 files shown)


jhand2:
Just to clarify the diagnosis, I think the issue is really just that big.Int will remove any leading zeros when writing to a byte slice. To fix this, the general solution is to use [FillBytes()](https://pkg.go.dev/math/big#Int.FillBytes). This will ensure any padding zeros are kept.

jhand2 (approved):

jhand2:
> Just to clarify the diagnosis, I think the issue is really just that big.Int will remove any leading zeros when writing to a byte slice. To fix this, the general solution is to use [FillBytes()](https://pkg.go.dev/math/big#Int.FillBytes). This will ensure any padding zeros are kept.

But anyway, this PR does fix the problem as well.

PR #335, opened 2024-03-20
Encode SubjectAlternativeName extension

sree-revoori1:
Currently, only the otherName choice is supported.

### DIFF
--- a/platform/src/default.rs
+++ b/platform/src/default.rs
@@ -1,8 +1,8 @@
 // Licensed under the Apache-2.0 license
 
 use crate::{
-    CertValidity, Platform, PlatformError, SignerIdentifier, MAX_CHUNK_SIZE, MAX_ISSUER_NAME_SIZE,
-    MAX_KEY_IDENTIFIER_SIZE,
+    CertValidity, Platform, PlatformError, SignerIdentifier, SubjectAltName, MAX_CHUNK_SIZE,
+    MAX_ISSUER_NAME_SIZE, MAX_KEY_IDENTIFIER_SIZE,
 };
 use arrayvec::ArrayVec;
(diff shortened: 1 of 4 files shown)


platform/src/lib.rs
```diff
@@ -125,4 +134,11 @@ pub trait Platform {
     ///
     /// Example: 99991231235959Z is December 31st, 9999 23:59:59 UTC
     fn get_cert_validity(&mut self) -> Result<CertValidity, PlatformError>;
+
+    /// Retrieves the SubjectAlternativeName extension
+    ///
+    /// Currently, only the otherName choice is supported. This function
+    /// can be left unimplemented if the SubjectAlternativeName extension is
+    /// not needed in the DPE leaf cert or CSR.
+    fn get_subject_alternative_name(&mut self) -> Result<OtherName, PlatformError>;
```
jhand2, line 143: nit: Should this return a generic SubjectAltName enum where once of the options is OtherName? Then if we want to support the other types we don't need to change this API.

sree-revoori1: Done

platform/src/default.rs
```diff
@@ -197,4 +197,8 @@ impl Platform for DefaultPlatform {
             not_after: not_after_vec,
         })
     }
+
+    fn get_subject_alternative_name(&mut self) -> Result<SubjectAltName, PlatformError> {
```
jhand2, line 201: Might be good to populate this with something so we have test coverage/cert linting against the go code.

sree-revoori1: Done except I just added it to tests in x509.rs so it is easier to test without using certifyKey.

jhand2 (approved):

PR #341, opened 2024-05-05
Include allow_ca and allow_x509 when computing meas hash

### LINKED ISSUE #339: In CDI derivation, we should include allow_ca and allow_x509
A DPE context chain which allows CA or X.509 at a specific context should have a different DICE key than one which doesn't.

sree-revoori1:
fixes #339

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -453,6 +453,11 @@ impl DpeInstance {
                 uses_internal_input_info || context.uses_internal_input_info();
             uses_internal_input_dice =
                 uses_internal_input_dice || context.uses_internal_input_dice();
+
+            // Add allow ca to hash
+            hasher.update(context.allow_ca().as_bytes())?;
+            // Add allow x509 to hash
+            hasher.update(context.allow_x509().as_bytes())?;
         }
 
         // Add internal input info to hash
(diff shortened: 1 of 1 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -468,6 +474,15 @@ impl DpeInstance {
...
         }
 
+        // Add allow ca to hash
+        if cfi_launder(allow_ca) {
+            hasher.update(true.as_bytes())?;
+        }
+        // Add allow x509 to hash
+        if cfi_launder(allow_x509) {
+            hasher.update(true.as_bytes())?;
+        }
```
jhand2, line 484: Instead of doing this conditionally, you could probably always call

```rust
hasher.update(allow_ca.as_bytes());
hasher.update(allow_x509.as_bytes());
```

dpe/src/dpe_instance.rs
```diff
@@ -446,6 +448,10 @@ impl DpeInstance {
                 uses_internal_input_info || context.uses_internal_input_info();
             uses_internal_input_dice =
                 uses_internal_input_dice || context.uses_internal_input_dice();
+
+            // Check if any context allows ca or x509
+            allow_ca = allow_ca || context.allow_ca();
+            allow_x509 = allow_x509 || context.allow_x509();
```
jhand2, line 454: Another option, we could just always call 

```rust
hasher.update(context.allow_ca().as_bytes());
hasher.update(context.allow_ca().as_bytes());
```

It adds a little time to key derivation, but not a ton. It has the added benefit that the key will change depending on which context in the chain had `allow_{ca/x509}` set.

dpe/src/dpe_instance.rs
sree-revoori1: Done

jhand2 (dismissed):

jhand2 (approved):

PR #342, opened 2024-06-04
Move DeriveContext(recursive) returned handle to NEW_CONTEXT_HANDLE

jhand2:
This aligns with the spec and is more consistent with all other commands.


fdamato (approved):

ekoohest:
Good change; simplified the client side.
Would be even better to align iROT spec doc.

PR #348, opened 2024-08-23
Clean up GetProfile test

mmiranda96:
## What does this PR do?
Clean up the `GetProfile` verification tests. The improvements include:
- Changing failures to Error instead of Fatal: this way we can run all tests and get all failures on a single run.
  - The first check remains as a Fatal error, as the object might be nil.
- Change the text on some of the error messages, as they were displaying "version" instead of the proper field checked.
- Clarify which value was expected and which value was received.


mmiranda96:
@jhand2

jhand2 (approved):

PR #351, opened 2024-09-12
Refactor X.509 encoding library to use the der crate

jhand2:
The der crate-based implementation is significantly less code and follows a more consistent pattern. Since this allows defining types with encoding rules, this will make it much simpler to add structures for ML-DSA structures for Caliptra 2.0.

The der crate is part of RustCrypto and is intended to be embedded/no_std friendly.


clundin25:
This is quite stale, closing to clean up the PR queue

PR #355, opened 2024-09-23
Optimize size of constant_time_eq

jhand2:
The constant_time_eq crate is used for comparing random context handles. Switch to the fixed-size constant_time_eq_16 comparison function, which uses quite a bit less code space.

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -1,6 +1,6 @@
 // Licensed under the Apache-2.0 license.
 use crate::{response::DpeErrorCode, tci::TciNodeData, U8Bool, MAX_HANDLES};
-use constant_time_eq::constant_time_eq;
+use constant_time_eq::constant_time_eq_16;
 use zerocopy::{AsBytes, FromBytes};
 use zeroize::Zeroize;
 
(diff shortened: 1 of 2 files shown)


zhalvorsen (approved):

dpe/src/context.rs
```diff
@@ -113,16 +113,21 @@ pub struct ContextHandle(pub [u8; ContextHandle::SIZE]);
...
     }
 
     /// Whether the handle is the default context handle.
     pub fn is_default(&self) -> bool {
-        constant_time_eq(&self.0, &Self::DEFAULT)
+        self.equals(&Self::DEFAULT)
+    }
+
+    #[inline(never)]
+    pub fn equals(&self, other: &ContextHandle) -> bool {
```
zhalvorsen, line 129: optional: I'm mildly interested to know if implementing the [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html) trait does something to the firmware size. This would have the added benefit of being able to use the `==` operator as well.

jhand2: Will look into this as a follow up. Thanks!

PR #367, opened 2024-12-18
Update the rust toolchain to 1.83

ArthurHeymans:


dpe/src/x509.rs
```diff
@@ -88,23 +88,23 @@ impl CertWriter<'_> {
     const CMS_V1: u64 = 1;
     const CSR_V0: u64 = 0;
 
-    const ECDSA_OID: &[u8] = match DPE_PROFILE {
+    const ECDSA_OID: &'static [u8] = match DPE_PROFILE {
```
zhalvorsen, line 91: Out of curiosity, did 1.83 recommend changing all of these to static references?

ArthurHeymans: it looks like this was not targetting a recent main. Main already had those as static.

jhand2 (approved):

PR #370, opened 2024-12-26
Add is_certificate and export_cdi flags to DeriveContext

clundin25:
As well as other boiler plate work to make sure the invariants laid out in the spec are enforced.

### DIFF
--- a/dpe/Cargo.toml
+++ b/dpe/Cargo.toml
@@ -20,6 +20,7 @@ disable_csr = []
 disable_internal_info = []
 disable_internal_dice = []
 disable_retain_parent_context = []
+disable_export_cdi = []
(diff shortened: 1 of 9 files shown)


dpe/Cargo.toml
```diff
@@ -20,6 +20,7 @@ disable_csr = []
 disable_internal_info = []
 disable_internal_dice = []
 disable_retain_parent_context = []
+disable_export_cdi = []
```
clundin25, line 23: The specification has a separate ABI if `EXPORT_CDI` is set.

Should this compilation flag also change the shape of the `DeriveContextResp` type?

clundin25:
Will follow up with clean up / refactors once most of the feature work is out of the way.

PR #371, opened 2025-01-06
Add is_certificate and export_cdi flags to DeriveContext

clundin25:
As well as other boiler plate work to make sure the invariants laid out in the spec are enforced.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -1,11 +1,11 @@
 // Licensed under the Apache-2.0 license.
 use super::CommandExecution;
 use crate::{
-    context::{ActiveContextArgs, Context, ContextHandle, ContextState},
+    context::{ActiveContextArgs, Context, ContextHandle, ContextState, ContextType},
     dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
     response::{DeriveContextResp, DpeErrorCode, Response, ResponseHdr},
     tci::TciMeasurement,
-    DPE_PROFILE,
+    DPE_PROFILE, MAX_CERT_SIZE, MAX_EXPORTED_CDI_SIZE,
 };
 use bitflags::bitflags;
 #[cfg(not(feature = "no-cfi"))]
@@ -36,6 +36,9 @@ bitflags! {
         const INPUT_ALLOW_CA = 1u32 << 26;
         const INPUT_ALLOW_X509 = 1u32 << 25;
         const RECURSIVE = 1u32 << 24;
+        const EXPORT_CDI = 1u32 << 23;
+        const CREATE_CERTIFICATE = 1u32 << 22;
+        const RETURN_CERTIFICATE = 1u32 << 21;
     }
 }
 
@@ -89,6 +92,18 @@ impl DeriveContextCmd {
 
     pub const fn is_recursive(&self) -> bool {
         self.flags.contains(DeriveContextFlags::RECURSIVE)
+    }
+
+    pub const fn allows_exports_cdi(&self) -> bool {
+        self.flags.contains(DeriveContextFlags::EXPORT_CDI)
+    }
+
+    pub const fn allows_create_certificate(&self) -> bool {
+        self.flags.contains(DeriveContextFlags::CREATE_CERTIFICATE)
+    }
+
+    pub const fn allows_return_certificate(&self) -> bool {
+        self.flags.contains(DeriveContextFlags::RETURN_CERTIFICATE)
     }
 
     /// Whether it is okay to make a default context.
@@ -196,6 +211,8 @@ impl CommandExecution for DeriveContextCmd {
             || (!dpe.support.internal_dice() && self.uses_internal_dice_input())
             || (!dpe.support.retain_parent_context() && self.retains_parent())
             || (!dpe.support.x509() && self.allows_x509())
+            || (!dpe.support.cdi_export()
+                && (self.allows_create_certificate() && self.allows_exports_cdi()))
             || (!dpe.support.recursive() && self.is_recursive())
         {
             return Err(DpeErrorCode::ArgumentNotSupported);
@@ -204,6 +221,11 @@ impl CommandExecution for DeriveContextCmd {
         let parent_idx = dpe.get_active_context_pos(&self.handle, locality)?;
         if (!dpe.contexts[parent_idx].allow_ca() && self.allows_ca())
             || (!dpe.contexts[parent_idx].allow_x509() && self.allows_x509())
+            || (self.allows_exports_cdi() && !self.allows_create_certificate())
+            || (self.allows_exports_cdi() && self.is_recursive())
+            || (self.allows_exports_cdi() && self.changes_locality())
+            || (self.allows_exports_cdi()
+                && dpe.contexts[parent_idx].context_type == ContextType::Simulation)
             || (self.is_recursive() && self.retains_parent())
         {
             return Err(DpeErrorCode::InvalidArgument);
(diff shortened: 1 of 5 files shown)


dpe/src/response.rs
```diff
@@ -139,6 +139,9 @@ pub struct DeriveContextResp {
     pub resp_hdr: ResponseHdr,
     pub handle: ContextHandle,
     pub parent_handle: ContextHandle,
+    pub exported_cdi: [u8; MAX_EXPORTED_CDI_SIZE],
+    pub certificate_size: u32,
+    pub new_certificate: [u8; MAX_CERT_SIZE],
```
clundin25, line 144: @jhand2 These are defined as a separate variant for the response ABI shape.

Do you want to enforce that via build flag or do you want the struct shape to always be the same?

dpe/src/commands/derive_context.rs
```diff
@@ -196,6 +211,8 @@ impl CommandExecution for DeriveContextCmd {
             || (!dpe.support.internal_dice() && self.uses_internal_dice_input())
             || (!dpe.support.retain_parent_context() && self.retains_parent())
             || (!dpe.support.x509() && self.allows_x509())
+            || (!dpe.support.cdi_export()
+                && (self.allows_create_certificate() && self.allows_exports_cdi()))
```
clundin25, line 215: Might also need to check that if `dpe.support.cdi_export()` is true then `dpe.support.x509()` is also true.

dpe/src/commands/derive_context.rs
```diff
@@ -36,6 +36,9 @@ bitflags! {
         const INPUT_ALLOW_CA = 1u32 << 26;
         const INPUT_ALLOW_X509 = 1u32 << 25;
         const RECURSIVE = 1u32 << 24;
+        const EXPORT_CDI = 1u32 << 23;
+        const CREATE_CERTIFICATE = 1u32 << 22;
+        const RETURN_CERTIFICATE = 1u32 << 21;
```
jhand2, line 41: In the spec, I think there is only `CREATE_CERTIFICATE` flag, not `RETURN_CERTIFICATE`. Is there a reason we need both?

dpe/src/commands/derive_context.rs
jhand2: I think technically we could create a CSR with IsCA = true

jhand2, line 215: Should this be `(self.allows_create_certificate() || self.allows_exports_cdi())`?

dpe/src/commands/derive_context.rs
```diff
@@ -262,11 +284,16 @@ impl CommandExecution for DeriveContextCmd {
                         // Should be ignored since retain_parent cannot be true
                         parent_handle: ContextHandle::default(),
                         resp_hdr: ResponseHdr::new(DpeErrorCode::NoError),
+                        exported_cdi: [0; MAX_EXPORTED_CDI_SIZE],
+                        certificate_size: 0,
+                        new_certificate: [0; MAX_CERT_SIZE]
```
jhand2, line 289: Currently in the spec, these fields are only returned if `EXPORT_CDI == true`: https://github.com/TrustedComputingGroup/Server-Internal/blob/main/dpe-irot-profile/main.tcg#L1450-L1471

So I think we would want to have two different response structs and conditionally device which one to return. Although given how the Response enum works, that might be a bit of a pain. So maybe it's better to just cut that part out at a higher level.

wdyt?

dpe/src/commands/derive_context.rs
```diff
@@ -996,4 +1042,105 @@ mod tests {
         let digest = hasher_2.finish().unwrap();
         assert_eq!(digest.bytes(), dpe.contexts[child_idx].tci.tci_cumulative.0);
     }
+
+    #[test]
+    fn test_cdi_export_flags() {
```
jhand2, line 1047: nit: It would be helpful to have a quick comment above each different edge case that's being tested.

dpe/src/response.rs
jhand2: The intention in the spec was that we would return a different one based on the input request flags. I left a related comment about this above.

dpe/src/commands/derive_context.rs
clundin25: Hmm, in that case I'll omit this check, if this invariant does occur it will fail (eventually)

dpe/src/commands/derive_context.rs
clundin25: Looking at "TCG DICE Protection Environment Specification v1.0 FINAL"

It says:

> return-certificate: Indicates whether a DPE MUST return the generated certificate when create-certificate is true. If true, the certificate is returned in the certificate output argument. If create-certificate is false, this argument is ignored. If omitted, the default value of this argument is false.

I think `return-certificate` is redundant personally. Did I look at the wrong spec?

dpe/src/commands/derive_context.rs
clundin25: From the spec:

> If a DPE supports certificates, the DPE MUST require that the create-certificate argument is true and return an invalid-argument error if it is false

That is the argument section for `export-cdi`.

dpe/src/commands/derive_context.rs
clundin25: I need to think deeper on this but maybe we could achieve something similar wrapping these fields with `Option<T>`.

That at least solves the serialization (I think).

dpe/src/commands/derive_context.rs
clundin25: Oh I see, the profile specifies this can be omitted. I think the profile specification will need to be updated to default to `true`.

dpe/src/commands/derive_context.rs
clundin25: In hindsight, I don't think using `Option<T>` will work.

> So maybe it's better to just cut that part out at a higher level.

I agree, a single response will be easier.

dpe/src/commands/derive_context.rs
jhand2: Right, but if `!dpe.support.cdi_export()`, I think we want to fail if _either_ of the flags are present. Right now I think it will only fail if both flags are set.

dpe/src/commands/derive_context.rs
jhand2: Closing since this is now handled with a speparate struct.

dpe/src/commands/derive_context.rs
jhand2: Thanks!

dpe/src/commands/derive_context.rs
```diff
@@ -91,6 +93,14 @@ impl DeriveContextCmd {
         self.flags.contains(DeriveContextFlags::RECURSIVE)
     }
 
+    pub const fn allows_exports_cdi(&self) -> bool {
+        self.flags.contains(DeriveContextFlags::EXPORT_CDI)
+    }
+
+    pub const fn allows_create_certificate(&self) -> bool {
+        self.flags.contains(DeriveContextFlags::CREATE_CERTIFICATE)
+    }
```
jhand2, line 102: nit: previously the function names used "allows" because the flags would allow specific behavior that you then had to use when calling CertifyKey.

But these flags impact the DeriveContext call directly. So I think we should change these function names to be `creates_certificate` and `exports_cdi`.

dpe/src/support.rs
```diff
@@ -161,6 +171,7 @@ pub mod test {
                 | (1 << 22)
                 | (1 << 21)
                 | (1 << 19)
+                | (1 << 18)
```
jhand2, line 174: nit: remove this flag given that `return-certificate` is implicit.

jhand2: oh jk. 18 is cdi export

dpe/src/commands/derive_context.rs
clundin25: Updated!

dpe/src/support.rs
clundin25: Closing this, as I think this is resolved but please re-open if I am mistaken.

dpe/src/commands/derive_context.rs
clundin25: Updated!

PR #373, opened 2025-01-13
Refactor certificate generation code in CertifyKey

clundin25:
Refactor certificate generation code in CertifyKey to common code so it can be used in DeriveContext.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -170,12 +170,38 @@ pub trait Crypto {
         info: &[u8],
     ) -> Result<Self::Cdi, CryptoError>;
 
+    /// Derive a CDI for an exported private key based on the current base CDI and measurements
+    ///
+    /// # Arguments
+    ///
+    /// * `algs` - Which length of algorithms to use.
+    /// * `measurement` - A digest of the measurements which should be used for CDI derivation
+    /// * `info` - Caller-supplied info string to use in CDI derivation
+    fn derive_exported_cdi(
+        &mut self,
+        algs: AlgLen,
+        measurement: &Digest,
+        info: &[u8],
+    ) -> Result<Self::Cdi, CryptoError>;
+
     /// CFI wrapper around derive_cdi
     ///
     /// To implement this function, you need to add the
     /// cfi_impl_fn proc_macro to derive_cdi.
     #[cfg(not(feature = "no-cfi"))]
     fn __cfi_derive_cdi(
+        &mut self,
+        algs: AlgLen,
+        measurement: &Digest,
+        info: &[u8],
+    ) -> Result<Self::Cdi, CryptoError>;
+
+    /// CFI wrapper around derive_cdi_exported
+    ///
+    /// To implement this function, you need to add the
+    /// cfi_impl_fn proc_macro to derive_exported_cdi.
+    #[cfg(not(feature = "no-cfi"))]
+    fn __cfi_derive_exported_cdi(
         &mut self,
         algs: AlgLen,
         measurement: &Digest,
@@ -199,12 +225,42 @@ pub trait Crypto {
         info: &[u8],
     ) -> Result<(Self::PrivKey, EcdsaPub), CryptoError>;
 
+    /// Derives an exported key pair using a cryptographically secure KDF
+    ///
+    /// # Arguments
+    ///
+    /// * `algs` - Which length of algorithms to use.
+    /// * `cdi` - Caller-supplied private key to use in public key derivation
+    /// * `label` - Caller-supplied label to use in asymmetric key derivation
+    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
+    ///
+    fn derive_key_pair_exported(
+        &mut self,
+        algs: AlgLen,
+        cdi: &Self::Cdi,
+        label: &[u8],
+        info: &[u8],
+    ) -> Result<(Self::PrivKey, EcdsaPub), CryptoError>;
+
     /// CFI wrapper around derive_key_pair
     ///
     /// To implement this function, you need to add the
     /// cfi_impl_fn proc_macro to derive_key_pair.
     #[cfg(not(feature = "no-cfi"))]
     fn __cfi_derive_key_pair(
+        &mut self,
+        algs: AlgLen,
+        cdi: &Self::Cdi,
+        label: &[u8],
+        info: &[u8],
+    ) -> Result<(Self::PrivKey, EcdsaPub), CryptoError>;
+
+    /// CFI wrapper around derive_key_pair_exported
+    ///
+    /// To implement this function, you need to add the
+    /// cfi_impl_fn proc_macro to derive_key_pair.
+    #[cfg(not(feature = "no-cfi"))]
+    fn __cfi_derive_key_pair_exported(
         &mut self,
         algs: AlgLen,
         cdi: &Self::Cdi,
--- a/crypto/src/openssl.rs
+++ b/crypto/src/openssl.rs
@@ -90,6 +90,35 @@ impl OpensslCrypto {
 
         EcKey::from_private_components(&group, priv_key_bn, &pub_point)
     }
+
+    fn derive_key_pair_inner(
+        &mut self,
+        algs: AlgLen,
+        cdi: &<OpensslCrypto as Crypto>::Cdi,
+        label: &[u8],
+        info: &[u8],
+    ) -> Result<(<OpensslCrypto as Crypto>::PrivKey, EcdsaPub), CryptoError> {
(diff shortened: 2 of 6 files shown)


crypto/src/lib.rs
```diff
@@ -182,6 +196,28 @@ pub trait Crypto {
...
+    #[cfg(not(feature = "no-cfi"))]
+    fn __cfi_derive_exported_cdi(
+        &mut self,
+        algs: AlgLen,
+        measurement: &Digest,
+        info: &[u8],
+    ) -> Result<Self::Cdi, CryptoError>;
+
+    /// Retrieves the CDI for an exported private key.
+    fn get_exported_cdi(&mut self) -> Result<Self::Cdi, CryptoError>;
```
clundin25, line 212: I'm not sure this will be necessary after moving `SignWithExported` from DPE to "caliptra-sw".

Leaving it in for now.

crypto/src/lib.rs
```diff
@@ -170,6 +170,20 @@ pub trait Crypto {
...
+    ///
+    /// * `algs` - Which length of algorithms to use.
+    /// * `measurement` - A digest of the measurements which should be used for CDI derivation
+    /// * `info` - Caller-supplied info string to use in CDI derivation
+    fn derive_exported_cdi(
+        &mut self,
+        algs: AlgLen,
+        measurement: &Digest,
+        info: &[u8],
+    ) -> Result<Self::Cdi, CryptoError>;
```
jhand2, line 185: nit: the calling code is expecting a raw buffer with some data it can return in the response. Should be return an array type instead? We can still leave it up to the platform to fill it with a random handle (v.s. an actual CDI).

crypto/src/lib.rs
jhand2: Agreed I don't think this is needed

crypto/src/lib.rs
```diff
@@ -199,6 +235,23 @@ pub trait Crypto {
...
+    /// Derives an exported key pair using a cryptographically secure KDF
+    ///
+    /// # Arguments
+    ///
+    /// * `algs` - Which length of algorithms to use.
+    /// * `cdi` - Caller-supplied private key to use in public key derivation
+    /// * `label` - Caller-supplied label to use in asymmetric key derivation
+    /// * `info` - Caller-supplied info string to use in asymmetric key derivation
+    ///
+    fn derive_key_pair_exported(
```
jhand2, line 247: This is probably up to Caliptra FW, so doesn't need to be in the platform layer

crypto/src/openssl.rs
```diff
@@ -42,22 +42,31 @@ impl Hasher for OpensslHasher {
 }
 
 #[cfg(feature = "deterministic_rand")]
-pub struct OpensslCrypto(StdRng);
+pub struct OpensslCrypto {
+    rng: StdRng,
+    // Cached CDI for exported CDI operations.
+    exported_cdi: Option<Vec<u8>>,
```
jhand2, line 48: no need to cache

jhand2: I think for the default platform, we should probably just return an actual CDI. It's just for the Caliptra overload that we'll do this extra handle stuff.

dpe/src/commands/certify_key.rs
```diff
@@ -94,115 +88,24 @@ impl CommandExecution for CertifyKeyCmd {
...
-
-        let mut authority_key_identifier = [0u8; MAX_KEY_IDENTIFIER_SIZE];
-        env.platform
-            .get_issuer_key_identifier(&mut authority_key_identifier)?;
-
-        let subject_alt_name = match env.platform.get_subject_alternative_name() {
-            Ok(subject_alt_name) => Some(subject_alt_name),
-            Err(PlatformError::NotImplemented) => None,
-            Err(e) => Err(DpeErrorCode::Platform(e))?,
+        let args = CreateDpeCertArgs {
```
jhand2, line 91: Nice! This is way cleaner

dpe/src/x509.rs
```diff
@@ -2227,6 +2235,286 @@ impl CertWriter<'_> {
...
+    let truncated_subj_serial = &subj_serial[..64];
+
+    let subject_name = Name {
+        cn: DirectoryString::PrintableString(b"DPE Leaf"),
+        serial: DirectoryString::PrintableString(truncated_subj_serial),
+    };
+    Ok(subject_name)
+}
+
+fn get_tci_nodes(
```
jhand2, line 2290: nit: maybe `get_tci_node_count`

dpe/src/x509.rs
```diff
@@ -2227,6 +2235,286 @@ impl CertWriter<'_> {
...
+    output_cert_or_csr: &mut [u8],
+) -> Result<CreateDpeCertResult, DpeErrorCode> {
+    let algs = DPE_PROFILE.alg_len();
+    let digest = get_dpe_measurement_digest(dpe, env, args.handle, args.locality)?;
+
+    let key_pair = match cert_type {
+        CertificateType::Exported => {
+            let cdi = env
+                .crypto
+                .derive_exported_cdi(algs, &digest, args.cdi_label)?;
```
jhand2, line 2388: nit: I wonder if the key derivation code should stay in the CertifyKey command implementation. Since it's not really related to X.509.

crypto/src/lib.rs
jhand2: Oh never mind. You need the pubkey to make the cert.

crypto/src/lib.rs
jhand2: Alternatively, we could add a function that turns a Self::Cdi into a raw buffer. And leave a comment that the platform can either populate with a raw CDI or a handle to a CDI. Not sure which you would prefer.

crypto/src/lib.rs
clundin25: I've removed this.

crypto/src/openssl.rs
clundin25: I needed it for `get_exported_cdi` but with `SignWithExtended` happening in SW instead, I don't think I will need these cached cdis. 

I've removed them.

dpe/src/x509.rs
clundin25: The `nodes` parameter is also an "out" parameter.

I'll refactor it to instead return a nodes buffer.

dpe/src/x509.rs
clundin25: I agree but I could not think of a way to do it.

Since they are an associated type of `DpeEnv<impl DpeTypes>` I don't think there is a type that we can use in the function signature.

Any ideas?

dpe/src/x509.rs
jhand2: Oh, got it. Well if this saves on stack space then feel free to leave it like this. I guess depends on how smart llvm is.

dpe/src/x509.rs
jhand2: The CDI is an associated type, but the key pair derivation functions return a raw public key that we can just pass into this function, right?

crypto/src/lib.rs
clundin25: I didn't understand, do you mind sharing additional context?

I think it's okay as is because the `CDI` is passed to `derive_key_pair_exported`, the caller shouldn't need to know it's underlying type.

dpe/src/x509.rs
clundin25: I think the private key is still needed for the exported certificate since it's self signed.

This has made me realized that when I refactored the code, I did not self-sign the exported certificate!

dpe/src/x509.rs
clundin25: I'll leave this for now since out params are not so nice.

crypto/src/lib.rs
jhand2: As discussed offline, in a follow-up PR we can have a PR for extracting the CDI buffer to return.

dpe/src/x509.rs
```diff
@@ -2227,6 +2235,302 @@ impl CertWriter<'_> {
...
+    Leaf,
+    Exported,
+}
+
+pub(crate) struct CreateDpeCertArgs<'a> {
+    pub handle: &'a ContextHandle,
+    pub locality: u32,
+    pub cdi_label: &'a [u8],
+    pub key_label: &'a [u8],
+    pub context: &'a [u8],
```
jhand2, line 2253: nit: Can you comment where each of these are used in derivation? Mainly that `context` is used for both (I think)

dpe/src/x509.rs
jhand2: As discussed offlne, the cert should still be signed by the alias key

PR #374, opened 2025-01-15
Exported CDI handle should be 32 bits not 32 bytes.

clundin25:


jhand2 (approved):
LGTM, just looks like `cargo fmt` is mad

clundin25:
@jhand2 thanks!

PR #376, opened 2025-01-15
Add certificate generation code to DeriveContext

clundin25:
This will return a self-signed certificate and a CDI for the key handle.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -3,9 +3,12 @@ use super::CommandExecution;
 use crate::{
     context::{ActiveContextArgs, Context, ContextHandle, ContextState, ContextType},
     dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
-    response::{DeriveContextResp, DpeErrorCode, Response, ResponseHdr},
+    response::{
+        DeriveContextExportedCdiResp, DeriveContextResp, DpeErrorCode, Response, ResponseHdr,
+    },
     tci::TciMeasurement,
-    DPE_PROFILE,
+    x509::{create_exported_dpe_cert, CreateDpeCertArgs, CreateDpeCertResult},
+    DPE_PROFILE, MAX_CERT_SIZE, MAX_EXPORTED_CDI_SIZE,
 };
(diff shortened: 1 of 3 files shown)


dpe/src/commands/derive_context.rs
```diff
@@ -284,7 +288,39 @@ impl CommandExecution for DeriveContextCmd {
                 }
             }
         } else if self.creates_certificate() && self.exports_cdi() {
-            todo!("(clundin): Generate ECA certificiate / export random CDI")
+            cfg_if! {
+                if #[cfg(not(feature = "disable_export_cdi"))] {
+                    let mut exported_cdi_handle = [0; MAX_EXPORTED_CDI_SIZE];
+                    env.crypto
+                        .rand_bytes(&mut exported_cdi_handle)
+                        .map_err(DpeErrorCode::Crypto)?;
```
clundin25, line 296: Once refactored this will happen in the call to `create_exported_dpe_cert` and return in `CreateDpeCertResult`

clundin25:
Will follow up in a separate PR with verification tests update.

dpe/src/commands/derive_context.rs
```diff
@@ -284,7 +288,39 @@ impl CommandExecution for DeriveContextCmd {
...
+                    let CreateDpeCertResult { cert_size, .. } = create_exported_dpe_cert(
+                        &args,
+                        dpe,
+                        env,
+                        &mut cert,
+                    )?;
+
+                    Ok(Response::DeriveContextExportedCdi(DeriveContextExportedCdiResp {
+                        handle: ContextHandle::new_invalid(),
+                        parent_handle: dpe.contexts[parent_idx].handle,
```
jhand2, line 314: If export_cdi, then `retains_parent` must be false. So I think the parent handle should be `new_invalid`. We should see if we can share the logic from line 374 so it doesn't need to be duploicated.

clundin25: Moved the parent invalidation logic further up so it is shared with `export-cdi`.

jhand2 (approved):
Thanks!

PR #379, opened 2025-01-16
Exchange CDI for CDI handle via DPE::Crypto trait

clundin25:


crypto/src/openssl.rs
```diff
@@ -177,6 +186,27 @@ impl Crypto for OpensslCrypto {
         Ok(cdi)
     }
 
+    fn get_exported_cdi_handle(&mut self, cdi: &Self::Cdi) -> Result<[u8; MAX_EXPORTED_CDI_SIZE], CryptoError> {
+        if self.export_cdi_slots.len() >= MAX_CDI_HANDLES {
+            return Err(CryptoError::ExportedCdiHandleLimitExceeded);
+        }
```
clundin25, line 192: TODO: If a slot has already been exported, return the existing handle?

crypto/src/openssl.rs
```diff
@@ -177,6 +194,40 @@ impl Crypto for OpensslCrypto {
...
 
+    fn get_exported_cdi_handle(
+        &mut self,
+        cdi: &Self::Cdi,
+    ) -> Result<ExportedCdiHandle, CryptoError> {
+        for (stored_cdi, stored_handle) in self.export_cdi_slots.iter() {
+            if stored_cdi == cdi {
+                return Ok(stored_handle.to_owned());
+            }
+        }
```
clundin25, line 205: Is this a valid scenario?

crypto/src/lib.rs
```diff
@@ -208,6 +212,31 @@ pub trait Crypto {
...
+    ) -> Result<[u8; MAX_EXPORTED_CDI_SIZE], CryptoError>;
+
+    /// Retrieve a CDI from an exported CDI Handle.
+    ///
+    /// If the handle is not associated with a CDI, `None` is returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `exported_cdi_handle` - The handle to the CDI.
+    fn get_cdi_from_exported_handle(
```
jhand2, line 235: Isn't this internal to Caliptra RT FW since SIGN_WITH_EXPORTED is a Caliptra command? I'm not sure DPE needs to be able to do this lookup.

clundin25: That's a fair point, I'll remove this from the trait.

crypto/src/openssl.rs
jhand2: Hm. It seems like a bug if the same CDI slot is used twice. We could return an error if it's already defined.

The only case where that wouldn't work is if the DPE code needs to retrieve the stored handle twice for some reason. But I don't think the current code does.

crypto/src/openssl.rs
```diff
@@ -177,6 +194,28 @@ impl Crypto for OpensslCrypto {
...
+
+        if self.export_cdi_slots.len() >= MAX_CDI_HANDLES {
+            return Err(CryptoError::ExportedCdiHandleLimitExceeded);
+        }
+
+        let mut exported_cdi_handle = [0; MAX_EXPORTED_CDI_SIZE];
+        self.rand_bytes(&mut exported_cdi_handle)?;
+        self.export_cdi_slots
```
jhand2, line 213: Another option would be to generate a handle when we create the exported CDI and then just lookup the handle in this function. I don't think it matters much either way, but it's maybe a bit more clear to make the binding always exist once the CDI is created.

dpe/src/x509.rs
```diff
@@ -2390,18 +2393,22 @@ fn create_dpe_cert_or_csr(
     let algs = DPE_PROFILE.alg_len();
     let digest = get_dpe_measurement_digest(dpe, env, args.handle, args.locality)?;
 
-    let key_pair = match cert_type {
+    let (key_pair, cdi) = match cert_type {
```
jhand2, line 2396: I think we talked about this in a past PR, but I can't remember the outcome. It seems like this code could/should just be handled in `derive_context.rs` since it's unrelated to generating X.509 certs. Then we can just pass the public key and CertificateType (to decide IsCA) into the X.509 code.

crypto/src/openssl.rs
clundin25: Added code to make this an error, we can always backtrack this in case we need it later.

crypto/src/openssl.rs
clundin25: I slightly prefer creating the handle since it gives each Crypto implementation some flexibility on creating it but there is no real advantage to it.

What do you think?

dpe/src/x509.rs
clundin25: The issue is below on line 2504. The CSR needs access to the private key to self-sign.

The private key cannot be a function parameter since it's an associated type.

An alternative could be passing in a callback to perform the signature but I think that would lead to a larger code size / confusing call flow.

crypto/src/openssl.rs
jhand2: I just mean creating the handle in `derive_exported_cdi`. So it would still be in the Crypto implementation, just a different place.

clundin25: Oh I understand now! 

That is much cleaner! Thanks! :D

crypto/src/openssl.rs
```diff
@@ -41,23 +44,37 @@ impl Hasher for OpensslHasher {
     }
 }
 
+// Currently only supports one CDI handle but in the future we may want to support multiple.
+const MAX_CDI_HANDLES: usize = 1;
+type ExportedCdiHandle = [u8; MAX_EXPORTED_CDI_SIZE];
```
jhand2, line 49: Should this be defined in lib.rs?

clundin25: Updated. I kept `MAX_CDI_HANDLES` private since I am not sure it belongs in the public API but I am happy to make it public (Not sure how much should be shared between this and the caliptra-sw implementation).

crypto/src/lib.rs
```diff
@@ -206,7 +221,7 @@ pub trait Crypto {
         algs: AlgLen,
         measurement: &Digest,
         info: &[u8],
-    ) -> Result<Self::Cdi, CryptoError>;
+    ) -> Result<[u8; MAX_EXPORTED_CDI_SIZE], CryptoError>;
```
jhand2, line 224: Note from discussion, use `ExportedCdiHandle` for these types (here & derive_key_pair_exported)

PR #380, opened 2025-01-22
Add verification tests for the "export-cdi" DeriveContext feature

clundin25:


verification/client/abi.go
```diff
@@ -238,16 +244,14 @@ type DeriveContextReq[Digest DigestAlgorithm] struct {
 type DeriveContextResp struct {
 	NewContextHandle    ContextHandle
 	ParentContextHandle ContextHandle
+	ExportedCdi         ExportedCdi
+	CertificateSize     uint32
+	NewCertificate      []byte
```
jhand2, line 249: nit: should these be pointers so we can make them `nil` in the non-export case? (Note that Go pointers are refcounted, so we don't need to worry about memory management)

verification/testing/deriveContext.go
```diff
@@ -73,6 +73,45 @@ func TestDeriveContext(d client.TestDPEInstance, c client.DPEClient, t *testing.
...
+	}
+	if resp.ParentContextHandle != client.ContextHandle(bytes.Repeat([]byte{0xFF}, 16)) {
+		t.Fatalf("[FATAL]: Expected invalid ParentContextHandle field but it was set to %v", resp.ParentContextHandle)
+	}
+	if resp.CertificateSize == 0 {
+		t.Fatalf("[FATAL]: Expected CertificateSize to be set but was set to %v", resp.CertificateSize)
+	}
+
+	// Check whether certificate is correctly encoded.
+	if _, err := x509.ParseCertificate(resp.NewCertificate); err != nil {
```
jhand2, line 110: iirc, the certifyKey tests have code to check CA certs specifically. Should refactor that into a helper (if it isn't already) and call those validation functions here?

verification/client/abi.go
clundin25: All Go types have a default value, for example the `[]byte` will be filled with 0s.

I think my preference is to keep it as is because:
1. Would this impact serialization (Acknowledging that I added anonymous types to deserialize).
2. I tried switching to pointers and couldn't make the type checker happy but I'm sure with more elbow grease I could find something that works.

verification/testing/deriveContext.go
clundin25: Added relevant checks from the certifyKey test suite.

I moved those checks into `certs.go` so the codesharing is more visible.

clundin25: Moved it back out of `certs.go`. The diff was unreadable, moving the validation functions should be an atomic commit.

verification/testing/deriveContext.go
```diff
@@ -73,6 +73,60 @@ func TestDeriveContext(d client.TestDPEInstance, c client.DPEClient, t *testing.
...
+	leafCert := checkCertificateStructure(t, resp.NewCertificate)
+
+	certChainBytes, err := c.GetCertificateChain()
+	certChain := checkCertificateChain(t, certChainBytes)
+	if err != nil {
+		t.Fatalf("[FATAL]: Could not get Certificate Chain: %v", err)
+	}
+
+	// []byte{0xA, 17} does not correctly match the deep equal :(
+	label := []byte{10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10}
```
jhand2, line 121: The CA UEID is chosen by the platform, so I don't think we can choose a hard-coded value in this test. Since the intention is that these tests can run against any target.

I think for CA certs, we can just not check the UEID since its up to the integrator anywa.

jhand2 (approved):

verification/testing/certifyKey.go
```diff
@@ -392,6 +391,9 @@ func checkCertifyKeySubjectKeyIdentifierExtension(t *testing.T, extensions []pki
 		if ski == nil {
 			t.Errorf("[ERROR]: The certificate is a CA but the SubjectKeyIdentifier extension is not present.")
 		}
+		if pubkey == nil {
+			return
+		}
```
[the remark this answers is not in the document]
clundin25: Opened https://github.com/chipsalliance/caliptra-dpe/issues/386 to track this.

PR #382, opened 2025-01-24
Always include authority key identifier and subject key identifier

clundin25:


dpe/src/x509.rs
```diff
@@ -489,7 +489,7 @@ impl CertWriter<'_> {
         tagged: bool,
         is_x509: bool,
     ) -> Result<usize, DpeErrorCode> {
-        if !measurements.is_ca || !is_x509 {
```
jhand2, line 492: Based on my read of RFC5280, I think we can keep this check for SKI. Just not for AKI. wdyt?

clundin25: SGTM!

jhand2 (approved):

PR #383, opened 2025-01-25
Add encoding options struct to modify behavior of the x509 encoder

clundin25:
For example, this allows DeriveContext to disable ueid encoding


clundin25:
Made irrelevant by https://github.com/chipsalliance/caliptra-dpe/pull/384

PR #385, opened 2025-01-28
Include ueid actual size in DPE Platform to support variable sized ueids

clundin25:


dpe/src/commands/derive_context.rs
```diff
@@ -302,6 +302,7 @@ impl CommandExecution for DeriveContextCmd {
             cfg_if! {
                 if #[cfg(not(feature = "disable_export_cdi"))] {
                     let ueid = &env.platform.get_ueid()?;
+                    let ueid = ueid.buf.get(..ueid.buf_size as usize).ok_or(DpeErrorCode::Platform(PlatformError::InvalidUeidError))?;
```
jhand2, line 305: uber-nit: Could make this a member function on the Ueid type that returns `Result<&[u8], DpeErrorCode>`

jhand2 (approved):

dpe/src/commands/derive_context.rs
clundin25: Done!

PR #391, opened 2025-02-03
Don't recompile dpe every single time

swenson:
The build script always writes to `arbitrary_max_handles.rs`, which means that a recompile is always required.

This then triggers a rebuild of any project (i.e., Caliptra) that uses this.

This fixes it so that we only write the file if the contents need to be changed, so the build can be cached.


clundin25:
Thanks! Looks like there is a clippy complaint

clundin25 (approved):

mhatrevi (approved):

clundin25 (approved):

swenson:
Could one of you merge? I do not appear to have permission.

PR #399, opened 2025-03-03
Remove leftover CertifyKey CA code

clundin25:
This code is unused / dead code for the previously removed CertifyKey CA feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,10 @@ jobs:
       # just need to update to use the most recent stable.
       - name: Update rustup
         run: |
-          rustup update
+          # Keep this in sync with rust-toolchain.toml
+          rustup toolchain install 1.83-x86_64-unknown-linux-gnu
+          rustup component add --toolchain 1.83-x86_64-unknown-linux-gnu rustfmt
+          rustup component add --toolchain 1.83-x86_64-unknown-linux-gnu clippy
 
       - name: Restore sccache binary
         uses: actions/cache/restore@v3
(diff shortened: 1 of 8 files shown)


dpe/src/commands/derive_context.rs
```diff
@@ -38,6 +38,7 @@ bitflags! {
         const RETAIN_PARENT_CONTEXT = 1u32 << 29;
         const MAKE_DEFAULT = 1u32 << 28;
         const CHANGE_LOCALITY = 1u32 << 27;
+        /// INPUT_ALLOW_CA is no longer supported. This flag is ignored.
         const INPUT_ALLOW_CA = 1u32 << 26;
```
ekoohest, line 42: It should not be replaced by "ALLOW_NEW_CONTEXT_TO_EXPORT"?

In general, I do not see any implementation for allow_new_context_to_export flag.
You're gonna add it in the next round?

clundin25: > You're gonna add it in the next round?

Yes we will add that as a flag in another PR.

I think there were some other features that @jhand2 mentioned we need to add as well, so I think those will land together.

ekoohest (approved):

.github/workflows/ci.yml
```diff
@@ -38,7 +38,10 @@ jobs:
       # just need to update to use the most recent stable.
       - name: Update rustup
         run: |
-          rustup update
+          # Keep this in sync with rust-toolchain.toml
+          rustup toolchain install 1.83-x86_64-unknown-linux-gnu
+          rustup component add --toolchain 1.83-x86_64-unknown-linux-gnu rustfmt
+          rustup component add --toolchain 1.83-x86_64-unknown-linux-gnu clippy
```
jhand2, line 44: nit: Is this related to this change?

jhand2 (approved):

PR #400, opened 2025-03-06
Add support for the ALLOW_NEW_CONTEXT_TO_EXPORT flag in DeriveContext

clundin25:


dpe/src/commands/derive_context.rs
```diff
@@ -44,6 +44,8 @@ bitflags! {
         const RECURSIVE = 1u32 << 24;
         const EXPORT_CDI = 1u32 << 23;
         const CREATE_CERTIFICATE = 1u32 << 22;
+        /// ALLOW_NEW_CONTEXT_TO_EXPORT will default to false when omitted.
+        const ALLOW_NEW_CONTEXT_TO_EXPORT = 1u32 << 21;
```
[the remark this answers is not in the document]
clundin25: Hmm a good point. `INPUT_ALLOW_CA` used that bit previously so I am concerned about backwards compatibility.

I'll make this change, and walk it back if @jhand2 disagrees.

dpe/src/commands/derive_context.rs
```diff
@@ -1215,6 +1227,125 @@ mod tests {
...
+            flags: DeriveContextFlags::EXPORT_CDI
+                | DeriveContextFlags::CREATE_CERTIFICATE
+                | DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT,
+            tci_type: 0,
+            target_locality: TEST_LOCALITIES[0],
+        }
+        .execute(&mut dpe, &mut env, TEST_LOCALITIES[0]);
+        assert_eq!(res, Err(DpeErrorCode::InvalidArgument));
+
+        // Children whose parents set `DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT` should be able to
```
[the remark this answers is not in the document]
clundin25: In v1.0 revision 0.14 I see this text:

> Once disabled, this attribute is inherited and cannot be re-enabled on
subsequent derivations from this context forward. A DPE MUST NOT allow export to be enabled for a derived
context when the parent context does not have export enabled.

Does that answer your question? I am not sure I have understood correctly.

dpe/src/commands/derive_context.rs
```diff
@@ -44,6 +44,8 @@ bitflags! {
         const RECURSIVE = 1u32 << 24;
         const EXPORT_CDI = 1u32 << 23;
         const CREATE_CERTIFICATE = 1u32 << 22;
+        /// ALLOW_NEW_CONTEXT_TO_EXPORT will default to false when omitted.
```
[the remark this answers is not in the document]
clundin25: Removed!

dpe/src/commands/initialize_context.rs
```diff
@@ -99,6 +99,7 @@ impl CommandExecution for InitCtxCmd {
             allow_x509: true,
             uses_internal_input_info: false,
             uses_internal_input_dice: false,
+            allow_export_cdi: true,
```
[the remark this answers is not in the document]
clundin25: This initializes the root node. If the root node does not `allow_export_cdi` then `ALLOW_NEW_CONTEXT_TO_EXPORT` always evaluates to false.

dpe/src/commands/derive_context.rs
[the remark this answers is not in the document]
jhand2: Ya, go ahead and reuse bit 26. Thanks!

dpe/src/commands/derive_context.rs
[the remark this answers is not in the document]
ekoohest: You're right by referring to the main DPE spec. I only had iROT spec in my considerations.

dpe/src/commands/derive_context.rs
```diff
@@ -103,6 +102,11 @@ impl DeriveContextCmd {
...
     }
 
+    pub const fn allows_new_context_to_export(&self) -> bool {
+        self.flags
+            .contains(DeriveContextFlags::ALLOW_NEW_CONTEXT_TO_EXPORT)
+    }
+
     /// Whether it is okay to make a default context.
     ///
     /// When a default context is in a locality, it MUST be the only context in the locality. This
```
ekoohest, line 112: Everything about allows_x509 can be removed including line 222, 214, etc.
It can be addressed under another PR, though.

dpe/src/commands/derive_context.rs
```diff
@@ -222,6 +226,7 @@ impl CommandExecution for DeriveContextCmd {
             || (self.exports_cdi() && self.retains_parent())
```
ekoohest, line 226: Random comment... Is it correct condition in here?
The client would like to export_cdi on the new context and retain the parent... sounds logical to me, why it has to be an error?

dpe/src/commands/derive_context.rs
```diff
@@ -222,6 +226,7 @@ impl CommandExecution for DeriveContextCmd {
...
             || (self.exports_cdi()
                 && dpe.contexts[parent_idx].context_type == ContextType::Simulation)
+            || (self.exports_cdi() && !dpe.contexts[parent_idx].allow_export_cdi())
```
ekoohest, line 229: Shouldn't be a condition to support 

     Once disabled, this attribute is inherited and cannot be re-enabled on subsequent derivations from this context forward.

Like : (self.allow_export_cdi() && !dpe.contexts[parent_idx].allow_export_cdi())

dpe/src/commands/derive_context.rs
clundin25: You're right, this condition is not in the spec and I think it should be removed.

dpe/src/commands/derive_context.rs
clundin25: This condition checks the current context, so `self.allow_export_cdi()` should not impact the current command, but subsequent derivations.

> Indicates whether the DPE permits export of the CDI from the newly derived
context. If false, a DPE MUST ensure subsequent DeriveContext operations using the derived context do not
allow export-cdi to be set to true.

I don't think this flag imposes any limitation on the current derivation. What do you think?

ekoohest: This condition is correct. That’s not the concern.

I am talking about an extra condition which addresses allow_export_cdi inheritance requirement.

dpe/src/commands/initialize_context.rs
```diff
@@ -99,6 +99,7 @@ impl CommandExecution for InitCtxCmd {
             allow_x509: true,
```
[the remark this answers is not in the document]
clundin25: I'll clean up these in a separate PR

dpe/src/commands/derive_context.rs
clundin25: Opened https://github.com/chipsalliance/caliptra-dpe/pull/403 to address this.

Thanks for catching this!

dpe/src/commands/derive_context.rs
clundin25: The condition is passed to the child here: https://github.com/chipsalliance/caliptra-dpe/blob/4392f3fae987ed066afc5c2506eefb12b978c366/dpe/src/commands/derive_context.rs#L372-L373.

That should address the inheritance, right?

ekoohest: Correct, thanks @clundin25

ekoohest (approved):

ekoohest (approved):

jhand2 (approved):

PR #401, opened 2025-03-11
Make DICE x509 extensions criticality configurable in DpeInstance

clundin25:
`DpeInstance::with_flags` adds an API to pass `DpeInstanceFlags` to
configure the behavior of a `DpeInstance`.

`DpeInstanceFlags::MARK_DICE_EXTENSIONS_CRITICAL` will mark all DICE
extensions to be critical.

By default extensions will no longer be marked critical.

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -2,7 +2,7 @@
 use super::CommandExecution;
 use crate::{
     context::ContextHandle,
-    dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
+    dpe_instance::{DpeEnv, DpeInstance, DpeInstanceFlags, DpeTypes},
     response::{CertifyKeyResp, DpeErrorCode, Response, ResponseHdr},
     x509::{create_dpe_cert, create_dpe_csr, CreateDpeCertArgs, CreateDpeCertResult},
     DPE_PROFILE, MAX_CERT_SIZE,
@@ -95,6 +95,9 @@ impl CommandExecution for CertifyKeyCmd {
             key_label: &self.label,
             context: b"ECC",
             ueid: &self.label,
+            dice_extensions_are_critical: dpe
+                .flags
+                .contains(DpeInstanceFlags::MARK_DICE_EXTENSIONS_CRITICAL),
         };
         let mut cert = [0; MAX_CERT_SIZE];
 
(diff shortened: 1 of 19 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -193,68 +196,25 @@ mod tests {
...
-            Ok(Command::CertifyKey(&TEST_CERTIFY_KEY_CMD)),
-            Command::deserialize(&command)
-        );
+    struct CertiKeyTestHelperArgs {
+        mark_dice_extensions_critical: bool,
     }
 
-    #[test]
-    fn test_certify_key_x509() {
+    fn certify_key_csr_test_helper(args: CertiKeyTestHelperArgs) {
```
jhand2, line 203: It looks like this is only called by one test. Should this just be moved to that test>

dpe/src/commands/derive_context.rs
```diff
@@ -437,6 +438,112 @@ mod tests {
         target_locality: 0x10CA_1171,
     };
 
+    struct TestCAHelperArgs {
+        mark_dice_extensions_critical: bool,
+    }
+
+    fn create_ca_test_helper(args: TestCAHelperArgs) {
```
jhand2, line 445: Same as other comment. Looks like this is only called in one place.

dpe/src/dpe_instance.rs
```diff
@@ -129,8 +172,55 @@ impl DpeInstance {
...
+    /// # Arguments
+    ///
+    /// * `env` - DPE environment containing Crypto and Platform implementations
+    /// * `support` - optional functionality the instance supports
+    /// * `tci_type`- tci_type of initialized context
+    /// * `auto_init_measurement` - TCI data of initialized context
+    /// * `flags` - configures `Self` behaviors.
+    #[cfg_attr(not(feature = "no-cfi"), cfi_impl_fn)]
+    #[cfg(not(feature = "disable_auto_init"))]
+    pub fn with_flags_auto_init(
```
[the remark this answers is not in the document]
clundin25: Done!

dpe/src/commands/derive_context.rs
clundin25: Done!

dpe/src/commands/certify_key.rs
clundin25: Done!

dpe/src/dpe_instance.rs
```diff
@@ -41,19 +43,40 @@ pub struct DpeEnv<'a, T: DpeTypes + 'a> {
...
+    Zeroize,
+)]
+pub struct DpeInstanceFlags(pub u32);
+
+bitflags! {
+    impl DpeInstanceFlags: u32 {
+        /// Can only successfully execute the initialize context command for non-simulation (i.e.
+        /// `InitializeContext(simulation=false)`) once per reset cycle.
+        ///
+        /// NOTE: This flag is for internal use only.
```
jhand2, line 64: Since flags are passed in the constructor, I think we may want to avoid tracking initialization state in flags. Probably makes sense as a separate internal bool.

jhand2: If you want to avoid increasing DpeInstance size, these flags can probably be a u16

clundin25: Done, moving it out of the flags uses 70 more bytes of code size 😭

jhand2 (approved):

PR #402, opened 2025-03-13
Add a compile time check to verify that MAX_HANDLES cannot be greater than 32

clundin25:
When compiling for tests the compiler will already catch this but this introduces an error
message that is easier to debug and adds a check for non-test targets.


dpe/src/context.rs
```diff
@@ -32,6 +32,12 @@ pub struct Context {
     pub reserved: [u8; 2],
 }
 
+// As long as a `u32` is used for the children bit map the MAX_HANDLES upper bound is 31.
+const _: () = assert!(
+    MAX_HANDLES <= 31,
+    "More than 31 MAX_HANDLES will cause an arithmatic overflow."
```
jhand2, line 38: Why 31 instead of 32?

clundin25: A u32 cannot contain a left shift of 32 bits

Here is the left shift: https://github.com/chipsalliance/caliptra-dpe/blob/main/dpe/src/context.rs#L103

clundin25: You are right, it should be 32 instead of 31, I misread https://github.com/chipsalliance/caliptra-dpe/blob/main/dpe/src/context.rs#L100.

zhalvorsen (approved):

PR #403, opened 2025-03-14
Allow RETAIN_PARENT_CONTEXT with EXPORT_CDI

clundin25:


dpe/src/commands/derive_context.rs
```diff
@@ -315,9 +327,17 @@ impl CommandExecution for DeriveContextCmd {
...
                     )?;
 
-                    Ok(Response::DeriveContextExportedCdi(DeriveContextExportedCdiResp {
+                    let parent_handle = if self.retains_parent() {
+                        // At this point we cannot error out anymore, so it is safe to set the updated child and parent contexts.
+                        dpe.contexts[parent_idx] = tmp_parent_context;
+                        dpe.contexts[parent_idx].handle
+                    } else {
+                        ContextHandle::new_invalid()
+                    };
```
jhand2: Should `dpe.contexts[parent_idx] = tmp_parent_context;` be unconditional? Because if `!self.retains_parent()`, we still need to mark the parent context as retuired.

jhand2 (approved):

PR #405, opened 2025-04-22
Add a version to `DpeInstance`.

zhalvorsen:
When Caliptra performs a hitless update, it retains the DPE state across a warm reset so it can continue with the same measurements.

The structure needs to change to be able to remove the profile feature flags. This version number can be used to ensure the firmware can correctly interpret the structure. Ideally this would be communicated before the hitless update is applied. In the meantime, this will at least cause an error without silently causing problems later.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -1412,7 +1412,7 @@ mod tests {
         // Create a new env to clear cached exported CDIs
         let mut env = DpeEnv::<TestTypes> {
             crypto: OpensslCrypto::new(),
-            platform: DefaultPlatform,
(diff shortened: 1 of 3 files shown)


clundin25 (approved):

jhand2:
Should we do this for the entire PersistentData driver instead of making it DPE-specific?

zhalvorsen:
Would it be overkill to do both? I'm afraid if they are too separated it might be hard to remember to update it.

jhand2:
> Would it be overkill to do both? I'm afraid if they are too separated it might be hard to remember to update it.

Makes sense. We can do both.

jhand2 (dismissed):

jhand2 (approved):

PR #407, opened 2025-04-25
Make the profile in the response header dynamic.

zhalvorsen:
This makes all of the constructors for `ResponseHdr` take a parameter for the profile. This is the first step in removing the cargo features for the different profiles. Instead it will be stored in the `DpeInstance` and filled in at runtime.

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -3,7 +3,7 @@ use super::CommandExecution;
 use crate::{
     context::ContextHandle,
     dpe_instance::{DpeEnv, DpeInstance, DpeInstanceFlags, DpeTypes},
-    response::{CertifyKeyResp, DpeErrorCode, Response, ResponseHdr},
+    response::{CertifyKeyResp, DpeErrorCode, Response},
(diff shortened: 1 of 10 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -151,7 +151,7 @@ impl CommandExecution for CertifyKeyCmd {
             derived_pubkey_y,
             cert_size,
             cert,
-            resp_hdr: ResponseHdr::new(DpeErrorCode::NoError),
+            resp_hdr: ResponseHdr::new(dpe.profile, DpeErrorCode::NoError),
```
clundin25, line 154: It might be nice to wrap this with a helper function in `DpeInstance`, in case a new call-site sets it incorrectly.

clundin25 (approved):

dpe/src/commands/certify_key.rs
zhalvorsen: That's a good idea, done.

clundin25 (approved):

jhand2 (approved):

PR #409, opened 2025-04-25
Make the profile in `CommandHdr` dynamic.

zhalvorsen:
This makes `CommandHdr` no longer require the profile cargo feature. This will now check the profile based on a parameter passed into the `Command::deserialize` function instead.

Many of the tests now check deserializing commands in both profiles. Some of these will need to specialized more if the profile affects the contents of the command. Currently the contents are still hard coded.


dpe/src/commands/mod.rs
```diff
@@ -148,13 +148,23 @@ pub struct CommandHdr {
 impl CommandHdr {
     const DPE_COMMAND_MAGIC: u32 = u32::from_be_bytes(*b"DPEC");
 
-    pub fn new_for_test(cmd_id: u32) -> CommandHdr {
+    pub fn new_for_test(profile: DpeProfile, cmd_id: u32) -> CommandHdr {
```
jhand2, line 151: nit: SInce there's nothing test-specific about this, maybe we should just call the function `new`

jhand2 (approved):

dpe/src/commands/mod.rs
zhalvorsen: That's better, done.

clundin25 (dismissed):

PR #410, opened 2025-04-25
Allow dangling retired contexts

clundin25:
ExportCDI can retire a parent without creating a new child.
This is because ExportCDI NEVER creates a new context.


zhalvorsen (approved):

zhalvorsen (dismissed):

jhand2:
Discussed offline, but I think during DeriveContext if parent isn't retained and no children are remaining, we should clean up the context instead of leaving it as retired.

dpe/src/commands/derive_context.rs
```diff
@@ -1570,4 +1852,35 @@ mod tests {
             };
         }
     }
+
+    fn dpe_command_and_expected_child_count(
```
zhalvorsen, line 1856: nit: maybe a name `dpe_command*` feels a bit generic because it can only support DeriveContextCmd. Would a name like `derive_context_and_expected_child_count` be more applicable?

zhalvorsen (approved):

dpe/src/commands/derive_context.rs
clundin25: I agree! I updated the names.

zhalvorsen (approved):

jhand2 (approved):

PR #412, opened 2025-05-06
crypto: Add ML-DSA87 test artifacts

clundin25:


crypto/build.rs
```diff
@@ -3,28 +3,32 @@
 #[cfg(feature = "rustcrypto")]
 use std::ops::Deref;
 
+#[cfg(any(feature = "rustcrypto", feature = "openssl"))]
+use std::fs;
+
+use std::{env, path::Path};
+
 fn main() {
-    #[cfg(feature = "openssl")]
```
clundin25, line 7: Sorry for the messy diff. I refactored some duplicate code.

crypto/build.rs
```diff
@@ -3,28 +3,32 @@
...
+
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let _path_256 = Path::new(&out_dir).join("alias_priv_256.pem");
+    let _path_384 = Path::new(&out_dir).join("alias_priv_384.pem");
 
-        println!("cargo:rerun-if-changed={ALIAS_PRIV_256}");
-        println!("cargo:rerun-if-changed={ALIAS_PRIV_384}");
+    //TODO(clundin): Refactor / include support for ML-DSA 87 when the openssl crate supports it.
+    #[cfg(feature = "openssl")]
+    let (_pem_256, _pem_384) = {
```
zhalvorsen, line 27: Correct me if I'm wrong, but I think `_pem_256` and `_pem_384` can both have their leading underscores removed because they are used to write the file at the end.

zhalvorsen (approved):

crypto/build.rs
clundin25: Good point! I shifted things around to remove the underscores without introducing an unused variable warning.

zhalvorsen (approved):

PR #416, opened 2025-05-14
DeriveContext SVN Support

clundin25:
Add SVN as a required field to `DeriveContext`.

The SVN will be encoded in the `DiceTcbInfo` extension.


zhalvorsen (approved):

clundin25:
Quick note: This change also requires a spec bump.

Issue #417, opened 2025-05-16
Remove OpenSSL tests / mock


clundin25:
Instead we will only use Rust Crypto.

PR #418, opened 2025-05-16
Remove OpenSSL

clundin25:
This resolves https://github.com/chipsalliance/caliptra-dpe/issues/417.


ci.sh
```diff
@@ -12,19 +12,19 @@ function build_rust_targets() {
   cargo build --release --manifest-path crypto/Cargo.toml --no-default-features
   cargo build --release --manifest-path platform/Cargo.toml --no-default-features
   cargo build --release --manifest-path dpe/Cargo.toml --features=$profile --no-default-features
-  cargo build --release --manifest-path simulator/Cargo.toml --features=$profile,openssl --no-default-features
+  cargo build --release --manifest-path simulator/Cargo.toml --features=$profile,rustcrypto --no-default-features
```
jhand2, line 15: should we just drop this feature and assume it's always set?

jhand2 (approved):

ci.sh
jhand2: ok to do in a separate pr

clundin25: I'll follow up - I think this will be a little tricky to do

PR #426, opened 2025-06-13
Simplify command dispatch.

zhalvorsen:
This removes the `Command` enum completely `deserialize_command` instead returns a `dyn` reference to the `CommandExecution` trait.

This uses a `dyn` reference instead of an `impl` because `dyn` can take a performance hit but saves on code size because it is dynamically dispatched. We are low on memory in Caliptra RT firmware and we are not concerned with performance.

Dyn references can't have `impl` types in any of the function parameters so this also switches the `impl DpeTypes` for `DpeEnv` to be a generic.

Also because we can't check equality anymore because the compiler doesn't know what kind of object it is, the deserialization tests just check for OK now.

### DIFF
--- a/dpe/fuzz/src/fuzz_target_1.rs
+++ b/dpe/fuzz/src/fuzz_target_1.rs
@@ -57,7 +57,7 @@ fn harness(data: &[u8]) {
     };
     let mut dpe = DpeInstance::new(&mut env).unwrap();
     trace!("----------------------------------");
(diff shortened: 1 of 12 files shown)


zhalvorsen:
It turns out this is used pretty extensibly in https://github.com/chipsalliance/caliptra-sw/blob/776d7f94339b17251f150d1335e0488a586980b0/runtime/src/invoke_dpe.rs#L94 so if we wanted to do something like this, we will need to be more do some refactoring first.

PR #433, opened 2025-07-09
Some Crypto interface optimizations.

zhalvorsen:
A couple optimizations found while using the new Crypto interface.

Feel free to push back on these if there is a specific reason for any of them.


clundin25 (approved):
Much nicer!

PR #437, opened 2025-07-10
ML-DSA x509 encoding

clundin25:
Adds support for encoding an ML-DSA 87 certificate.

Adds a few new types, for example to CertifyKey, so the new certificate can be encoded using the DPE API.

Manually verified using the sample DPE cert tool + parsing the certificate with the OpenSSL CLI.

I will add:
* Unit tests
* Verification tests

In subsequent PRs.


zhalvorsen:
@jhand2 can you take a look at the ASN.1 encoding? You are a lot more familiar with it than I am.

dpe/src/commands/certify_key.rs
```diff
@@ -398,6 +438,10 @@ mod tests {
             let hash_alg_oid = match DPE_PROFILE.alg() {
                 SignatureAlgorithm::Ecdsa(EcdsaAlgorithm::Bit256) => "2.16.840.1.101.3.4.2.1",
                 SignatureAlgorithm::Ecdsa(EcdsaAlgorithm::Bit384) => "2.16.840.1.101.3.4.2.2",
+                #[cfg(feature = "ml-dsa")]
+                SignatureAlgorithm::MlDsa(MldsaAlgorithm::ExternalMu87) => {
+                    "2.16.840.1.101.3.4.3.19"
```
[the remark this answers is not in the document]
clundin25: I maybe should swap this with a `todo`.

I think for the cert HASH OID _is_ included in the `DiceTcbInfo` extension. I have not yet dug into the CSR so it may need some tweaks for that case.

zhalvorsen (dismissed):

dpe/src/commands/sign.rs
```diff
@@ -198,6 +202,15 @@ pub struct SignP384Cmd {
     pub digest: [u8; 48],
 }
 
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq, IntoBytes, FromBytes, Immutable, KnownLayout)]
+pub struct SignMldsaExternalMu87Cmd {
+    pub handle: ContextHandle,
+    pub label: [u8; 48],
+    pub flags: SignFlags,
+    pub digest: [u8; 48],
```
jhand2, line 211: This should be 64 bytes.

But also, related to 1 v.s. 2 profiles, let's chat about whether it makes sense to make the to-be-signed data hash-sized or 1k.

dpe/src/lib.rs
```diff
@@ -31,7 +31,7 @@ use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, TryFromBytes};
 pub use crypto::{ecdsa::EcdsaAlgorithm, ExportedCdiHandle, MAX_EXPORTED_CDI_SIZE};
 
 // Max cert size returned by CertifyKey
-const MAX_CERT_SIZE: usize = 6144;
+const MAX_CERT_SIZE: usize = 7872;
```
jhand2, line 34: I think we discussed this a little bit before, but we may want to consider a chunking mechanism for retrieving certs in 2.x. Ok to leave out of this PR, but could you open an issue?

dpe/src/x509.rs
```diff
@@ -2526,36 +2751,26 @@ fn create_dpe_cert_or_csr(
...
+                &measurements,
+                &cert_validity,
+            )?;
+            if bytes_written > MAX_CERT_SIZE {
+                return Err(DpeErrorCode::InternalError);
             }
+            let tbs_digest = env.crypto.hash(&scratch_buf[..bytes_written])?;
+
+            // TODO(clundin): We need to make sure we handle ExternalMu well here.
+            let sig = env.crypto.sign_with_alias(&tbs_digest)?;
```
jhand2, line 2769: This part is interesting. Since we have all the data for this signature in SRAM we have two choices:

1. Just pass the whole buffer to the ML-DSA peripheral
2. Compute mu and pass that

(1) is probably slightly more efficient. Although we would need to change `sign_with_alias` to accept the whole message.

If we decide to do (2), we could add a crypto API with something like `prepare_tbs` which would hash for ECDSA and compute mu for ML-DSA.

crypto/src/hkdf.rs
```diff
@@ -38,9 +38,9 @@ pub fn hkdf_derive_cdi(
             // This block assumes that the size of `xi` is the same as `SHA256`.
             const _: () = assert!(MldsaAlgorithm::ExternalMu87.seed_size() == 256 / 8);
 
-            let hk = Hkdf::<Sha256>::new(Some(info), measurement.bytes());
+            let hk = Hkdf::<Sha256>::new(Some(info), measurement.as_slice());
```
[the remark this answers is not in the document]
clundin25: Tracking in https://github.com/chipsalliance/caliptra-dpe/issues/441

dpe/src/commands/certify_key.rs
```diff
@@ -50,6 +52,10 @@ impl CertifyKeyCommand<'_> {
             DpeProfile::P384Sha384 => {
                 CertifyKeyCommand::parse_command(CertifyKeyCommand::P384, bytes)
             }
+            #[cfg(feature = "ml-dsa")]
+            DpeProfile::Mldsa87ExternalMu => {
```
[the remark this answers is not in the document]
clundin25: Chatted offline, punting this for now.

dpe/src/x509.rs
jhand2: Nice. I like that. We can handle in a follow-up PR

PR #438, opened 2025-07-11
Convert some TciMeasurement parameters to Digest.

zhalvorsen:
This removes more dependencies on `DPE_PROFILE`.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -5,7 +5,6 @@ use crate::{
     context::{ActiveContextArgs, Context, ContextHandle, ContextState, ContextType},
     dpe_instance::{DpeEnv, DpeInstance, DpeTypes},
     response::{DeriveContextExportedCdiResp, DeriveContextResp, DpeErrorCode, Response},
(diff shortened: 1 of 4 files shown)


dpe/src/commands/derive_context.rs
```diff
@@ -457,6 +456,19 @@ mod tests {
...
+            Self {
+                handle: ContextHandle::default(),
+                data: [0; DPE_PROFILE.tci_size()],
+                flags: DeriveContextFlags(0),
+                tci_type: 0,
+                target_locality: 0,
+                svn: 0,
+            }
+        }
+    }
```
clundin25, line 470: Why put it in the test module? Could this be useful elsewhere?

Not a huge deal but curious

clundin25 (approved):

dpe/src/commands/derive_context.rs
zhalvorsen: That's a good point. I put it there because we weren't using elsewhere, but there is no reason to stop us from using it in the future. Move to non-test code.

clundin25 (approved):

PR #440, opened 2025-07-22
X509 module code size optimization

clundin25:
This saves ~56 bytes in caliptra-sw.


zhalvorsen:
Do we get similar behavior if we use `okref` instead?

zhalvorsen:
> > Do we get similar behavior if we use `okref` instead?
> 
> I think that is likely. Do we want to add that pattern to this crate? I have no preference

Yeah, I think that would be good. It will probably be helpful in other scenarios too where we are returning large `Result` types

clundin25:
Opened https://github.com/chipsalliance/caliptra-dpe/issues/449 to track this.

PR #444, opened 2025-07-23
Build DeriveContext for multiple profiles.

zhalvorsen:
I will change the DPE response structure in a follow on PR. This is already a big change.


dpe/src/commands/certify_key.rs
```diff
@@ -263,11 +263,15 @@ pub struct CertifyKeyMldsaExternalMu87Cmd {
 mod tests {
     use super::*;
     #[cfg(feature = "dpe_profile_p256_sha256")]
-    use crate::commands::CertifyKeyP256Cmd as CertifyKeyCmd;
+    use crate::commands::{
```
clundin25, line 266: Can you see if this will compile with `cargo check --features ml-dsa`?

I suspect it will complain

PR #445, opened 2025-07-23
Simplify DeriveContext tests using the new default implementation.

zhalvorsen:


clundin25 (approved):
Much nicer. The caliptra-sw diff will be incredible :D

PR #447, opened 2025-07-24
Add CI build check for ml-dsa feature

clundin25:


clundin25:
> Thanks for also combining `ml-dsa` and `dpe_ml_dsa`. I hadn't noticed there were two until today.

Yes, sorry for that :D

PR #466, opened 2025-11-07
fixing getProfile method signature

arul-sri:


dpe/src/commands/get_profile.rs
```diff
@@ -0,0 +1,35 @@
...
+    Debug,
+    PartialEq,
+    Eq,
+    zerocopy::FromBytes,
+    zerocopy::IntoBytes,
+    zerocopy::Immutable,
+    zerocopy::KnownLayout,
+    Default,
+)]
+pub struct GetProfileCmd {}
```
zhalvorsen: In Rust, typically if a type has no parameters you can leave off the curly braces:

```suggestion
pub struct GetProfileCmd;
```

dpe/src/dpe_instance.rs
```diff
@@ -143,7 +143,7 @@ impl DpeInstance {
     ) -> Result<Response, DpeErrorCode> {
         let command = self.deserialize_command(cmd)?;
         let resp = match cfi_launder(command) {
-            Command::GetProfile => Ok(Response::GetProfile(
+            Command::GetProfile(_) => Ok(Response::GetProfile(
                 self.get_profile(&mut env.platform, env.state.support)?,
             )),
```
zhalvorsen: Now this command is shaped like the others, you can let the command handle it internally.

```suggestion
            Command::GetProfile(cmd) => cmd.execute(self, env, locality),
```

zhalvorsen (commented):
Looks good! Nice job!

PR #507, opened 2026-01-28
Vendor `okref` functions

zhalvorsen:
Unfortunately, we cannot depend on caliptra-okref directly due to dependency cycles. So I copied the relevant code here.

Fortunately, it is very small and simple.


arul-sri (approved):
The mitigation to the circular dependency looks good to me!

PR #509, opened 2026-01-29
Introduce 64 byte ML-DSA digest

arul-sri:


verification/testing/sign.go
```diff
@@ -27,6 +27,10 @@ func TestAsymmetricSigning(d client.TestDPEInstance, c client.DPEClient, t *test
 	}
 
 	digestLen := profile.GetDigestSize()
```
zhalvorsen: Can you add a new function to `Profile` to get the `SignData` size? I think it would make these more readable when we come back to it at some point.

verification/testing/negativeCases.go
```diff
@@ -41,7 +41,11 @@ func TestInvalidHandle(d client.TestDPEInstance, c client.DPEClient, t *testing.
 	}
 
 	// Check Sign with invalid handle
-	if _, err := c.Sign(&InvalidHandle, make([]byte, digestLen), 0, make([]byte, digestLen)); err == nil {
+	tbsLen := digestLen
+	if profile == client.ProfileMldsa87 {
```
zhalvorsen: Can you update this file to call that new function as well?

verification/client/profile.go
```diff
@@ -40,6 +40,14 @@ func (p Profile) GetDigestSize() int {
 	return 0
 }
 
+// GetSignDataSize gets the sign data size of the profile
+func (p Profile) GetSignDataSize() int {
+	if p == ProfileMldsa87 {
+		return 64
+	}
+	return p.GetDigestSize()
```
zhalvorsen: Let's use a switch statement for extra readability:


```suggestion
	switch p {
	case ProfileMinP256SHA256:
		fallthrough
	case ProfileP256SHA256:
		return 32
	case ProfileMinP384SHA384:
		fallthrough
	case ProfileP384SHA384:
		return 48
	case ProfileMldsa87:
		return 64
	}
	return 0
```

PR #514, opened 2026-02-05
Start work for multi-part responses

zhalvorsen:
This starts implementation of the change in the DPE iRoT profile to enable breaking large responses into chunks so they can more easily be sent over small mailbox interfaces.

This is especially relevant for Caliptra 2.1 ML-DSA certificates. The mailbox is only 16KB and CSRs can be as large as 22KB for 64 contexts.

Integration with commands will come in subsequent PRs.

### DIFF
--- a/dpe/src/operation_handle.rs
+++ b/dpe/src/operation_handle.rs
@@ -1,0 +1,90 @@
+// Licensed under the Apache-2.0 license.
+
+use constant_time_eq::constant_time_eq_16;
+use crypto::Crypto;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+use zeroize::Zeroize;
+
+use crate::{
+    dpe_instance::{DpeEnv, DpeTypes},
+    response::DpeErrorCode,
+};
+
+#[repr(C)]
+#[derive(
+    Debug,
+    Default,
+    Clone,
+    Copy,
+    IntoBytes,
+    FromBytes,
+    Immutable,
+    KnownLayout,
+    Zeroize,
+    PartialEq,
+    Eq,
+)]
+pub struct OperationHandle(pub [u8; OperationHandle::SIZE]);
+
+impl OperationHandle {
+    pub const SIZE: usize = 16;
+    pub const MAX_NEW_HANDLE_ATTEMPTS: usize = 8;
+    const DEFAULT: OperationHandle = OperationHandle([0; Self::SIZE]);
+
+    /// Returns the default operation handle.
+    pub const fn default() -> OperationHandle {
+        Self::DEFAULT
+    }
+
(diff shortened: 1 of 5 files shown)


dpe/src/state.rs
```diff
@@ -25,6 +26,44 @@ bitflags! {
...
+
+    pub fn blank(&self) -> bool {
+        self.handle.blank() && self.offset == 0 && self.digest == [0; HASH_SIZE]
+    }
+
+    pub fn active(&self) -> bool {
+        !self.handle.blank() && self.offset != 0 && self.digest != [0; HASH_SIZE]
+    }
+
+    pub fn continued_operation(&self, handle: &OperationHandle, digest: &Digest) -> bool {
```
clundin25: nit: 

```suggestion
    pub fn is_continued_operation(&self, handle: &OperationHandle, digest: &Digest) -> bool {
```

dpe/src/state.rs
```diff
@@ -55,25 +95,22 @@ impl Default for State {
...
     pub const MAGIC: u32 = u32::from_be_bytes(*b"DPES");
     pub const VERSION: u32 = 1;
+    pub const MAX_MULTIPART_OPERATIONS: usize = 2;
 
     pub fn new(support: Support, flags: DpeFlags) -> Self {
         let updated_support = support.preprocess_support();
-        const CONTEXT_INITIALIZER: Context = Context::new();
         State {
-            marker: Self::MAGIC,
-            version: Self::VERSION,
```
clundin25: Won't these be `0` now?

Maybe there is something later in the PR.

clundin25 (dismissed):

dpe/src/operation_handle.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub const MAX_NEW_HANDLE_ATTEMPTS: usize = 8;
+    const DEFAULT: OperationHandle = OperationHandle([0; Self::SIZE]);
+
+    /// Returns the default operation handle.
+    pub const fn default() -> OperationHandle {
+        Self::DEFAULT
+    }
+
+    pub fn generate(env: &mut DpeEnv<impl DpeTypes>) -> Result<OperationHandle, DpeErrorCode> {
+        for _ in 0..Self::MAX_NEW_HANDLE_ATTEMPTS {
```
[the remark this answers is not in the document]
zhalvorsen: That's true, the odds of it happening once are essentially 0, let alone 8 times in a row. Maybe we should add a function to the `Platform` trait to call a fatal error.

Created #515 to track

dpe/src/state.rs
zhalvorsen: It takes the values from `State::default()` now. I added a unit test just to make sure

PR #517, opened 2026-02-12
[certificates] Remove criticality flags when always false

### LINKED ISSUE #513: DPE leaf certificate's AIK (authorityKeyIdentifier)'s critical element is set and exposed as it's default value (FALSE) which violates x.690 (ASN.1 encoding rule)
In Caliptra-DPE, the current forming of the certificate during CertifyKey hard code a false boolean flag to the "Critical" element of the AIK extension.
According to x.690 specification https://www.itu.int/rec/T-REC-X.690 under 11.5 it states that "11.5 Set and sequence components with default value
The encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value."  This actually causes some stricter cert verifier to error out as the default critical definition is FALSE.  
```
 1458:d=5  hl=2 l=   3 prim: OBJECT            :X509v3 Authority Key Identifier
 1463:d=5  hl=2 l=   1 prim: BOOLEAN           :0  <- Error
 1466:d=5  hl=2 l=  24 prim: OCTET STRING      [HEX DUMP]:30168014CDA8870601D824E5CAA09819A704423FAE3136F2
 1492:d=1  hl=2 l=  10 cons: SEQUENCE
 1494:d=2  hl=2 l=   8 prim: OBJECT            :ecdsa-with-SHA384
 1504:d=1  hl=2 l= 104 prim: BIT STRING
```

Caliptra-SW and the forming of AliasFMC/AliasRT does not have this problem as it does not encode the critical flag at all
```
  622:d=5  hl=2 l=   3 prim: OBJECT            :X509v3 Authority Key Identifier
  627:d=5  hl=2 l=  24 prim: OCTET STRING      [HEX DUMP]:30168014B0395AF420291BB4B096216D4D3E9B01C61C8CF3
  653:d=1  hl=2 l=  10 cons: SEQUENCE
  655:d=2  hl=2 l=   8 prim: OBJECT            :ecdsa-with-SHA384
  665:d=1  hl=2 l= 105 prim: BIT STRING
```

@jhand2  do you agree with this being an issue?

zhalvorsen:
Fixes #513

From the issue: According to x.690 specification https://www.itu.int/rec/T-REC-X.690 under 11.5 it states that "The encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value."

Criticality's default value is false which is what we were hard-coding in for these values. This removes them from the certificate.

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -723,7 +723,6 @@ impl CertWriter<'_> {
         // in tagging twice.
         let ext_size = Self::get_structure_size(aki_size, /*tagged=*/ true)?;
         let size = Self::get_structure_size(Self::AUTHORITY_KEY_IDENTIFIER_OID.len(), /*tagged=*/true)? // Extension OID
-            + Self::get_structure_size(Self::BOOL_SIZE, /*tagged=*/true)? // Critical bool
             + Self::get_structure_size(ext_size, /*tagged=*/true)?; // OCTET STRING
 
         Self::get_structure_size(size, tagged)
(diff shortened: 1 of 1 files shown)


jameszhang-nvidia (approved):
LGTM

PR #523, opened 2026-02-19
Refactor CI

embediver:
Intention is to split CI into multiple workflows and jobs.
- [x] Rust workflow with jobs for different packages
  - [x] Utilize matrix feature to exercise multiple feature flags
- [x] Verification test workflow
- [x] Certificate Parser and Verification test job
- [x] Fuzzer build check
- [x] File header/license check

A workflow for Miri will likely also be added in the future.


.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+  dpe:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        profile: [p256, p384, ml-dsa, hybrid]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Cache cargo registry and build
+        uses: Swatinem/rust-cache@v2
```
clundin25: Can you pin these actions to a commit hash?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install clippy
+        run: rustup component add clippy
+      - name: rustfmt check
+        run: cargo fmt -p dpe -v --check
+      - name: clippy
+        run: cargo clippy -p dpe --features=${{ matrix.profile }} --no-default-features -- --deny=warnings
```
clundin25: Let's pull FMT & Clippy into a separate "link" job that is a pre-req to these jobs.

 I think clippy supports checking against all features at once, can you do that in the lint step?

clundin25 (commented):
This seems to remove the go verification tests?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: build
+        run: cargo build -p dpe --features=${{ matrix.profile }} --no-default-features
+      - name: build(no-cfi)
+        run: cargo build -p dpe --features=${{ matrix.profile }},no-cfi --no-default-features
+      - name: build-release
+        run: cargo build --release -p dpe --features=${{ matrix.profile }} --no-default-features
```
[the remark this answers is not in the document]
embediver: It was present in the original CI script, so I included it, but you are probably right.

clundin25:
I think this got deprioritized in favor of using xtask. Going to close this

PR #532, opened 2026-02-27
Add Sign raw option

arul-sri:


dpe/src/commands/sign.rs
```diff
@@ -35,6 +38,12 @@ pub enum SignCommand<'a> {
     P384(&'a SignP384Cmd),
     #[cfg(feature = "ml-dsa")]
     Mldsa87(&'a SignMldsa87Cmd),
+    #[cfg(feature = "ml-dsa")]
+    Mldsa87Raw {
+        handle: &'a ContextHandle,
+        label: &'a [u8],
+        raw_data: &'a [u8],
```
zhalvorsen: Let's keep this a fixed sized command and give the raw data a maximum size. Let's start with 1024 bytes. It will be easier from the caller's if it is similar to all of the other commands.

dpe/src/commands/sign.rs
```diff
@@ -45,10 +54,61 @@ impl SignCommand<'_> {
...
-            DpeProfile::Mldsa87 => SignCommand::parse_command(SignCommand::Mldsa87, bytes),
+            DpeProfile::Mldsa87 => SignCommand::deserialize_mldsa87(bytes),
             _ => Err(DpeErrorCode::InvalidArgument)?,
         }
     }
+
+    #[cfg(feature = "ml-dsa")]
+    fn deserialize_mldsa87(bytes: &[u8]) -> Result<SignCommand, DpeErrorCode> {
+        // We only need the first 68 bytes (handle + label + flags) to inspect the flag.
+        const FIXED_PREFIX_SIZE: usize = 16 + 48 + 4;
```
zhalvorsen: There is a really cool library called `zerocopy` that makes it pretty easy to deserialize commands. It can do a lot of the heavy lifting for you. I'd recommend using that. You give it a struct for how the layout should look. I'd recommend giving just the first few fields here to be able to pull out the flags. Maybe something like:

```
#[repr(C)]
#[derive(Debug, PartialEq, Eq, IntoBytes, FromBytes, Immutable, KnownLayout)]
struct SignMldsa87Header {
    pub handle: ContextHandle,
    pub label: [u8; 48],
    pub flags: SignFlags,
}
```

Then you can use [`ref_from_prefix`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#method.ref_from_prefix) to deserialize the data.

dpe/src/commands/sign.rs
```diff
@@ -489,4 +558,99 @@ mod tests {
...
+        let cmd = Command::deserialize(DPE_PROFILE, &command);
+        assert!(cmd.is_ok());
+
+        match cmd.unwrap() {
+            Command::Sign(SignCommand::Mldsa87(_)) => {
+                // Expected
+            }
+            _ => panic!("Expected Mldsa87 variant for non-raw mode"),
+        }
+    }
```
zhalvorsen: Can you add a test case that calls `SignCommand::Mldsa87Raw` and verify the signature is correct?

verification/testing/sign_raw.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		t.Fatalf("[FATAL]: Could not CertifyKey: %v", err)
+	}
+	handle = &certifiedKey.Handle
+
+	// Create a test message
+	testMessage := []byte("This is a test message for ML-DSA raw signing")
+
+	// Calculate external mu using the helper function
+	externalMu := client.CalculateExternalMu(testMessage, certifiedKey.Pub.X)
```
zhalvorsen: This test shouldn't need external mu. You should be able to send the `testMessage` directly.

PR #540, opened 2026-03-11
Add a checks for panics

embediver:
- [x] Add binary package for comile-time panic check
- [x] Add clippy lints to prevent panicing behaviour
- [ ] Add CI job to check for panics by compiling the added binary package


clundin25:
I think this was completed in a separate PR

PR #541, opened 2026-03-17
Make `Hasher` dyn compatible

zhalvorsen:
This converts the `Hasher` struct to no longer be `Sized` which will make it easier to start making the `Crypto` trait dyn compatible.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -130,7 +130,10 @@ impl CryptoError {
     }
 }
 
-pub trait Hasher: Sized {
+pub trait Hasher {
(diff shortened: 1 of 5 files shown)


crypto/src/lib.rs
```diff
@@ -142,7 +145,7 @@ pub trait Hasher: Sized {
     ///
     /// Once this function has been called, the object can no longer be used and
     /// a new one must be created to hash more data.
```
clundin25: nit: Documentation no longer matches code.

Out of curiosity, is this a change to make this trait object safe?

clundin25 (approved):

PR #556, opened 2026-04-11
Disallow multiple of the same TCI type

zhalvorsen:
This adds a check to make sure before creating a new context that there isn't already a context with the same type. This should help disambiguate measurements.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -199,6 +199,13 @@ impl DeriveContextCmd {
             c.state == ContextState::Active && c.locality == target_locality
         })?;
 
+        // check if the typ already exists
+        for c in state.contexts.iter() {
+            if c.state != ContextState::Inactive && c.tci.tci_type == self.tci_type {
+                return Ok(false);
+            }
+        }
+
         Ok(if self.flags.makes_default() {
             self.safe_to_make_default(parent_idx, default_context_idx, num_contexts_in_locality)
         } else {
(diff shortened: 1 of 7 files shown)


dpe/src/commands/derive_context.rs
```diff
@@ -194,10 +194,24 @@ impl DeriveContextCmd {
             .get_active_context_pos(&ContextHandle::default(), target_locality)
             .ok();
 
-        // count active contexts in target_locality
-        let num_contexts_in_locality = state.count_contexts(|c: &Context| {
-            c.state == ContextState::Active && c.locality == target_locality
-        })?;
+        // Count active contexts in target_locality and check if there is
```
fdamato: for Attestation Type / Vendor / Model is the target identifier, and has to be unique even across localities

zhalvorsen: Oh really? I didn't realize that. I'll update it to check for the same type in any `Active`/`Retired` contexts

dpe/src/commands/derive_context.rs
```diff
@@ -194,10 +194,24 @@ impl DeriveContextCmd {
...
+        let mut num_contexts_in_locality = 0;
+        let mut tci_type_exists = false;
+        for context in state.contexts.iter() {
+            if context.locality == target_locality {
+                if context.state == ContextState::Active {
+                    num_contexts_in_locality += 1;
+                }
+                if context.tci.tci_type == self.tci_type {
+                    tci_type_exists = true;
+                }
```
clundin25: Could return early here?

clundin25 (dismissed):

dpe/src/commands/derive_context.rs
zhalvorsen: Done

clundin25 (approved):

PR #559, opened 2026-04-15
[x509] Populate registerNum value in the integrity register

### LINKED ISSUE #543: IntegrityRegisters encoding does not follow TCG DICE attestation specification
In [https://trustedcomputinggroup.org/wp-content/uploads/DICE-Attestation-Architecture-v1.2_pub.pdf](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Attestation-Architecture-v1.2_pub.pdf) 6.1.1.4 Integrity Registers, it is clearly specified that

> The register name facilitates comparison with Reference Values given differently ordered lists. **Either registerName or registerNum MUST be supplied**.

However, in this DPE implementation, neither registerName nor registerNum is supplied as in the following snippet: https://github.com/chipsalliance/caliptra-dpe/blob/6dfc502a30d443e5e054db1344a97eeb823ea5b9/dpe/src/x509.rs#L1622-L1629

Without registerName or registerNum, there is no way to model integrity-registers in CoRIM. Here’s the CDDL from CoRIM spec showing that we need a map from registerName or registerNum to the digest:
```
integrity-register-id-type-choice = uint / text

integrity-registers = {
  + integrity-register-id-type-choice => digests-type
}
```

zhalvorsen:
Fixes #543

registerNum and registerName are listed as optional fields, but the spec says one of those values must be set. This sets the registerNum field. Because there is only ever one integrity register per TCB info, we set the value to 0.

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -147,6 +147,11 @@ impl CertWriter<'_> {
     #[cfg(not(feature = "disable_csr"))]
     const CSR_V0: u64 = 0;
 
+    /// registerNum and registerName are listed as optional fields, but the spec
+    /// says one of those values must be set. Because there is only ever one
+    /// integrity register per TCB info, the registerNum is hardcoded to 0.
+    const INTEGRITY_REGISTER_NUM: u64 = 0;
+
     /// ASN.1 encoding with length stripped of the following OID.
     /// id-ml-dnsa-87 OBJECT IDENTIFIER ::= { joint-iso-itu-t(2)
     ///     country(16) us(840) organization(1) gov(101) csor(3)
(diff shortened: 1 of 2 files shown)


dpe/src/x509.rs
```diff
@@ -1608,21 +1609,25 @@ impl CertWriter<'_> {
...
-            // integrityRegusters[0] SEQUENCE
+            // integrityRegisters[0] SEQUENCE
             bytes_written += self.encode_byte(Self::SEQUENCE_TAG)?;
-            bytes_written += self.encode_size_field(fwid_list_size)?;
+            bytes_written += self.encode_size_field(ir_num_size + fwid_list_size)?;
+
+            // IMPLICIT [1] Primitive
+            // registerNum INTEGER
+            bytes_written += self.encode_byte(Self::CONTEXT_SPECIFIC | 0x01)?;
+            bytes_written += self.encode_integer(0, false)?;
```
clundin25: Can you add a comment that explains why we always set this to 0

clundin25 (dismissed):

clundin25 (dismissed):

dpe/src/x509.rs
zhalvorsen: Good call! Done

clundin25 (approved):

Issue #460, opened 2025-10-28
Improved error handling


embediver:
Since call-stacks can get quite deep, a more sophisticated (internal) error handling could be beneficial during development.

`core::error::Error` is stable since Rust 1.81.0 and can be used to to provide a source across abstraction boundaries.

This might add unwanted code size and memory usage by bloating error types.

leongross:
Taking over this issue.

> Thanks for raising this, I totally agree. Because the error codes are defined upstream, it might be best to add a u16 with additional details similar to how the platform and crypto error codes work.

@zhalvorsen could you elaborate on the upstream definition these errors must match?

zhalvorsen:
I was thinking changing the definition of `InternalError` to do something like this:

```Rust
pub enum DpeErrorCode {
    NoError = 0,
    InternalError(u16) = 1,
    ...
```

Let me know if you need more clarification

zhalvorsen:
This would be a similar model to the way `PlatformError` is done in https://github.com/chipsalliance/caliptra-dpe/blob/6c29cbdf13e35b942d233824c60c6c2ed31ac57c/platform/src/lib.rs#L80

leongross:
I see, thanks! What are your thoughts on the `core::error::Error` trait implementations? This would provide us with the maximum level of verbosity and enable precise error tracing when implementing `fn source(.)`, but will also increase our binary size.

leongross:
As I see it, there are two ways to add semantical meaning to the errors: Either we create (1) a enum and add well-defined errors for each use case

```rust
#[repr(u16)]
pub enum InternalErrorCode {
	CertBufferOverflow = 1,
    ...
}
```

 or (2) we add a struct and add associated constants to it with the same semantic meaning.


```rust
pub struct InternalErrorCode(u16);

impl InternalErrorCode {
    pub const fn code(self) -> u16 {
        self.0
    }

    pub const DERIVE_CTX_RESP_SLICE_OOB: Self = Self(1);
    ...
```

Since the error is an _internal_ error, I assume we could ignore outward-facing restrictions and just go for the enum.

On the other hand, adding a new internal error has become non-trivial and requires quite a lot of boilerplate code. 

The issue with just returning opaque `v as u16` values is that there is no way to connect any semantical value to them. They might change in the future or be numerically incoherent and scattered across files. Enabling backwards compatibility can also not be guaranteed.

PR #563, opened 2026-04-17
ci: fw post-build panic checking

leongross:
Built upon https://github.com/chipsalliance/caliptra-dpe/pull/540.

Build a minimal firmware image for `riscv32imc-unknown-none-elf` and analyze the ELF artifact for symbol `panic_is_possible` in a separate `panic-checker` library test.


verification/panic-check/README.md
```diff
@@ -0,0 +1,39 @@
+# panic-check
+
+Verify that the firmware image does not the possibily for panics.
```
embediver:
```suggestion
Ensure that the DPE library is free of panics by building and checking a sample firmware.
```

verification/Cargo.toml
embediver, line 1: Delete workspace and move profiles directly into crate.

embediver (changes requested):

leongross:
@clundin25 @zhalvorsen

embediver (dismissed):

crypto/Cargo.toml
```diff
@@ -11,7 +11,20 @@ publish.workspace = true
 [features]
```
zhalvorsen, line 11: What did you use to format this file? Was it a tool? It would be nice to have a more consistent style for *.toml files.

dpe/src/commands/certify_key.rs
```diff
@@ -268,7 +272,13 @@ impl CommandExecution for CertifyKeyCommand<'_> {
         let len = response.size()?;
         // Rotate handle if it isn't the default
         dpe.roll_onetime_use_handle(env, idx)?;
-        response.set_handle(env.state().contexts[idx].handle);
+        response.set_handle(
```
zhalvorsen: I would hope the compiler was smart enough to know the index was checked with the previous change. Can we put this line back and still avoid a panic? I'm mostly concerned that if this line fails. The DPE context is lost forever because the handle changed, but the caller didn't receive the new handle.

dpe/src/commands/derive_context.rs
```diff
@@ -376,19 +388,19 @@ impl CommandExecution for DeriveContextCmd {
...
                         // We either retained the parent or it has other children, so retire it and
                         // make it's handle invalid.
                         // At this point we cannot error out anymore, so it is safe to set the parent context.
-                        env.state().contexts[parent_idx] = tmp_parent_context;
+                        *env.state().contexts.get_mut(parent_idx).ok_or(DpeErrorCode::InternalError)? = tmp_parent_context;
                     }
 
                     response.handle = ContextHandle::new_invalid();
-                    response.parent_handle = env.state().contexts[parent_idx].handle;
+                    response.parent_handle = env.state().contexts.get(parent_idx).ok_or(DpeErrorCode::InternalError)?.handle;
```
zhalvorsen: I'm hoping we can avoid this one as well to avoid losing access to the context

verification/panic-check/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+std = ["anyhow", "elf"]
+p256 = ["dpe/p256"]
+p384 = ["dpe/p384"]
+ml-dsa = ["dpe/ml-dsa"]
+hybrid = ["p384", "ml-dsa"]
+# Intentionally introduces panic-able code to verify the checker detects it.
+# Build with this feature should FAIL the panic check.
+should-fail = []
+
+[[bin]]
```
zhalvorsen: I think it would make it more intuitive in use and maintenance to have `firmware` and `checker` be separate crates. A slight tangent, if the checker were a library rather than a binary, could you make this check a unit test instead of running a binary?

verification/panic-check/firmware/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+use dpe::support::Support;
+use dpe::DpeFlags;
+use dpe::DpeInstance;
+use dpe::DpeProfile;
+use platform::dummy::{DefaultPlatform, DefaultPlatformProfile};
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
```
zhalvorsen, line 27: The `caliptra-sw` tests use a system like the following which should make it easier to look for panics because you only need to search for the `panic_is_possible` symbol instead of making sure you have all possible panic symbols listed or trying to keep them up-to-date of the rust compiler changes.


```suggestion
#[panic_handler]
#[inline(never)]
fn panic_handler(_: &PanicInfo) -> ! {
    panic_is_possible();
    loop {}
}

#[no_mangle]
#[inline(never)]
fn panic_is_possible() {
    black_box(());
    // The existence of this symbol is used to inform test_panic_missing
    // that panics are possible. Do not remove or rename this symbol.
}
```

verification/panic-check/firmware/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    let mut dpe = match DpeInstance::new(&mut env, dpe_profile) {
+        Ok(dpe) => dpe,
+        Err(_) => return,
+    };
+
+    let _ = black_box(dpe.execute_serialized_command(&mut env, 1, black_box(&[0])));
+    let _ = black_box(dpe.roll_onetime_use_handle(&mut env, black_box(1)));
+    let _ = black_box(dpe.get_profile(env.platform, black_box(env.state.support)));
+    let _ = black_box(dpe.deserialize_command(black_box(&[0])));
```
zhalvorsen: Lets start with just `execute_serialized_command` we can add more later if we need to. But these 3 should all be captured within the `execute_serialized_command` call.
```suggestion
```

dpe/src/commands/derive_context.rs
```diff
@@ -167,10 +167,17 @@ impl DeriveContextCmd {
     /// * `parent_idx` - Index of the parent context.
     /// * `tci_type` - INPUT_TYPE to check for uniqueness.
     fn tci_type_is_unique_among_children(state: &State, parent_idx: usize, tci_type: u32) -> bool {
```
zhalvorsen: Can we change the API here to take the parent context by reference instead of the index? This function should never be called with an invalid parent context index, but I don't want to mask an false negative

dpe/src/validation.rs
```diff
@@ -73,7 +73,11 @@ impl DpeValidator<'_> {
             return Err(ValidationError::VersionMismatch);
         }
         for i in 0..MAX_HANDLES {
-            let context = &self.dpe.contexts[i];
+            let context = self
+                .dpe
+                .contexts
+                .get(i)
+                .ok_or(ValidationError::BadContextState)?;
```
zhalvorsen: If we use an iterator, this would be a lot cleaner:

```Rust
for (i, context) in self.dpe.contexts.iter().enumerate() {
```

dpe/src/commands/derive_context.rs
```diff
@@ -456,12 +473,20 @@ impl CommandExecution for DeriveContextCmd {
         tmp_parent_context.children = children_with_child_idx;
 
         // At this point we cannot error out anymore, so it is safe to set the updated child and parent contexts.
-        env.state().contexts[child_idx] = tmp_child_context;
-        env.state().contexts[parent_idx] = tmp_parent_context;
+        *env.state()
```
zhalvorsen: We should be able to keep these direct indexing calls as well. I'm worried for a similar reason if the child is modified and the parent fails, the caller loses access to the parent. If we use direct indexing here, the compiler makes some guarantees for us, especially with these panic checks.

dpe/src/state.rs
```diff
@@ -102,6 +102,24 @@ impl State {
         Ok(idx)
     }
 
+    /// Finds the context having `handle` in `locality` and returns both context and index
+    ///
+    /// # Arguments
+    ///
+    /// * `handle` - handle to search
+    /// * `locality` - locality to search
+    pub fn get_active_context_and_idx(
```
zhalvorsen, line 111: If this function returned a copy of the context, would it make compiler checks for `mut` and non-`mut` references easier to deal with?

crypto/Cargo.toml
leongross: Maybe we should consider adding a format check to the CI? This would enforce a consistent style.

clundin25: Possible that a bug snuck in, but the CI should be checking formatting: https://github.com/chipsalliance/caliptra-dpe/blob/main/xtask/src/main.rs#L295-L302

verification/panic-check/firmware/src/crypto_dummy.rs
```diff
@@ -0,0 +1,146 @@
+// Licensed under the Apache-2.0 license
```
clundin25, line 1: Could this & the platform implementation live in the panic check crate, to avoid bloating the "real" crates?

dpe/src/commands/certify_key.rs
```diff
@@ -151,7 +151,11 @@ impl CommandExecution for CertifyKeyCommand<'_> {
             CertifyKeyCommand::Mldsa87(cmd) => (&cmd.handle, cmd.format, cmd.label.as_slice()),
         };
         let idx = env.state().get_active_context_pos(handle, locality)?;
-        let context = env.state().contexts[idx];
+        let context = *env
+            .state()
+            .contexts
+            .get(idx)
+            .ok_or(DpeErrorCode::InternalError)?;
```
clundin25, line 158: @leongross can you add a task to use a unique inner code in InternalError? There ar e so many now that I am worried we will get hard to debug errors

crypto/Cargo.toml
zhalvorsen: I didn't think rustfmt did anything with TOML files?

dpe/src/commands/certify_key.rs
zhalvorsen: @embediver already created this https://github.com/chipsalliance/caliptra-dpe/issues/460

leongross: After this PR is finished, I can take over the implementation of #460 as well, but I propose to do it one after another to not get things mixed up here.

verification/panic-check/Cargo.toml
leongross: I agree. I refactored the code just as you proposed; the panic symbol checks are run in the checker library crate. The xtask wrapper is still in place for the `cargo xtask ci` and I modified it to execute said test, which also builds the firmware targets under test.

crypto/Cargo.toml
leongross: @zhalvorsen you are right, "cargo fmt" does not format toml files. In my development setup, I use `taplo` for toml linting and formatting. Reverting the cargo file to the old state and applying the formatting yields the same result, leading me to the conclusion that the taplo formatter is responsible for this change. 

What do you think, should we enforce a uniform style for toml as well, using `taplo` or any other formatter?

clundin25: I am okay with using `taplo`, not sure what the alternatives would be.

dpe/src/commands/certify_key.rs
leongross: Verifying this in the field using the test infrastructure and a manual inspection of the artifact shows that indeed the compiler can guarantee `idx` to be valid and no panic symbol is built into the binary so that I will revert the change.

crypto/Cargo.toml
leongross: @clundin25 I drafted a PR that integrates taplo to the other precheckin checks in `xtask` https://github.com/chipsalliance/caliptra-dpe/pull/567

dpe/src/commands/derive_context.rs
```diff
@@ -164,13 +164,21 @@ impl DeriveContextCmd {
...
-        let parent_children = state.contexts[parent_idx].children;
-        parent_children
-            .iter()
-            .all(|idx| state.contexts[idx].tci.tci_type != tci_type)
+    fn tci_type_is_unique_among_children(
+        state: &State,
+        parent_ctx: &Context,
+        tci_type: u32,
+    ) -> bool {
+        // let parent_children = parent_ctx.children;
```
zhalvorsen: Feel free to remove this commented code

dpe/src/commands/derive_context.rs
```diff
@@ -164,13 +164,21 @@ impl DeriveContextCmd {
...
+        parent_ctx.children.iter().all(|idx| {
+            state
+                .contexts
+                .get(idx)
+                .map(|ctx| ctx.tci.tci_type != tci_type)
+                .unwrap_or(false)
```
zhalvorsen: There are already checks in the `IntoIter` implementation so all children iterator usage can use the direct access without `get`

dpe/src/commands/derive_context.rs
```diff
@@ -309,14 +318,16 @@ impl CommandExecution for DeriveContextCmd {
                     // Rotate the handle if it isn't the default context.
                     dpe.roll_onetime_use_handle(env, parent_idx)?;
 
-                    env.state().contexts[parent_idx] = Context {
-                        handle: env.state().contexts[parent_idx].handle,
+                    let parent_ctx = env.state().contexts.get_mut(parent_idx).ok_or(DpeErrorCode::InternalError)?;
```
zhalvorsen: You should be able to use the direct index here. If this fails the caller will lose access to the context

zhalvorsen: This is a common problem in this PR. Maybe we should change the `roll_onetime_use_handle` implementation to take a mutable context so it will alter the temporary copy instead of the actual context. Then we can copy the full temporary context at once. Does that make sense? This could be done in a follow-on PR.

leongross:
@zhalvorsen

dpe/src/lib.rs
```diff
@@ -5,6 +5,10 @@ Abstract:
     DPE Library Crate.
 --*/
 #![cfg_attr(not(any(test, target_arch = "x86_64")), no_std)]
+#![warn(clippy::panic)]
+#![warn(clippy::indexing_slicing)]
```
zhalvorsen: I think its okay to remove this clippy warning. Using `get` can sometimes make the code harder to read and we know the times we use `[]` directly are okay because it will get caught by this new panic checker otherwise

dpe/src/x509.rs
```diff
@@ -1677,7 +1680,10 @@ impl CertWriter<'_> {
     ) -> Result<usize, DpeErrorCode> {
         let tcb_infos_size = if !measurements.tci_nodes.is_empty() {
             self.get_tcb_info_size(
-                &measurements.tci_nodes[0],
+                measurements
```
zhalvorsen: We can use `[]` directly here

dpe/src/x509.rs
```diff
@@ -2671,7 +2677,7 @@ fn get_subject_name<'a>(
     crypto.get_pubkey_serial(pub_key, subj_serial)?;
 
     // The serial number of the subject can be at most 64 bytes
-    let truncated_subj_serial = &subj_serial[..64];
+    let truncated_subj_serial = subj_serial.get(..64).ok_or(DpeErrorCode::InvalidArgument)?;
```
zhalvorsen: We can use `[]` directly here

dpe/src/x509.rs
```diff
@@ -2696,7 +2702,9 @@ fn get_tci_nodes<'a>(
     if tcb_count > MAX_HANDLES {
         return Err(DpeErrorCode::InternalError);
     }
-    Ok(&mut nodes[..tcb_count])
+    nodes
```
zhalvorsen: We can use `[]` directly here because it is checked above

dpe/src/x509.rs
```diff
@@ -2717,7 +2725,12 @@ fn get_subject_key_identifier(
         return Err(DpeErrorCode::InternalError);
     }
     // truncate key identifier to 20 bytes
-    subject_key_identifier.copy_from_slice(&hashed_pub_key.as_slice()[..MAX_KEY_IDENTIFIER_SIZE]);
+    subject_key_identifier.copy_from_slice(
```
zhalvorsen: We can use `[]` directly here

dpe/src/x509.rs
```diff
@@ -2949,7 +2962,9 @@ fn create_dpe_cert_or_csr(
                 if issuer_len > MAX_ISSUER_NAME_SIZE {
                     return Err(DpeErrorCode::InternalError);
                 }
-                &issuer_name[..issuer_len]
+                issuer_name
```
zhalvorsen: We can use `[]` directly here because it is checked above

verification/panic-check/checker/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+/// of the `&str` needles to be found in the haystack.
+/// As soon as one match is found the function returns `true`.
+/// If none of the symbols in [needles] can be found in the binary, this function
+/// returns `false`.
+pub fn elf_sym_contain_one_of(path: PathBuf, needles: Vec<&str>) -> Result<bool, anyhow::Error> {
+    let elf: Vec<u8> = fs::read(&path)?;
+    let elf = elf::ElfBytes::<LittleEndian>::minimal_parse(&elf[..])?;
+
+    if let Some((symbols, strings)) = elf.symbol_table().map_err(|e| anyhow::anyhow!("{e}"))? {
+        for sym in symbols {
```
zhalvorsen, line 16: I don't think we need to plan for more panic symbols, feel free to only use `panic_is_possible`

leongross: That's a good point. I made the ELF symbol search function private and added a public wrapper that calls this function with the panic symbol, so the API is clear.
I would prefer to keep the elf symbol searching function with 1 needle argument, so we can reliably test our test code, so that we can guarantee that this symbol search really works.

dpe/src/commands/derive_context.rs
leongross: I agree, this would make things cleaner. Once this is done, I can follow up on that.

zhalvorsen (dismissed):

zhalvorsen:
Thanks man, this is great!

dpe/src/commands/derive_context.rs
leongross: Solved in https://github.com/chipsalliance/caliptra-dpe/pull/578, starting the rebase.

leongross:
@zhalvorsen the rebase is done, and all tests pass. I think this can be merged.

zhalvorsen (approved):

Issue #502, opened 2026-01-21
Enable Miri for UB detection


embediver:
To discover undefined behavior that the rust type system cant catch (e.g. due to unsafe code) [Miri](https://github.com/rust-lang/miri) can be used for further analysis.

Unfortunately Miri can't verify some special operations.
Running Miri on the DPE Crate, isn't working out of the box.
Following are the Problems I discovered so far:
- The `constant_time_eq` dependency is unsupported, since it relies on _inline assembly_.
  The only use of this seems to be the equality check for two context handles.
- Miri utilizes the unit tests to verify the code, since some of them rely on FFI bindings to OpenSSL, which isn't supported either.

The unit tests that utilize OpenSSL could be excluded from verification.
The `constant_time_eq` needs to be updated, since a newer version is miri compatible.

PR #564, opened 2026-04-23
ci: add miri execution

leongross:
Follow-up on #502.

Related:
- https://github.com/chipsalliance/caliptra-cfi/pull/22


.github/workflows/ci.yml
```diff
@@ -39,3 +39,9 @@ jobs:
         run: |
           export NIX_REMOTE=local
           nix develop --command cargo xtask ci
```
clundin25: Could we add a `--miri` flag that adds it to the CI test path? Then we keep a single entry point for the CI.

dpe/src/commands/certify_key.rs
```diff
@@ -20,7 +20,8 @@ use cfg_if::cfg_if;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 #[cfg(not(feature = "disable_x509"))]
-#[repr(C)]
+// miri alignment: align(4) ensures zerocopy can safely reference from byte slices
```
clundin25: nit: I think we can skip the comment

dpe/src/commands/certify_key.rs
```diff
@@ -468,16 +473,19 @@ mod tests {
 
     #[test]
     fn test_deserialize_certify_key() {
```
clundin25, line 479: Is niri scanning the test code too? Is it possible to avoid that?

clundin25: Hmm maybe that is incorrect, since the test code is the harness for niri

xtask/src/main.rs
```diff
@@ -357,6 +376,43 @@ fn test_rust_targets(profile: &str) -> Result<()> {
     Ok(())
 }
 
+fn run_miri_target(profile: &str, args: &MiriArgs) -> Result<()> {
+    // Note: CFI feature is excluded because it uses inline assembly which Miri does not support
+    let opts = CargoOptions::with_features("dpe/Cargo.toml", &format!("{},cfi", profile));
```
clundin25, line 386: Isn't this including CFI?

Cargo.toml
```diff
@@ -25,8 +25,8 @@ anyhow = "1.0.70"
 arrayvec = { version = "0.7.4", default-features = false, features = [
   "zeroize",
 ] }
-caliptra-cfi-derive = { git = "https://github.com/chipsalliance/caliptra-cfi.git", package = "caliptra-cfi-derive", rev = "aa96c53073ce7f298f7482c1ac57f5f65ffc115e" }
```
clundin25, line 30: We can't commit this so first we'd need the CFI PR to land

xtask/src/main.rs
leongross: Yes, I use my `caliptra-cfi` fork. This comment is outdated, I will remove it.

.github/workflows/ci.yml
leongross: I am not sure if I understand what you mean. Would you like me to add a `--miri` flag to the CI subcommand so that we can enable/disable it for the CI subcommand?

dpe/src/commands/certify_key.rs
leongross: What exactly do you mean by "scanning test code"? Miri lifts the rust code into an IR and then interprets it to find UB for all the tests that we define. It is not a static analysis tool that analyses all the Rust code. To be nitty, this also means it can only find UB in our code when it is tested, so UB detection relies on high test coverage, iirc.

.github/workflows/ci.yml
leongross: See ff2aa8b

Cargo.toml
leongross: https://github.com/chipsalliance/caliptra-cfi/pull/22 was merged, so we can consider this unblocked.

leongross: @clundin25 @zhalvorsen since the CFI is merged and the CI for the commit updating the dependency succeeded (52459e330f303d85b42338ba142d761ccc66a515), can we get this merged?

.github/workflows/ci.yml
```diff
@@ -38,4 +38,4 @@ jobs:
       - name: Run CI tests
         run: |
           export NIX_REMOTE=local
-          nix develop --command cargo xtask ci
+          nix develop --command cargo xtask ci --miri
```
clundin25: Let's make this a separate nightly job, since it's so slow.

failed
clundin25, line 1: Did you mean to commit this?

Cargo.toml
```diff
@@ -27,8 +27,8 @@ anyhow = "1.0.70"
...
-caliptra-cfi-lib = { git = "https://github.com/chipsalliance/caliptra-cfi.git", package = "caliptra-cfi-lib", rev = "aa96c53073ce7f298f7482c1ac57f5f65ffc115e", default-features = false }
+caliptra-cfi-derive = { git = "https://github.com/chipsalliance/caliptra-cfi.git", package = "caliptra-cfi-derive", rev = "72c75dc70cde4120a9ce149d4f5cb21621399958" }
```
clundin25, line 30: Do you mind sending a PR to caliptra-sw to keep the CFI versions aligned

xtask/src/main.rs
```diff
@@ -120,12 +140,20 @@ fn run_ci() -> Result<()> {
...
     run_panic_checks()?;
 
+    if args.miri {
+        let miri_args = &MiriArgs {
+            nextest: true,
+            nthreads: num_cpus::get(),
+        };
+
+        run_miri_tests(miri_args)?;
+    }
```
clundin25: I guess we can drop this from the `CI` command, since this should run standalone

xtask/src/main.rs
```diff
@@ -7,6 +7,8 @@ use std::process::Command;
 
 // TODO(clundin): Support "hybrid"
 const PROFILES: &[&str] = &["ml-dsa", "p256", "p384"];
+/// Pinned nightly toolchain version. Must match the version in flake.nix shellHook.
+const NIGHTLY: &str = "nightly-2025-07-08";
```
clundin25, line 11: We should pull the fuzzers in with nix so we can simplify this code. I can send a quick PR

leongross: Thanks for taking care of that.

failed
leongross: No, I fumbled there, will remove.

Cargo.toml
leongross: https://github.com/chipsalliance/caliptra-sw/pull/3874

clundin25 (approved):
LGTM - Awesome work!

PR #567, opened 2026-04-30
ci: add taplo as toml formatter to xtask

leongross:
As discussed in #563 there is no enforcement for toml formatting yet. This PR adds toml format checking to `cargo xtask precheckin` that throws an error when non-conforming changes are detected and prints the command to automatically fix the issues.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -303,8 +303,6 @@ fn format_rust_targets() -> Result<()> {
 }
 
 fn format_toml_targets() -> Result<()> {
-    println!("Installing taplo");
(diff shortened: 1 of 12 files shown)


xtask/src/main.rs
```diff
@@ -301,6 +302,28 @@ fn format_rust_targets() -> Result<()> {
     Ok(())
 }
 
+fn format_toml_targets() -> Result<()> {
+    println!("Installing taplo");
+    cargo().args(["install", "taplo-cli", "--locked"]).run()?;
```
clundin25: Let's assume it's installed instead of embedding an installer script. Stick in the nix developer flake & we should be good

clundin25 (approved):

PR #568, opened 2026-04-30
dpe: Introduce size/encode helpers

ternst-nv:
Add size and encode helper functions to the dpe x509 encoder.  This logic is repeated for a number of sub-blocks and thus introducing the helper functions results in a code instruction size savings of ~1200bytes on caliptra-sw caliptra-1.x branch.


zhalvorsen (approved):
This is great, thanks!

PR #577, opened 2026-05-20
dpe: Update cfi library dependency

ternst-nv:
Update the CFI library dependency to the most recent main.  This can then be updated in the calling `caliptra-1.x` branch to allow the same CFI symbols to be used throughout the code base, reducing overall code usage.

### DIFF
--- a/crypto/src/openssl.rs
+++ b/crypto/src/openssl.rs
@@ -5,7 +5,7 @@ use crate::{
     MAX_EXPORTED_CDI_SIZE,
 };
 #[cfg(not(feature = "no-cfi"))]
-use caliptra_cfi_derive_git::cfi_impl_fn;
+use caliptra_cfi_derive::cfi_impl_fn;
 use openssl::{
     bn::{BigNum, BigNumContext},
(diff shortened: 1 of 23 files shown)


crypto/src/openssl.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
     MAX_EXPORTED_CDI_SIZE,
 };
 #[cfg(not(feature = "no-cfi"))]
```
zhalvorsen: When Carl, did this upgrade to the `main` branch he also switched the feature flag to `cfi` instead of `no-cfi` so it is more readable (https://github.com/chipsalliance/caliptra-dpe/pull/531). Do you want to do that as part of this change? Or we can ignore it for now and jump to the other flag when we switch `caliptra-sw` to point to `main`

zhalvorsen (dismissed):

crypto/src/openssl.rs
ternst-nv: Thats a good point.  I'll make that change in this PR as well, thanks!

zhalvorsen (approved):

PR #579, opened 2026-05-21
dpe: store profile as a raw u32

pjhades:
caliptra-sw stores DpeInstance inside the PersistentData, which is expected to be FromZeros. However, the only member of DpeInstance, profile, refers to a DpeProfile, which is an enum for which 0 is not a valid variant.

This commit stores the DPE profile as a raw u32 instead of the DpeProfile enum, so that we can derive FromZeros on it. It also adds a `profile` method as the public interface to fetch the profile, which will return an error if the raw u32 value doesn't represent a valid profile.

### DIFF
--- a/dpe/src/lib.rs
+++ b/dpe/src/lib.rs
@@ -83,17 +83,22 @@ impl From<bool> for U8Bool {
     }
 }
 
+pub const DPE_PROFILE_SHA256: u32 = 3;
+pub const DPE_PROFILE_SHA384: u32 = 4;
+#[cfg(feature = "ml-dsa")]
+pub const DPE_PROFILE_MLDSA87: u32 = 5;
+
 #[derive(
     Copy, Clone, Debug, PartialEq, Eq, IntoBytes, TryFromBytes, KnownLayout, Immutable, Zeroize,
 )]
 #[repr(u32)]
 pub enum DpeProfile {
     // Note: Min profiles (1 & 2) are not supported by this implementation
-    P256Sha256 = 3,
-    P384Sha384 = 4,
+    P256Sha256 = DPE_PROFILE_SHA256,
+    P384Sha384 = DPE_PROFILE_SHA384,
     #[cfg(feature = "ml-dsa")]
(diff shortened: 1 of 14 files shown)


dpe/src/lib.rs
```diff
@@ -83,17 +83,22 @@ impl From<bool> for U8Bool {
...
-    P256Sha256 = 3,
-    P384Sha384 = 4,
+    P256Sha256 = DPE_PROFILE_SHA256,
+    P384Sha384 = DPE_PROFILE_SHA384,
     #[cfg(feature = "ml-dsa")]
-    Mldsa87 = 5, // TODO(clundin): Added this to get past compiler / feature flags. We
-                 // will want a real solution here.
+    Mldsa87 = DPE_PROFILE_MLDSA87, // TODO(clundin): Added this to get past compiler / feature flags. We
+                                   // will want a real solution here.
 }
```
zhalvorsen, line 102: Its probably worth while to have an `TryFrom` implementation to convert a u32 to `DpeProfile`. It could probably be used elsewhere. `DpeInstance::profile` would then call `try_from` in the implementation

zhalvorsen (dismissed):

dpe/src/lib.rs
pjhades: Good catch. Added in [205ec33](https://github.com/chipsalliance/caliptra-dpe/pull/579/commits/205ec33fcaf13d1346aacef893a44e73160d8339).

dpe/src/lib.rs
```diff
@@ -140,6 +145,20 @@ impl From<DpeProfile> for u32 {
     }
 }
 
+impl TryFrom<u32> for DpeProfile {
```
ternst-nv, line 148: [nit] Looks like we have good coverage on the happy path, but might be good to add a `0` test.

ternst-nv (approved):
Small test nit otherwise LGTM

pjhades:
I'm closing this one since it's hacky-ish and I've been working on upgrading `caliptra-sw/caliptra-1.x` to use this crate.

PR #580, opened 2026-05-28
Improved error handling

### LINKED ISSUE #314: Align errors with the spec
Some error codes here do not align with the DPE spec

### LINKED ISSUE #460: Improved error handling

leongross:
Resolves #460 and #314 .


crypto/src/lib.rs
```diff
@@ -112,22 +112,28 @@ impl CryptoError {
             Self::MismatchedAlgorithm => 0x9,
         }
     }
+}
 
-    pub fn get_error_detail(&self) -> Option<u32> {
```
zhalvorsen, line 116: As we discussed offline, `caliptra-sw` uses this function to populate extra error reporting registers, so we will need to keep this function.

leongross:
While we're at it, should we address https://github.com/chipsalliance/caliptra-dpe/issues/314 as well if it is still an issue? @jhand2 @zhalvorsen.

Are you referring to _5.9.4 Command and Response Headers_ in the [TCG DPE spec ](https://trustedcomputinggroup.org/wp-content/uploads/TCG-DICE-Protection-Environment-Specification_14february2023-1.pdf)?

zhalvorsen:
I'm pretty sure they still need to be aligned. This would be good to get right. [This](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Protection-Environment-Version-1.0_pub.pdf) should be the specification.

leongross:
> I'm pretty sure they still need to be aligned. This would be good to get right. [This](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Protection-Environment-Version-1.0_pub.pdf) should be the specification.

## Design

The TCG DPE Spec defines a set of DPE Errors, which calipra-dpe currently violates. I suspect that the spec expanded, and now there is a mismatch.
 

| ErrCode | DpeSpec | Caliptra-DPE|
|---|---|---|
| 0 | No error | **NoError** |
| 1 | Internal Error | **InternalError(u16)** |
| 2 | Invalid Command | **InvalidCommand** | 
| 3 | Invalid Argument | **InvalidArgument** | 
| 4 | Session Exhausted | ArgumentNotSupported | 
| 5 | Initialization Seed Locked | x509CsrUnset | 
| 6 | Out of Memory | x509InvalidState | 
| 7 | Cancelled Command | x509SkipsExhausted  | 

As I read the spec, there is no explicit mechanism for vendor-defined error codes, but defining error codes that do not collide with `$dpe-error` (Section 5.11) does not violate it either.

I am unsure about the current development history, but it seems as if the X. 509 errors might have been defined at a point in time where the range of well-defined errors was smaller. So, adding error codes after the last well-defined error in the spec may lead us to fixing the same error pattern over and over again.

My proposal is to add a large enough constant offset to the start of our custom/vendor-defined error codes, e.g. `0x2000`.

## Compatability

As I see it, comopnents issuing dpe commands (calipra-sw and libcaliptra, calipra-mcu-sw) do not parse the error codes themselves really (at least I could not find the corresponding code locations) and hence SoC code interfacing with the caliptra components might have to take care of that. I can't make an assumption on how this might be done atm, but semantically changing the error codes here has an impact to the entire stack and this must be considered as well.

platform/src/lib.rs
```diff
@@ -98,24 +101,42 @@ impl PlatformError {
         // field, so we can read the discriminant without offsetting the pointer.
         unsafe { *<*const _>::from(self).cast::<u16>() }
     }
+}
 
-    pub fn get_error_detail(&self) -> Option<u32> {
```
zhalvorsen, line 102: We will need this function as well

zhalvorsen (dismissed):
This is great! Thanks @leongross!

zhalvorsen:
> My proposal is to add a large enough constant offset to the start of our custom/vendor-defined error codes, e.g. 0x2000.

Another source I forgot to mention is the DPE profile specification we are supposed to be aligning with as well. https://github.com/jhand2/Security/blob/main/specifications/dpe-irot-profile/spec.ocp#L1211 starts vendor defined error codes at 0x80 which I think should be enough of a gap.

I think its best to align with the specifications rather than ensure backwards compatibility with these error codes.

dpe/src/error.rs
```diff
@@ -0,0 +1,285 @@
...
+/// For the definition of required DPE Error Codes, consult the
+/// [TCG DPE Specification](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Protection-Environment-Version-1.0_pub.pdf), Section 5.11.
+///
+/// This base definition is extended by a set of useful vendor errors.
+///
+/// # Custom Definitions
+/// The DICE spec (v.1.0) only covers errors 0-7 as well-defined.
+/// Further, there is no mentioning of vendor defined errors.
+/// The `OCP Security Project` proposes a [numerical range](https://github.com/opencomputeproject/Security/blob/main/specifications/dpe-irot-profile/spec.ocp#L1391) for these errors, to which we adhere.
+pub enum DpeStatus {
```
zhalvorsen: Maybe for this review to minimize the differences we should keep this `DpeErrorCode`

zhalvorsen: nit: This would probably be better at the top of the file for discoverability. I would guess this is the main reason we will mostly refer to this file.

dpe/src/error.rs
```diff
@@ -0,0 +1,285 @@
...
+    OutOfMemory = 6,
+    CancelledCommand = 7,
+    // OCP Spec "Server iRoT Profile for DPE", v0.13.0
+    InvalidHandle = 0x80,
+    InvalidLocality = 0x81,
+    HandleDefined = 0x82,
+    ArgumentNotSupported = 0x83,
+    AlreadyInitialized = 0x84,
+    InvalidParentLocality = 0x85,
+    // The following errors, aren't neither defined in the OCP Security spec,
```
zhalvorsen: typo
```suggestion
    // The following errors, are neither defined in the OCP Security spec,
```

dpe/src/error.rs
```diff
@@ -0,0 +1,285 @@
...
+    X509InvalidState = 0x87,
+    X509SkipsExhausted = 0x88,
+    X509InvalidWidth = 0x89,
+    X509AlgorithmMismatch = 0x90,
+    MaxTcis = 0x91,
+    InvalidMutRefBuf = 0x92,
+    InvalidResponseBuf = 0x93,
+    UninitializedResponseHeader = 0x94,
+    /// Returned by UpdateContextMeasurement when PARENT_CONTEXT_HANDLE doesn't
+    /// exist in the caller's locality. Value matches the OCP iROT profile spec (0x85).
```
zhalvorsen: This comment looks out of place? We can probably remove it

dpe/src/error.rs
```diff
@@ -0,0 +1,285 @@
...
+    // OCP Spec "Server iRoT Profile for DPE", v0.13.0
+    InvalidHandle = 0x80,
+    InvalidLocality = 0x81,
+    HandleDefined = 0x82,
+    ArgumentNotSupported = 0x83,
+    AlreadyInitialized = 0x84,
+    InvalidParentLocality = 0x85,
+    // The following errors, aren't neither defined in the OCP Security spec,
+    // nor the TCG spec. It is unclear why the original values were chosen the
+    // way there were. For coherence, we continue enumeration on 0x80 + e.
```
zhalvorsen: nit: this comment might become stale pretty quickly. Maybe drop the part about the legacy values?

zhalvorsen (approved):

PR #581, opened 2026-05-28
dpe: optimize stack usage

pjhades:
This commit introduces two optimizations to reduce stack usage:

1. In `DpeInstance::new`, call `execute_serialized` instead of `execute` on the mailbox command, to avoid creating a big stack frame for `execute`, since `execute` will create a on-stack buffer to hold a generic `Response`, which is much bigger than the response we need - a `NewHandleResp`.

2. In `DpeInstance::new_auto_init`, avoid creating a mutable `Context` on the stack, but passing a `&mut Context` to have `dpe.add_tci_measurement` modify the context in-place. Changes are made on the interface of `add_tci_measurement` as well to avoid hitting compile errors.

### DIFF
--- a/dpe/src/dpe_instance.rs
+++ b/dpe/src/dpe_instance.rs
@@ -9,7 +9,7 @@ use crate::INTERNAL_INPUT_INFO_SIZE;
 use crate::{
     commands::{Command, CommandExecution, CommandHdr, InitCtxCmd},
     context::{ChildToRootIter, Context, ContextHandle, ContextState},
-    response::{DpeErrorCode, GetProfileResp, Response, ResponseHdr},
+    response::{DpeErrorCode, GetProfileResp, NewHandleResp, Response, ResponseHdr},
     support::Support,
     tci::TciMeasurement,
     DpeProfile, State, MAX_HANDLES,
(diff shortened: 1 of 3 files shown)


dpe/src/dpe_instance.rs
```diff
@@ -114,14 +115,16 @@ impl DpeInstance {
...
-            .state()
-            .get_active_context_pos(&ContextHandle::default(), locality)?;
-        let mut tmp_context = env.state().contexts[idx];
+        let (crypto, _, state) = env.get();
+        let idx = state.get_active_context_pos(&ContextHandle::default(), locality)?;
         // add measurement to auto-initialized context
-        dpe.add_tci_measurement(env, &mut tmp_context, auto_init_measurement, locality)?;
-        env.state().contexts[idx] = tmp_context;
-        env.state().contexts[idx].tci.tci_type = tci_type;
+        dpe.add_tci_measurement(
```
zhalvorsen, line 121: Just a side note here. The main reason behind the previous implementation here was to avoid altering the state of DPE in case of a failure, but looking at the implementation it should never alter the state on error paths. So this change is fine.

zhalvorsen (approved):

dpe/src/dpe_instance.rs
pjhades: Yeah, I checked that before making the change. Thanks for calling it out.

PR #585, opened 2026-06-09
x509: Unify signed payload

ternst-nv:
Unify the signed payload logic for both certificates and csrs.  This reduces some code duplication, saving ~350 bytes from the end library utilized by the `caliptra-sw` software.

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -2362,43 +2362,100 @@ impl CertWriter<'_> {
         measurements: &MeasurementData,
         validity: &CertValidity,
     ) -> Result<usize, DpeErrorCode> {
+        self.encode_signed_payload(
+            sign_cb,
+            SignedPayload::Tbs {
+                serial_number,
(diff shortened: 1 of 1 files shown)


zhalvorsen (approved):

dpe/src/x509.rs
clundin25, line 2458: Is the CSR case still validated somewhere?

ternst-nv: Yupp, there are `encode_cms` tests which validate the underlying function.

clundin25: Sorry I meant regarding the backtrack count check

ternst-nv: Thanks for asking about this, I took a deeper look and two things:

1. There is already a check for this in `encode_cms`.
2. There isn't a unit test verifying that functionality either.

I'll open a follow on PR which both removes this superflous check (since this isn't a public API), and adds a test for the given case.

PR #588, opened 2026-06-10
x509: truncate request

ternst-nv:
Switch the error profile of the x509 encoder to fail at the end of the encoding if the buffer would be exceeded instead of failing immediately when the condition is detected.  This allows a significant reduction of error machinery, reducing the instruction memory used by the caliptra core firmware on `caliptra-1.x` by ~3.5Kb.

Safe operations are still used for all encoding operations over the buffer, so there is no possibility of buffer overflow.

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -603,24 +603,24 @@ impl CertWriter<'_> {
         oid: &[u8],
         critical: Option<u8>,
         content_size: usize,
-    ) -> Result<usize, DpeErrorCode> {
+    ) -> usize {
         let total_size = Self::get_extension_size(
             oid,
             critical.is_some(),
             content_size,
             /*tagged=*/ false,
         );
-        let mut bytes_written = self.encode_byte(Self::SEQUENCE_TAG)?;
-        bytes_written += self.encode_size_field(total_size)?;
-        bytes_written += self.encode_oid(oid)?;
+        let mut bytes_written = self.encode_byte(Self::SEQUENCE_TAG);
+        bytes_written += self.encode_size_field(total_size);
+        bytes_written += self.encode_oid(oid);
         if let Some(crit_val) = critical {
-            bytes_written += self.encode_byte(Self::BOOL_TAG)?;
-            bytes_written += self.encode_size_field(Self::BOOL_SIZE)?;
-            bytes_written += self.encode_byte(crit_val)?;
+            bytes_written += self.encode_byte(Self::BOOL_TAG);
+            bytes_written += self.encode_size_field(Self::BOOL_SIZE);
+            bytes_written += self.encode_byte(crit_val);
         }
-        bytes_written += self.encode_byte(Self::OCTET_STRING_TAG)?;
-        bytes_written += self.encode_size_field(content_size)?;
-        Ok(bytes_written)
+        bytes_written += self.encode_byte(Self::OCTET_STRING_TAG);
+        bytes_written += self.encode_size_field(content_size);
+        bytes_written
     }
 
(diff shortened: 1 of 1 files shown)


dpe/src/x509.rs
```diff
@@ -1506,35 +1501,35 @@ impl CertWriter<'_> {
         sig: &MldsaSignature,
     ) -> Result<usize, DpeErrorCode> {
```
zhalvorsen: Looks like this `Result` can be dropped too

dpe/src/x509.rs
```diff
@@ -1925,32 +1896,32 @@ impl CertWriter<'_> {
...
+        bytes_written += self.encode_byte(Self::SEQUENCE_OF_TAG);
         bytes_written += self.encode_size_field(self.get_extensions_size(
             measurements,
             /*tagged=*/ false,
             /*explicit=*/ false,
             is_x509,
-        )?)?;
+        )?);
 
         bytes_written += self.encode_multi_tcb_info(measurements)?;
```
pjhades, line 1914: Why isn't this function refactored? Is it because it calls various `Self::get_...` methods that return a result?

dpe/src/x509.rs
```diff
@@ -1056,113 +1056,118 @@ impl CertWriter<'_> {
...
         }
+        size_width
+    }
 
-        Ok(size_width)
+    /// The unchecked write primitives advance `offset` even when a write is
+    /// skipped (buffer full), so once a structure is fully encoded an `offset`
+    /// past the buffer end means the output was silently truncated. Public
+    /// encoders call this before returning so a too-small buffer fails loudly.
+    fn check_not_truncated(&self) -> Result<(), DpeErrorCode> {
```
pjhades, line 1114: I'm wondering if there's a way to enforce the invocation of this function at the end of the public `fn`s, in case any future maintainer forgets to do so. Maybe wrapping the public `fn`s in a macro? I'm not sure if you feel it's worth it.

pjhades (dismissed):
LGTM, left a few comments.

dpe/src/x509.rs
ternst-nv: This is tied back to the `hash_oid` method, which currently can still error.  @zhalvorsen and I have been discussing methods of removing this as well in a future PR.

swenson (dismissed):
LGTM; great find

dpe/src/x509.rs
ternst-nv: Yeah, I've been thinking on that one as well.  I'm wondering if a refactor PR, which has only the `pub fn` in a block/file to emphasize the necessity would work as well?  I worry a macro would have the same issue as the function call, in that new methods would need to remember it.

I don't think there is a way to have a macro on the impl itsefl to guarantee this.

On the plus side, any new public function should include overly large encoding unit tests, which should cover this.

clundin25: Could you have an intermediate type `CertWriterUnchecked` that has either an `unsafe` API or private API, and only surface a `CertWriter` to external modules that validates this invariant?

ternst-nv: Yeah, thats kinda what I was thinking a `CertWriterInternal`, since we don't need any unsafe, it would just return an incomplete buffer.

Then the requirement is that `CertWriter` would use `check_not_truncated` on any public method.

clundin25 (dismissed):
Very nice optimization!

zhalvorsen (dismissed):

dpe/src/x509.rs
zhalvorsen: Yeah, I like that

zhalvorsen (approved):

PR #590, opened 2026-06-11
x509: Split internal and public functions

ternst-nv:
Split the internal and public facing encode functions.  This reduces the burden for reading the code base, and also emphasizes the CertWriter must use `check_for_truncated` to ensure encodings are valid and not truncated prior to returning to the user.

### DIFF
--- a/dpe/src/x509/internal.rs
+++ b/dpe/src/x509/internal.rs
@@ -1,0 +1,3304 @@
+// Licensed under the Apache-2.0 license
+
+//! Lightweight X.509 encoding routines for DPE
+//!
+//! DPE requires encoding variable-length certificates. This module provides
+//! this functionality for a no_std environment.
+
+use crate::{
+    context::ContextHandle,
+    dpe_instance::DpeEnv,
+    okref,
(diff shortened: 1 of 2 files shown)


dpe/src/x509/internal.rs
zhalvorsen, line 1: This might be really nit-picky, but would it be possible to break this into multiple commits within the review? That might make it easier to consume. For example, in the first commit move the file to `dpe/src/x509/internal.rs` unmodified, then the second commit creates a new file at `dpe/src/x509.rs` and makes the changes to both files. This would make reviewing the `internal.rs` file a lot easier.

ternst-nv: Yeah, I think I'm going to close this PR in general.  While going through the code I found a location where the translation dropped a field in the `encode_signer_info` code block.

I'm going to revisit this PR after adding some additional validation tests to the package to ensure there are no regressions.

zhalvorsen: Sounds good, thanks!

PR #593, opened 2026-06-15
dpe: reduce stack usage by computing TCI nodes on the fly

pjhades:
Enhance the `ChildToRootIter` type such that it supports traversing the TCI nodes in both directions. Avoid storing TCI data for `MAX_HANDLES` nodes on the stack but instead computing the TCI nodes information by using the iterator on the fly.

This will save `size_of::<TciNodeData>() * MAX_HANDLES`, which is `(48*2 + 4*3) * 64 == 6912` bytes from the stack when ML-DSA is enabled. The cost is the time spent on maintaining the iterator's internal state. But given the maximum node is 64 it should be acceptable compared to the reduction of stack usage.

### DIFF
--- a/dpe/src/context.rs
+++ b/dpe/src/context.rs
@@ -629,6 +629,11 @@ mod tests {
 
     #[test]
     fn test_child_to_root_iter_backwards() {
+        // This creates a chain of nodes:
+        //
+        //     ROOT <- 0 <- 1 <- 2 <- 3 <- 4
+        //
+        // where <- points to the parent.
         let mut contexts = [CONTEXT_INITIALIZER; MAX_HANDLES];
         let mut parent_idx = Context::ROOT_INDEX;
         let leaf_idx = 4;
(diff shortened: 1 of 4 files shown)


dpe/src/context.rs
```diff
@@ -489,49 +594,66 @@ mod tests {
...
-            i += 1;
-            // fail test if we iterate over all nodes without terminating, meaning we are in infinite loop
-            if i > MAX_HANDLES {
-                panic!("child to root iterator loops without termination")
+        let iter = ChildToRootIter::new(30, &contexts);
+        assert!(matches!(iter, Err(DpeErrorCode::InvalidHandle)));
+    }
+
+    #[test]
+    fn test_child_to_root_iter_backwards() {
```
ternst-nv, line 631: [nit] Might be good to include a test case with multiple children, some of whom are not on the path.

ternst-nv (commented):
LGTM, good find!

zhalvorsen (approved):
Great find!

pjhades:
Thank you! Can any of you merge it please? I don't have push permission here.

PR #596, opened 2026-06-16
Optimize MLDSA CSR stack usage

parvathib:
Add a ML-DSA-specific CertifyKey CSR path that writes the derived public key and signatures into caller-provided output buffers instead of returning large PubKey and Signature enum values by value.

This avoids stack-resident ML-DSA public key/signature temporaries in the CSR path while preserving the existing generic path for other profiles and certificate formats.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -462,11 +462,91 @@ pub trait CryptoSuite: Crypto + SignatureType + DigestType {
 }
 
 pub trait Signer {
-    /// Sign `data` with the derived private key
-    fn sign(&mut self, data: &SignData) -> Result<Signature, CryptoError>;
+    /// Sign `data` with the derived private key into a caller-provided byte slice.
+    /// For ECDSA, writes raw r||s bytes (64 bytes for P-256, 96 bytes for P-384).
+    /// For ML-DSA, writes raw signature bytes (4627 bytes for ML-DSA-87).
+    /// The `pub_key` parameter is used by implementations that need the public key
+    /// for post-sign verification (e.g., ML-DSA hardware).
+    fn sign_into_slice(
+        &mut self,
+        data: &SignData,
+        pub_key: Option<&[u8]>,
+        signature: &mut [u8],
+    ) -> Result<(), CryptoError>;
 
-    /// Get the public key associated with the derived key-pair
-    fn public_key(&mut self) -> Result<PubKey, CryptoError>;
+    /// Sign `data` with the derived private key, returning the signature by value.
+    /// Default implementation calls `sign_into_slice`.
+    fn sign(&mut self, data: &SignData) -> Result<Signature, CryptoError> {
+        match self.signature_alg() {
+            SignatureAlgorithm::Ecdsa(alg) => {
+                let curve_size = alg.curve_size();
+                let mut buf = [0u8; ecdsa::curve_384::CURVE_SIZE * 2];
+                self.sign_into_slice(data, None, &mut buf[..curve_size * 2])?;
+                let (r, s) = buf[..curve_size * 2].split_at(curve_size);
+                Ok(Signature::Ecdsa(match alg {
+                    ecdsa::EcdsaAlgorithm::Bit256 => {
+                        ecdsa::EcdsaSignature::Ecdsa256(ecdsa::EcdsaSig::from_slice(
+                            r.try_into().map_err(|_| CryptoError::Size)?,
+                            s.try_into().map_err(|_| CryptoError::Size)?,
+                        ))
+                    }
+                    ecdsa::EcdsaAlgorithm::Bit384 => {
+                        ecdsa::EcdsaSignature::Ecdsa384(ecdsa::EcdsaSig::from_slice(
+                            r.try_into().map_err(|_| CryptoError::Size)?,
+                            s.try_into().map_err(|_| CryptoError::Size)?,
+                        ))
+                    }
+                }))
+            }
+            #[cfg(feature = "ml-dsa")]
+            SignatureAlgorithm::Mldsa(_) => {
+                let mut buf = [0u8; ml_dsa::MldsaAlgorithm::Mldsa87.signature_size()];
+                self.sign_into_slice(data, None, &mut buf)?;
+                Ok(Signature::Mldsa(ml_dsa::MldsaSignature(buf)))
+            }
+        }
+    }
+
+    /// Get the public key associated with the derived key-pair into a caller-provided byte slice.
+    /// For ECDSA, writes raw x||y bytes (64 bytes for P-256, 96 bytes for P-384).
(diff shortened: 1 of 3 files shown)


crypto/src/lib.rs
```diff
@@ -465,8 +465,67 @@ pub trait Signer {
     /// Sign `data` with the derived private key
     fn sign(&mut self, data: &SignData) -> Result<Signature, CryptoError>;
 
+    /// Sign `data` with the derived private key into a caller-provided byte slice.
+    /// For ECDSA, writes raw r||s bytes (64 bytes for P-256, 96 bytes for P-384).
+    /// For ML-DSA, writes raw signature bytes (4627 bytes for ML-DSA-87).
+    fn sign_into_slice(
+        &mut self,
+        data: &SignData,
+        _pub_key: Option<&[u8]>,
```
zhalvorsen: Should we drop this parameter if it isn't used? Or is it used by the `caliptra-sw` implementation?

crypto/src/lib.rs
```diff
@@ -465,8 +465,67 @@ pub trait Signer {
...
+        _signature: &mut [u8],
```
zhalvorsen: Let's drop the preceeding `_` because the parameter is used.
```suggestion
        signature: &mut [u8],
```

crypto/src/lib.rs
```diff
@@ -465,8 +465,67 @@ pub trait Signer {
...
+    ) -> Result<(), CryptoError> {
+        let sig = self.sign(data)?;
```
zhalvorsen: Does this actually help the stack issue? This would simply allocate the signature here on the stack instead of at the call site. I would recommend swapping the dependency order. In other words have `sign` call `sign_into_slice` instead of `sign_into_slice` calling `sign.

crypto/src/lib.rs
```diff
@@ -465,8 +465,67 @@ pub trait Signer {
...
+        }
+    }
+
     /// Get the public key associated with the derived key-pair
     fn public_key(&mut self) -> Result<PubKey, CryptoError>;
+
+    /// Get the public key associated with the derived key-pair into a caller-provided byte slice.
+    /// For ECDSA, writes raw x||y bytes (64 bytes for P-256, 96 bytes for P-384).
+    /// For ML-DSA, writes raw public key bytes (2592 bytes for ML-DSA-87).
+    fn public_key_into(&mut self, _pub_key: &mut [u8]) -> Result<(), CryptoError> {
```
zhalvorsen: Let's drop the preceeding underscore here too:

```suggestion
    fn public_key_into(&mut self, pub_key: &mut [u8]) -> Result<(), CryptoError> {
```

crypto/src/lib.rs
```diff
@@ -465,8 +465,67 @@ pub trait Signer {
...
+        let key = self.public_key()?;
```
zhalvorsen: I'd swap the dependencies here for a similar reason

dpe/src/commands/certify_key.rs
```diff
@@ -197,6 +197,64 @@ impl CommandExecution for CertifyKeyCommand<'_> {
         };
 
         let mut response = self.response_bytes(dpe.profile, out)?;
+
+        // For CSR format, use optimized buffer-backed path for all profiles
+        if format == Self::FORMAT_CSR {
```
zhalvorsen: Is there a specific reason this shouldn't be done for the X509 certificate path as well? It should simplify the logic here

dpe/src/x509.rs
```diff
@@ -32,7 +32,7 @@ use caliptra_dpe_platform::{
 use zerocopy::IntoBytes;
 
 #[cfg(feature = "ml-dsa")]
-use caliptra_dpe_crypto::ml_dsa::{MldsaPublicKey, MldsaSignature};
+use caliptra_dpe_crypto::ml_dsa::{MldsaAlgorithm, MldsaPublicKey, MldsaSignature};
```
zhalvorsen: Would it be possible to break the certificate/CSR encoding to a separate PR?

crypto/src/lib.rs
parvathib: Yes, it's used by the `caliptra-sw` implementation. Pubkey is needed for verify-after-sign check. Removed the underscore to make that clearer.

crypto/src/lib.rs
parvathib: Done.

crypto/src/lib.rs
parvathib: Good point - swapped.  `sign_into_slice` and `public_key_into` are now the required methods. `sign` and `public_key`  have default implementations that call them. Added `signature_alg()` method so the defaults can construct the correct variant.

crypto/src/lib.rs
parvathib: done

crypto/src/lib.rs
parvathib: Done. Same as the sign path, `public_key_into`  is now the required method, `public_key` has a default that calls it.

dpe/src/commands/certify_key.rs
parvathib: The CSR path was the immediate stack overflow but agree it makes sense to unify both paths.  Planning to split into:
- **PR1**: Signer trait changes ( `sign_into_slice` ,  `public_key_into`  as required methods with defaults for  `sign` / `public_key` )
- **PR2**: Buffer-backed encoding in `x509/certify_key` covering both CSR and X509 with a unified path 

Does that split work for you?

dpe/src/x509.rs
parvathib: Done — split into two commits:

- PR1 (**This PR**): Signer trait changes (sign_into_slice, public_key_into as required methods with defaults for sign/public_key) 

- PR2 (#600): Buffer-backed CertifyKey for both CSR and X509 formats (depends on PR1)

crypto/src/lib.rs
```diff
@@ -656,6 +759,35 @@ pub trait Crypto {
     /// * `data` - Data to be signed.
     fn sign_with_alias(&mut self, data: &SignData) -> Result<Signature, CryptoError>;
 
+    /// Sign `data` with the platform Alias Key into a caller-provided byte slice.
+    fn sign_with_alias_into_slice(
+        &mut self,
+        data: &SignData,
+        _signature: &mut [u8],
+    ) -> Result<(), CryptoError> {
+        let sig = self.sign_with_alias(data)?;
```
zhalvorsen: Should we swap this as well?

zhalvorsen (dismissed):

crypto/src/lib.rs
parvathib: Done.  `sign_with_alias_into_slice`  is now the required method and  `sign_with_alias`  calls it. Same pattern as the Signer trait

zhalvorsen (approved):

PR #597, opened 2026-06-18
Minimize MAX_CERT_SIZE using linear approximation 

### LINKED ISSUE #575: MAX_CERT_SIZE should be derived from MAX_HANDLES, not hardcoded
`MAX_CERT_SIZE` is currently a hardcoded constant (22 * 1024 with ml-dsa, 11 * 1024 without), sized for the worst-case 64-handle scenario introduced in #511 as per the description. However, MAX_HANDLES is independently configurable via the `arbitrary_max_handles` feature (defaulting to 24 in build.rs).

This means a build with `ARBITRARY_MAX_HANDLES=24` still allocates 22KB cert buffers sized for 64 handles. Since `CertifyKeyResp` variants embed `cert: [u8; MAX_CERT_SIZE]`, this oversized constant propagates into every response struct in downstream consumers (e.g., caliptra-mcu-sw), causing unnecessary stack pressure on resource-constrained targets.

leongross:
- Add `benchmark` mode to `tools/cert-size` to empirically determine linear approximation  for `p256`, `mldsa`, and `p384`.
- Update `dpe/build.rs` to provide approximated sizes for `MAX_CERT_SIZE`  to reduce memory footprint.

### DIFF
--- a/dpe/build.rs
+++ b/dpe/build.rs
@@ -34,7 +34,7 @@ fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
     let max_handles_str = format!("pub const MAX_HANDLES: usize = {};", max_handles);
-    let handles_path = PathBuf::from(format!("{}/arbitrary_max_handles.rs", out_dir));
+    let handles_path = PathBuf::from(format!("{}/max_handles.rs", out_dir));
 
     write_if_changed(&handles_path, &max_handles_str);
     println!("cargo:rerun-if-changed={}", handles_path.display());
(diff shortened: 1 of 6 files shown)


dpe/build.rs
```diff
@@ -3,27 +3,70 @@
 use std::env;
 use std::path::PathBuf;
 
+/// Default MAX_HANDLES when the `arbitrary_max_handles` feature is NOT enabled.
+/// Must match the hardcoded constant in `dpe/src/lib.rs`.
+const DEFAULT_MAX_HANDLES: usize = 64;
```
zhalvorsen, line 8: Should we instead completely remove the assignment of `MAX_HANDLES` from `dpe/src/lib.rs` and include it in the `arbitrary_max_handles.rs` generated file so we have a single source of truth? If so, we should probably rename the `arbitrary_max_handles.rs` file as well.

Another option would be to combine the two files and have a `constants.rs` or something

dpe/build.rs
```diff
@@ -3,27 +3,70 @@
...
+        // P384
+        (P384_BASE, P384_HANDLE)
+    };
+
+    let max_cert_size = base + per_handle * max_handles;
+
+    // Round up to next multiple of 4 for alignment. The response structs (e.g.
+    // CertifyKeyP384Resp) contain [u8; MAX_CERT_SIZE] followed by u32 fields,
+    // and `zerocopy::IntoBytes` requires no padding.
+    let max_cert_size = (max_cert_size + 3) & !3;
```
zhalvorsen: nit: you could use the `next_multiple_of` usize function to make this a little more readable

zhalvorsen: Can you have it round to the next multiple of 256 instead just to be on the safe side?

zhalvorsen (dismissed):

dpe/build.rs
leongross: Yes, that sounds like a good way to clean the constant generation up.

zhalvorsen (approved):

PR #599, opened 2026-06-22
dpe: Add a response buffer abstraction

ternst-nv:
Add an abstraction over how a ResponseBuffer is populated and used.  This allows the DPE to maintain byte level semantics over the location to write its response to, even if the underlying hardware does not (like the MBOX SRAM).

This can be used for significant stack optimizations which remove the need for Caliptra responses to be stored on the stack at all.

### DIFF
--- a/response-buffer/src/lib.rs
+++ b/response-buffer/src/lib.rs
@@ -1,0 +1,315 @@
+// Licensed under the Apache-2.0 license
+
+#![no_std]
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ResponseBufError {
+    Overflow,
+}
+
+/// Sink for a serialized DPE response.
+pub trait ResponseBuffer {
+    /// Zero the entire backing storage.
+    fn clear(&mut self) -> Result<(), ResponseBufError>;
+
+    /// Write `bytes` starting at `offset`.
+    ///
+    /// Returns `Overflow` if `offset + bytes.len()` exceeds the backing
+    /// store capacity.
+    fn write_at(&mut self, offset: usize, bytes: &[u8]) -> Result<(), ResponseBufError>;
+
+    /// Read back a range of already-written bytes by invoking `f` one or more
+    /// times with consecutive chunks that together cover exactly the bytes in
+    /// `range`.
+    ///
+    /// Implementations can `f` exactly once, e.g. if they are backed by
+    /// contiguous storage.
+    /// Implementations backed by word-addressed storage (e.g. MBOX SRAM) may
+    /// call `f` multiple times to handle unaligned head/tail bytes.
+    fn read_range(
+        &self,
+        range: core::ops::Range<usize>,
+        f: &mut dyn FnMut(&[u8]) -> Result<(), ResponseBufError>,
+    ) -> Result<(), ResponseBufError>;
+
+    /// Total number of bytes the backing store can hold.
+    fn capacity(&self) -> usize;
+}
+
+/// A [`ResponseBuffer`] view that shifts every offset by a fixed `base`.
+///
+/// Writes at `offset` on this wrapper land at `base + offset` in the
+/// underlying buffer.  Useful when one response buffer is divided into a header
+/// section (written directly at absolute offsets by the caller) and a body
+/// section managed by a subsystem that writes relative to its own 0-origin.
+pub struct OffsetResponseBuffer<'a> {
+    inner: &'a mut dyn ResponseBuffer,
+    base: usize,
+}
+
+impl<'a> OffsetResponseBuffer<'a> {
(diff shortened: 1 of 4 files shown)


response-buffer/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub fn new(buf: &'a mut [u8]) -> Self {
+        Self { buf }
+    }
+}
+
+impl<'a> ResponseBuffer for SliceResponseBuffer<'a> {
+    fn clear(&mut self) -> Result<(), ResponseBufError> {
+        for b in self.buf.iter_mut() {
+            *b = 0;
+        }
```
pjhades: nit: maybe just `self.buf.fill(0);`.

response-buffer/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    ///
+    /// Returns `Overflow` if `offset + bytes.len()` exceeds the backing
+    /// store capacity.
+    fn write_at(&mut self, offset: usize, bytes: &[u8]) -> Result<(), ResponseBufError>;
+
+    /// Read back a range of already-written bytes by invoking `f` one or more
+    /// times with consecutive chunks that together cover exactly the bytes in
+    /// `range`.
+    ///
+    /// Implementations can `f` exactly once, e.g. if they are backed by
```
pjhades: nit: can _call_ `f`?

response-buffer/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// Implementations backed by word-addressed storage (e.g. MBOX SRAM) may
+    /// call `f` multiple times to handle unaligned head/tail bytes.
+    fn read_range(
+        &self,
+        range: core::ops::Range<usize>,
+        f: &mut dyn FnMut(&[u8]) -> Result<(), ResponseBufError>,
+    ) -> Result<(), ResponseBufError>;
+
+    /// Total number of bytes the backing store can hold.
+    fn capacity(&self) -> usize;
```
pjhades, line 36: I wonder if it makes sense to expose an interface that returns the number of bytes written into the buffer, as we eventually need to set the DLEN register of the mailbox.

Also, nit: personally the name `capacity` makes me think about the length/capacity semantics in `Vec`.

pjhades (dismissed):
Looks good, left a few minor comments.

response-buffer/src/lib.rs
zhalvorsen: nit: this could be simplified with `fill`
```suggestion
        self.buf.fill(0)
```

response-buffer/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+
+impl<'a> OffsetResponseBuffer<'a> {
+    pub fn new(inner: &'a mut dyn ResponseBuffer, base: usize) -> Self {
+        Self { inner, base }
+    }
+}
+
+impl ResponseBuffer for OffsetResponseBuffer<'_> {
+    fn clear(&mut self) -> Result<(), ResponseBufError> {
+        self.inner.clear()
```
zhalvorsen: Should this only clear the offset range?

response-buffer/src/lib.rs
```diff
@@ -0,0 +1,315 @@
+// Licensed under the Apache-2.0 license
```
zhalvorsen, line 1: Is there are reason this is pulled into a separate crate? I think it could live within the `caliptra-dpe` to avoid including more crates in dependent projects

response-buffer/src/lib.rs
ternst-nv: Yeah, and I think it could be thought of as similar to capacity in `Vec`.  The physical size of the buffer, but not how much is being used.

Since the previous code uses slices for this value, there are already parallel structures for returning the size of a Response.  For now I think staying with that semantic works, and in the future we could investigate including the size within this struct.

response-buffer/src/lib.rs
ternst-nv: Good point, I'll update that.

response-buffer/src/lib.rs
ternst-nv: Yes, we need this to live in DPE since it is going to be referenced by it, but a lot of other `caliptra-sw` responses which don't utilize DPE functionality will also need it to serialize out their response buffers.  

I thought the best way of representing that would be to separate it out into its own crate to indicate what specific functionality is being used, and not that it is generally dependent on DPE.

zhalvorsen: Okay, I don't feel too strongly about it.

Looking back, I think we maybe should have collapsed the `caliptra-dpe-crypto` and `caliptra-dpe-platform` crates into `caliptra-dpe`. It may have simplified some of the feature flags we have which wouldn't apply to this crate.

zhalvorsen (approved):

PR #601, opened 2026-06-23
dpe: Utilize ResponseBuffer abstraction

ternst-nv:
Instead of having the DPE write directly into a u8 slice, abstract the access over a `ResponseBuffer`.  This allows the use of hardware buffers, like the MBOX SRAM, which have additional constraints like word aligned access.

This allows an optimization of the `caliptra-1.x` which saves significant Stack utilization allowing for the use of MLDSA libraries with in DPE invocations.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -5,8 +5,12 @@ Abstract:
 --*/
 #![cfg_attr(not(any(feature = "rustcrypto", test)), no_std)]
 
+use core::ops::Range;
+
 use ecdsa::EcdsaSignature;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+use caliptra_dpe_response_buffer::ResponseBuffer;
 
 #[cfg(feature = "rustcrypto")]
 pub use crate::rustcrypto::*;
@@ -313,10 +317,10 @@ impl From<Mu> for PrecomputedSignData {
     }
 }
 
-#[derive(Debug)]
 pub enum SignData<'a> {
     Digest(Digest),
     Mu(Mu),
+    ResponseBuffer(&'a dyn ResponseBuffer, Range<usize>),
     Raw(&'a [u8]),
 }
 
@@ -325,14 +329,8 @@ impl SignData<'_> {
         match self {
             Self::Digest(dig) => dig.size(),
             Self::Mu(mu) => mu.0.len(),
+            Self::ResponseBuffer(_, range) => range.len(),
             Self::Raw(raw) => raw.len(),
-        }
-    }
-    pub fn as_slice(&self) -> &[u8] {
-        match self {
-            Self::Digest(dig) => dig.as_slice(),
-            Self::Mu(mu) => mu.0.as_slice(),
-            Self::Raw(raw) => raw,
         }
     }
 }
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -4,7 +4,6 @@ use crate::{
     context::ContextHandle,
     dpe_instance::{DpeEnv, DpeInstance},
     error::{DpeErrorCode, InternalErrorCode},
-    mutresp, okref,
     x509::{create_dpe_cert, CreateDpeCertArgs, CreateDpeCertResult},
     DpeFlags, DpeProfile,
 };
@@ -16,6 +15,7 @@ use caliptra_cfi_lib::{cfi_assert, cfi_assert_bool, cfi_assert_eq};
 #[cfg(any(feature = "p256", feature = "p384"))]
 use caliptra_dpe_crypto::ecdsa::EcdsaPubKey;
 use caliptra_dpe_crypto::PubKey;
+use caliptra_dpe_response_buffer::{OffsetResponseBuffer, ResponseBuffer};
 use cfg_if::cfg_if;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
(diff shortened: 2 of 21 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -368,52 +330,86 @@ impl CommandExecution for CertifyKeyMldsa87Cmd {
...
+                let hdr = CertifyKeyP256RespHdr {
+                    resp_hdr: dpe.response_hdr(DpeErrorCode::NoError),
+                    new_context_handle: new_handle,
+                    derived_pubkey_x: *pk_x,
+                    derived_pubkey_y: *pk_y,
+                    cert_size,
+                };
+                out.write_at(0, hdr.as_bytes())
+                    .map_err(|_| DpeErrorCode::InvalidResponseBuf)?;
+                Ok(core::mem::size_of::<CertifyKeyP256RespHdr>() + cert_size as usize)
```
pjhades: Nit: maybe calling `self.hdr_size()` here and below, so that writing to `out` can be extracted out of the match block.

dpe/src/commands/certify_key.rs
```diff
@@ -94,20 +94,15 @@ impl CertifyKeyCommand<'_> {
         }
     }
 
-    fn response_bytes<'a>(
-        &self,
-        p: DpeProfile,
-        out: &'a mut [u8],
-    ) -> Result<CertifyKeyResponseBytes<'a>, DpeErrorCode> {
-        let r = match self {
+    fn response_bytes(&self) -> CertifyKeyResponseBytes {
```
pjhades: Nit: the name of this function feels a bit misleading now. But I'm not sure what would be a good alternative.

dpe/src/commands/derive_context.rs
```diff
@@ -290,7 +292,8 @@ impl CommandExecution for DeriveContextCmd {
         if flags.is_recursive() {
             cfg_if! {
                 if #[cfg(not(feature = "disable_recursive"))] {
-                    let response = mutresp::<DeriveContextResp>(dpe.profile, out)?;
+                    let mut scratch = AlignedBuf::<{ size_of::<DeriveContextResp>() }>::new();
```
pjhades, line 295: Out of curiosity, why does this buffer need to be 4-byte aligned?

dpe/src/x509.rs
```diff
@@ -2539,23 +2562,22 @@ impl CertWriter<'_> {
...
-            let Some(csr_range) = self.csr_range else {
+        let (csr_start, csr_end) = {
+            let Some(r) = self.csr_range else {
                 Err(DpeErrorCode::X509CsrUnset)?
             };
-            self.certificate
-                .get(csr_range.0..csr_range.1)
-                .ok_or(DpeErrorCode::from(InternalErrorCode::CmsCsrRangeOob))?
+            r
         };
```
pjhades: Nit: could be simplified as `self.csr_range.ok_or(...)`.

dpe/src/x509.rs
```diff
@@ -2147,7 +2149,12 @@ impl CertWriter<'_> {
     #[allow(clippy::identity_op)]
     fn encode_encapsulated_content_info(
         &mut self,
-        sign_cb: &mut (impl FnMut(&[u8], bool) -> Result<Signature, CryptoError> + ?Sized),
+        sign_cb: &mut (impl FnMut(
+            &dyn ResponseBuffer,
+            core::ops::Range<usize>,
+            bool,
+        ) -> Result<Signature, CryptoError>
+                  + ?Sized),
```
pjhades: Nit: since the closure type becomes a bit more complex, would it make sense to define a type alias for it?

pjhades (dismissed):
Looks good in general. Left a few comments where the code could possibly be simplified.

dpe/src/commands/derive_context.rs
ternst-nv: The underlying type want to be 4 byte aligned: https://github.com/chipsalliance/caliptra-dpe/blob/9407a91bc66112af7c8315d50aba75fd223a1a6e/dpe/src/response.rs#L249

dpe/src/commands/certify_key.rs
ternst-nv: I deduplicated the `Ok` to use `hdr_size`.  Changing the `write_at` would require storing the header type so didn't think it was worth it.

dpe/src/commands/certify_key.rs
ternst-nv: Agreed, renamed the function and enum to be ResponseHeader.

crypto/src/lib.rs
```diff
@@ -313,10 +317,10 @@ impl From<Mu> for PrecomputedSignData {
     }
 }
 
-#[derive(Debug)]
 pub enum SignData<'a> {
     Digest(Digest),
     Mu(Mu),
+    ResponseBuffer(&'a dyn ResponseBuffer, Range<usize>),
```
zhalvorsen, line 323: Would it make more sense to replace the `Raw` inner with a `ResponseBuffer` instead of a separate type?

pjhades (approved):
LGTM!

zhalvorsen (approved):

crypto/src/lib.rs
zhalvorsen: Separate *variant* is probably a better term here

ternst-nv: Hmm, it looks like we support using the `Raw` sign data in other commands: https://github.com/chipsalliance/caliptra-dpe/blob/9407a91bc66112af7c8315d50aba75fd223a1a6e/dpe/src/commands/sign.rs#L167, so I think we want to leave them separate.

zhalvorsen: Ack, I can do something in a follow-on

PR #606, opened 2026-06-26
Optimize DPE signing paths to reduce stack usage

parvathib:
Optimized `SignCallback` in X.509/CSR generation to pass a mutable reference of `Signature` instead of returning the large `Signature` enum by value. This reclaims up to ~4.6 KB of active stack space during bare-metal signature generation.

This entirely removes the need to pass public keys or change public cryptographic traits, keeping `CryptoSuite` and `Crypto` interfaces perfectly clean and unchanged.

### DIFF
--- a/dpe/src/x509.rs
+++ b/dpe/src/x509.rs
@@ -37,16 +37,26 @@ use caliptra_dpe_crypto::ml_dsa::{MldsaPublicKey, MldsaSignature};
 
 /// Signing callback passed to cert and CSR encoding functions.
 ///
-/// Arguments: `(buf, range, use_derived)` where `range` is the byte span of
-/// the payload to sign within `buf`, and `use_derived` selects the derived key
-/// (`true`) or the alias key (`false`).
+/// Arguments: `(buf, range, use_derived, out_sig)` where `range` is the byte span of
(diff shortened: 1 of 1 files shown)


dpe/src/x509.rs
```diff
@@ -2555,7 +2565,8 @@ impl CertWriter<'_> {
         let (csr_start, csr_end) = self.csr_range.ok_or(DpeErrorCode::X509CsrUnset)?;
         let csr_len = csr_end - csr_start;
 
-        let sig = sign_cb(&*self.certificate, (csr_start)..(csr_end), false)?;
```
zhalvorsen, line 2558: Is this the main hot-path? I wonder if just changing this line in the original implementation to use `okref` would get you close enough. Just using the `?` here puts two signatures on the stack. One `Signature` and one `Result<Signature, DpeErrorCode>`.

Using the following instead would reduce this to only `Result<Signature, DpeErrorCode>` on the stack:

```Rust
let sig = sign_cb(&*self.certificate, (csr_start)..(csr_end), false)?;
let sig = okref(&sig)?;
```

parvathib: The clever and simple. I've opened an official PR for that change. I'll close this one.

PR #612, opened 2026-07-15
Add feature gates for hybrid test flag

leongross:
- Gate unit tests behind accurate feature gates
- Add a hybrid profile to xtask ci runner


clundin25 (approved):
Thanks @leongross !

zhalvorsen:
Just double checking my logic here. This will run all the P384 profile tests against the hybrid model right? And then all the ML-DSA tests will only run when ML-DSA is enabled, but not hybrid?

Could we also add a test that runs against both profiles to make sure a couple of the commands can be run with either profile when hybrid is enabled? I would say `DeriveContext`, `CertifyKey`, and `Sign` would be enough

PR #616, opened 2026-07-22
Add V1 definition of `DeriveContext`

zhalvorsen:
This can help with backwards compatibility logic. See https://github.com/chipsalliance/caliptra-sw/pull/3963 for context.

This also simplifies the `Default` implementation for `DeriveContextCmd`.

### DIFF
--- a/dpe/src/commands/derive_context.rs
+++ b/dpe/src/commands/derive_context.rs
@@ -22,7 +22,9 @@ use core::mem::size_of;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 #[repr(C)]
-#[derive(Debug, PartialEq, Eq, Clone, Copy, FromBytes, IntoBytes, Immutable, KnownLayout)]
+#[derive(
+    Debug, Default, PartialEq, Eq, Clone, Copy, FromBytes, IntoBytes, Immutable, KnownLayout,
+)]
 pub struct DeriveContextFlags(pub u32);
 
 bitflags! {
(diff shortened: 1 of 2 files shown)


clundin25:
@zhalvorsen @ternst-nv I feel this is a Caliptra issue, not a DPE issue. We've already bumped major versions, shouldn't the translation occur upstream?

zhalvorsen:
Yeah, we'll add the handling logic on the Caliptra side. This just defines the ABI layout of the previous version so it can be referenced from `caliptra-sw`

clundin25:
> Yeah, we'll add the handling logic on the Caliptra side. This just defines the ABI layout of the previous version so it can be referenced from `caliptra-sw`

Would it make more sense to import the older ABI as it's own crate? I don't really think that would cause much code bloat since it's just a small struct

ternst-nv:
> Would it make more sense to import the older ABI as it's own crate? I don't really think that would cause much code bloat since it's just a small struct

Hmm, I think it is fair for a library crate to be required to maintain the history of its protocol to support backwards compatibility, and so we shouldn't use a new crate for this. 

In the future I think we'll need to be able to support this pattern in general, otherwise as we move to support new Caliptra releases and add DPE features we'd need to support various branches for each Caliptra SW release which I think will become quite the maintenance burden.

clundin25:
> > Would it make more sense to import the older ABI as it's own crate? I don't really think that would cause much code bloat since it's just a small struct
> 
> Hmm, I think it is fair for a library crate to be required to maintain the history of its protocol to support backwards compatibility, and so we shouldn't use a new crate for this.
> 
> In the future I think we'll need to be able to support this pattern in general, otherwise as we move to support new Caliptra releases and add DPE features we'd need to support various branches for each Caliptra SW release which I think will become quite the maintenance burden.

I don't think a new crate, you can import two copies of the same crate using https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml.

So in caliptra-sw we can import the version of DPE that has the pre-SVN API, instead of redefining it on main

zhalvorsen:
I pushed a second commit that shows what it would look like to import both versions of the DPE library. Let me know what you think. I don't have a super strong opinion about this.

zhalvorsen:
Oh sorry I pushed it to https://github.com/chipsalliance/caliptra-sw/pull/3963

PR #618, opened 2026-07-31
x509: add golden x509 certificate tests

### LINKED ISSUE #598: Add a golden X509 certificate and CSR to check encoding changes
https://github.com/chipsalliance/caliptra-sw/blob/main/test/tests/caliptra_integration_tests/smoke_test.rs has good examples of how this might work. It would be good to see when a change to the X509 library affects the generation of the certificate versus being purely an optimization to the functionality.

leongross:
- add golden x509 cert/csr unit tests (see #598)
- add golden certificates and `openssl` scripts to generate them in-tree 
- add x509 and CSR in human-readable text format


dpe/src/x509_testdata/golden_cert_mldsa87.der
zhalvorsen, line 1: Can you also upload the text version of each of these so we can easily see the differences when they happen?

dpe/src/x509_testdata/regen.sh
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env bash
```
zhalvorsen, line 1: Sorry, if this is a lot of toil. But could you convert these to xtask flows so they are easier to manage? If you prefer, it would be fine to do in a follow-on

dpe/src/x509.rs
```diff
@@ -5006,4 +5006,109 @@ pub(crate) mod tests {
...
+        #[cfg(feature = "p384")]
+        const GOLDEN_CSR_DER: &[u8] = include_bytes!("x509_testdata/golden_csr_p384.der");
+        #[cfg(all(feature = "ml-dsa", not(feature = "p384"), not(feature = "p256")))]
+        const GOLDEN_CSR_DER: &[u8] = include_bytes!("x509_testdata/golden_csr_mldsa87.der");
+
+        /// Parse the golden certificate, verify its self-signature, and assert
+        /// on its structure and DICE / TCI extensions.
+        #[test]
+        #[cfg_attr(miri, ignore)]
+        fn test_golden_cert() {
```
zhalvorsen, line 5043: Do you plan to add tests that check the golden certificate against a certificate created by the library in a later change?

zhalvorsen (dismissed):

dpe/src/x509_testdata/regen.sh
leongross: Sure thing! I would prefer a follow-up, though.

dpe/src/x509.rs
leongross: I'd prefer to do that in a follow-up as well to keep things clean.

dpe/src/x509_testdata/golden_cert_mldsa87.der
leongross: See c972e64c1e46b9af3e343b662789fbb09cf3b915

zhalvorsen (approved):

dpe/src/x509_testdata/golden_cert_mldsa87.der
zhalvorsen: Thanks!

PR #620, opened 2026-08-06
dpe: local refactor to reduce code size

pjhades:
- x509.rs: simplify check of exported_cdi_handle
- sign.rs: simplify response match arms

Reclaims 240 bytes of code size.

```
┌────────────────────────┬────────┬──────┬─────────┬───────┬────────┐
│         Binary         │  Text  │ Data │ Budget  │ Used% │  Free  │
├────────────────────────┼────────┼──────┼─────────┼───────┼────────┤
│ Runtime with ML-DSA    │ 99,564 │ 128  │ 103,684 │ 96.1% │ 3,992  │
├────────────────────────┼────────┼──────┼─────────┼───────┼────────┤
│ Runtime with ML-DSA    │ 99,324 │ 128  │ 103,684 │ 95.8% │ 4,232  │
└────────────────────────┴────────┴──────┴─────────┴───────┴────────┘
```

### DIFF
--- a/dpe/src/commands/sign.rs
+++ b/dpe/src/commands/sign.rs
@@ -20,6 +20,13 @@ use cfg_if::cfg_if;
 use core::mem::size_of;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
+#[cfg(feature = "ml-dsa")]
+use crate::response::SignMlDsaResp;
+#[cfg(feature = "p256")]
+use crate::response::SignP256Resp;
+#[cfg(feature = "p384")]
+use crate::response::SignP384Resp;
+
 #[repr(C, align(4))]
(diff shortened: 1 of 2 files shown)


dpe/src/commands/sign.rs
```diff
@@ -190,69 +197,65 @@ impl CommandExecution for SignCommand<'_> {
         }
 
         let sig = sign(dpe, env, idx, label, &data);
+        let sig_ref = okref(&sig)?;
+
+        // Rotate the handle if it isn't the default context.
+        let mut ctx = env.state().contexts[idx];
```
zhalvorsen, line 203: Part of the reason this is duplicated everywhere is because, if for some reason the match fails below, it will rotate the context handle, but the caller won't receive it so it will no longer be accessible. Instead we could rotate the copied context like you are doing and then assign the copy to the `state` after the match statement

zhalvorsen (approved):

PR #621, opened 2026-08-06
dpe: Make signature out parameter

ternst-nv:
Change the `sign` trait to return the signature as an out parameter instead of returning by value.

This allows the internal DPE logic to utilize a consistent buffer for signatures, saving ~9Kb of stack space for standard signing requests.

This requires a change to the public facing API as well.

### DIFF
--- a/dpe/src/commands/certify_key.rs
+++ b/dpe/src/commands/certify_key.rs
@@ -193,7 +193,10 @@ impl CommandExecution for CertifyKeyCommand<'_> {
         let response_header = self.response_header();
         let hdr_size = response_header.hdr_size();
         let mut cert_view = OffsetResponseBuffer::new(out, hdr_size);
-        let result = Self::run_cert_format(format, &args, dpe, env, &mut cert_view)?;
+        // Single result slot, filled in place by the cert/CSR path, so only one
+        // `CreateDpeCertResult` — and thus one `PubKey` — occupies this frame.
+        let mut result = CreateDpeCertResult::default();
+        Self::run_cert_format(format, &args, dpe, env, &mut cert_view, &mut result)?;
 
         #[allow(clippy::indexing_slicing)]
         let new_handle = {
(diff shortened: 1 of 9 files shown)


dpe/src/commands/certify_key.rs
```diff
@@ -395,14 +399,29 @@ impl CertifyKeyResponseHeader {
...
+                use crate::ResponseHdr;
+
+                const HANDLE_OFFSET: usize = core::mem::size_of::<ResponseHdr>();
+                const PUB_OFFSET: usize = HANDLE_OFFSET + core::mem::size_of::<ContextHandle>();
+                const CERT_SIZE_OFFSET: usize =
+                    PUB_OFFSET + size_of::<caliptra_dpe_crypto::ml_dsa::MldsaPublicKey>();
+
+                const _: () = assert!(
+                    CERT_SIZE_OFFSET + size_of::<u32>() == size_of::<CertifyKeyMldsa87RespHdr>()
+                );
```
zhalvorsen: Can we use `offset_of!` here just in case something moves we won't need to remember to switch it here?

dpe/src/x509.rs
```diff
@@ -40,14 +40,21 @@ use caliptra_dpe_crypto::ml_dsa::{MldsaPublicKey, MldsaSignature};
 ///
 /// Arguments: `(buf, range, use_derived)` where `range` is the byte span of
 /// the payload to sign within `buf`, and `use_derived` selects the derived key
-/// (`true`) or the alias key (`false`).
+/// (`true`) or the alias key (`false`).  The `Signature` is the out buffer
+/// is where the signature is placed.  The callee is responsible for ensuring it
```
zhalvorsen: uber nit: a couple typos in this comment and updating the arguments list:

```Rust
/// Arguments: `(buf, range, use_derived, out_sig)` where `range` is the byte span of
/// the payload to sign within `buf`, and `use_derived` selects the derived key
/// (`true`) or the alias key (`false`).  The `out_sig` is the out buffer
/// where the signature is placed.  The caller is responsible for ensuring it
```

zhalvorsen (dismissed):

zhalvorsen (approved):

pjhades (approved):

PR #622, opened 2026-08-07
x509: migrate certificate/csr regeneration to xtask

### LINKED ISSUE #598: Add a golden X509 certificate and CSR to check encoding changes

leongross:
- `cargo xtask regen-x509-testdata <profile>` regenerates all golden X.509 certificates and CSR
- `cargo xtask regen-x509-testdata --check` verify equality of generated certificates for regression
- Add shared `caliptra-dpe-dice-asn1` crate containing reusable UEID, FWID, and TCB ASN.1 structures
- Update `caliptra-dpe-tool`s to reuse the shared ASN.1 parsing structures

Follow-up on #618


xtask/src/x509_testdata.rs
```diff
@@ -0,0 +1,345 @@
...
+            args.join(" "),
+            String::from_utf8_lossy(&output.stderr).trim()
+        );
+    }
+    Ok(output.stdout)
+}
+
+fn check_openssl() -> Result<()> {
+    run_openssl(&["version"], None).map(|_| ())
+}
```
zhalvorsen, line 295: nit: It might be helpful to add a little extra context to this error:


```suggestion
fn check_openssl() -> Result<()> {
    run_openssl(&["version"], None)
        .context("OpenSSL is required but could not be found. Please ensure it is installed and in your PATH.")
        .map(|_| ())
}
```

xtask/src/x509_testdata.rs
```diff
@@ -0,0 +1,345 @@
...
+            "subjectKeyIdentifier=hash",
+            "-addext",
+            "basicConstraints=critical,CA:FALSE",
+            "-addext",
+            "keyUsage=critical,digitalSignature",
+        ]);
+    } else {
+        args.extend(["-addext", "basicConstraints=critical,CA:FALSE"]);
+    }
+    let ueid = format!("2.23.133.5.4.4=critical,DER:{}", hex(&extensions.ueid));
```
zhalvorsen: nit: could we instead use the `hex` crate here?

zhalvorsen (dismissed):

xtask/src/x509_testdata.rs
leongross: From your input, I reviewed the `caliprta-dpe` codebase, and it seems as if there are several different ways of how byte-to-hex-string encodings are performed.

For now, I will use the `hex` crate as suggested. We should, however, consider unifying hex encodeing/decoding in this repo.

zhalvorsen (approved):

PR #623, opened 2026-08-10
dpe: make public key an output argument for Signer

pjhades:
Avoid returning the public key by-value from `Signer::public_key`, saving stack space for ML-DSA signing path.

Stack size comparison before and after the change:
```
command                               bytes  % stack    command                               bytes  % stack
----------------------------------------------------    ----------------------------------------------------
CERTIFY_KEY_EXTENDED_MLDSA87          47760    54.9%    CERTIFY_KEY_EXTENDED_MLDSA87          45168    51.9%
```

Caliptra-sw runtime code size is about 100 bytes smaller. Comparison before and after this change:
```
│ Runtime with ML-DSA    │ 97,652 │ 128  │ 103,684 │ 94.2% │ 6,032 │
│ Runtime with ML-DSA    │ 97,420 │ 128  │ 103,684 │ 94.1% │ 6,136 │
```

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -365,20 +365,6 @@ pub enum PubKey {
     Mldsa(ml_dsa::MldsaPublicKey),
 }
 
-impl Default for PubKey {
-    /// A zeroed placeholder key. Used only to pre-initialize a caller-provided
-    /// `CreateDpeCertResult` slot that the cert/CSR path overwrites before use;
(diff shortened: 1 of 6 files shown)


ternst-nv (approved):
LGTM

pjhades:
Adding Carl for having another review.

Also, could any of you merge once this is approved by both of you? Not sure why, but I remember I had push access before.

clundin25 (approved):

PR #625, opened 2026-08-14
x509: refactor common test data to GoldenProfile structs

leongross:
The PR is quite large, but only 2d875eefadd4bb7c4096e4591ca6a0fe08cbf260 0aa5bdb5b4a484a6cccdd75af2722d8eb54cb76c carry relevant code changes.


dice-asn1/src/lib.rs
```diff
@@ -1,6 +1,179 @@
...
+    pub fwid_digest_fill: u8,
+}
+
+impl GoldenDefinitions {
+    const fn build(profile: GoldenProfile) -> Self {
+        let (artifact_suffix, hash_size, hash_oid, key_algorithm) = match profile {
+            GoldenProfile::P256 => (
+                "p256",
+                32,
+                Oid {
```
zhalvorsen, line 99: Can you add constants for some of these other OIDs to make this function a little more readable?

dice-asn1/src/lib.rs
```diff
@@ -146,25 +158,36 @@ impl GoldenDefinitions {
             subject_common_name: "DPE Leaf",
             csr_subject_serial: "0000",
             certificate_serial: "0",
-            not_before: "20230227000000Z",
+            // Use GeneralizedTime on both the OpenSSL and custom encoder paths.
+            not_before: "20500227000000Z",
```
zhalvorsen, line 162: Is there are reason this isn't using the previous value of `20230227000000Z`? This value is in 2050 so any verifier with a concept of time would reject this value

dice-asn1/src/lib.rs
```diff
@@ -146,25 +158,36 @@ impl GoldenDefinitions {
...
                 oid: MULTI_TCB_INFO_OID,
-                critical: false,
+                critical: true,
             },
             tcb_info_count: 1,
             fwids_per_tcb_info: 1,
             fwid_digest_fill: 0,
+            svn: 0,
+            locality: 0,
+            tci_type: 0,
```
zhalvorsen, line 189: Can we use non-zero values for these to make sure they are properly captured in the certificate and encoded correctly?

PR #629, opened 2026-08-28
fix: resolve #628 - Migrate verification tests to standard library ML…

### LINKED ISSUE #628: Migrate verification tests to standard library ML-DSA
Currently we use circl: https://github.com/chipsalliance/caliptra-dpe/blob/main/verification/testing/go.mod#L14.

Now that Go v1.27 is out and supports ML-DSA, we should drop this dependency and use the standard library.

clundin25:
…-DSA (#2)

* fix: resolve #628 - Migrate verification tests to standard library ML-DSA

### DIFF
--- a/verification/testing/certifyKey.go
+++ b/verification/testing/certifyKey.go
@@ -5,6 +5,7 @@ package verification
 import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"crypto/mldsa"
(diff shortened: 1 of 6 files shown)


zhalvorsen (approved):

zhalvorsen:
We should probably drop the `#2` from the description

PR #633, opened 2026-09-17
Fix to multi TCB info size calculations

### LINKED ISSUE #632: MultiTcbInfo size can be miscalculated if SVN values require different length encoding
get_multi_tcb_info_size and encode_multi_tcb_info functions calculate the total MultiTcbInfo size by multiplying the first node's size by the node count. SVN field is DER-encoded as a minimal-length INTEGER whose byte length depends on the value (e.g. SVN=0 encodes to 3 bytes as [3] 0x00, while SVN=128 requires a leading 0x00 sign byte and encodes to 4 bytes). This can lead to an inaccurate size and a corrupted cert.

nquarton:


dpe/src/x509.rs
```diff
@@ -640,13 +636,20 @@ impl CertWriter<'_> {
...
-        // Size of concatenated tcb infos
-        let tcb_infos_size = measurements.tci_nodes.num_nodes()?
-            * self.get_tcb_info_size(
-                measurements.tci_nodes.first_node()?,
-                measurements.supports_recursive,
-                /*tagged=*/ true,
-            );
+        // Size of concatenated tcb infos — sum each node individually since
+        // SVN is DER-encoded and its byte length depends on the value.
+        let tcb_infos_size = measurements
```
zhalvorsen: nit: can we make this a helper function so it is common in both places?

nquarton: Yeah we should definitely do that. Sorry, it crossed my mind then I just didn't follow through on it.

zhalvorsen (approved):

zhalvorsen:
Looks good, thanks Nick!

